PR #103, opened 2026-05-21
fix: escape quotes in hooks.json to fix JSON parse error

KorenKrita:
Commit 87c6ec56 accidentally removed the backslash escapes from the double quotes in the SessionStart hook command, breaking JSON parsing.

**Error:**
```
JSON Parse error: Expected '}', line 8 column 31
```

**Before (broken):**
```json
"command": "bash "${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh""
```

**After (fixed):**
```json
"command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh\""
```

The inner quotes around `${CLAUDE_PLUGIN_ROOT}/...` must be escaped as `\"` for valid JSON.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash "${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh""
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh\""
           }
         ]
       }


fdaviddpt:
Thanks @KorenKrita — clean fix, merging.

Side note: we noticed our own `.supertool.json` in this repo had no `jsonlint` validator hooked into the mutating ops, which is why supertool didn't catch the bad escapes when we wrote them. Adding one now so the next time we (or anyone) edit a `*.json` through supertool, it'll be rolled back on syntax fail. Belt + suspenders.

Appreciated! 🙏

PR #179, opened 2026-05-24
watch: github-pr source (reuses gh-pr op's _gh helper)

fdaviddpt:
Adds a github-pr source for the watch preset.

## Summary

- New source under \`presets/watch/sources/github-pr/\` (events.json + poller.py)
- Polls a GitHub PR via \`gh pr view --json state,mergeable,reviewDecision,statusCheckRollup,...\` every 30s
- Emits transitions: \`checks_failed\` / \`checks_succeeded\` / \`checks_pending\` (statusCheckRollup aggregated like the GitHub UI), \`review_approved\` / \`review_changes_requested\`, \`merged\` / \`closed\`, \`conflicts_appeared\` (rising-edge)
- **Reuses \`_gh\` from \`presets/github/pr.py\` via importlib** — single source of truth for gh CLI invocation, no duplicated wrapping
- Adds \`watch\` to \`.supertool.example.json\` presets list (missed in #177)

## Real use case

\`./supertool 'watch:github-pr:178'\` — Claude reacts when checks flip, reviews land, or the PR merges/conflicts. Pairs with the Phase 2 channel for async wake.

## Test plan

- [x] 17 unit tests pass (rollup aggregator + every state-diff path)
- [x] Real smoke against PR #177 — got \`MERGED\` + \`checks=SUCCESS\` + correct title/URL
- [ ] Live end-to-end: \`watch:github-pr:179\` (this PR), watch Claude react when it merges

## Out of scope

- \`github-actions\` standalone source (workflow runs without a PR) — follow-up if useful

PR #180, opened 2026-05-24
gitlab-mr: reuse gl-mr's _glab_api + add comment_added event

fdaviddpt:
Mirrors the github-pr refactor from #179.

## Summary

- Drops the inline \`_glab_api\` helper from \`presets/watch/sources/gitlab-mr/poller.py\`
- Imports \`_glab_api\` from \`presets/gitlab/mr.py\` via importlib — same pattern github-pr uses to import \`_gh\` from \`presets/github/pr.py\`
- Adds \`comment_added\` event. Uses the existing \`user_notes_count\` field from the MR JSON, so no extra API call. Fires on rising edge with \`new_count\` payload. First poll baselines without firing.

## Why it matters for your "10 MRs open" use case

Previously the gitlab-mr source missed comment-driven feedback — pipelines and merge events fired, but a reviewer leaving a comment was invisible. Now you get a wake-up when someone drops a note on a tracked MR.

## Test plan

- [x] 14 unit tests pass (was 10, +4 for the comment path + helper-reuse check)
- [ ] Live smoke on a DVSI MR: \`./supertool 'watch:gitlab-mr:<iid>'\` → add a comment via glab → confirm \`comment_added\` lands in the channel

## Out of scope

- Author detection on the comment payload — GitLab's \`user_notes_count\` doesn't include identity. Would require a second API call to \`/notes\`. Follow-up if useful.
- gitlab-job source (workflow jobs without an MR) — separate PR if you want it.

### DIFF
--- a/presets/watch/sources/gitlab-mr/events.json
+++ b/presets/watch/sources/gitlab-mr/events.json
@@ -4,6 +4,7 @@
     {"key": "pipeline_failed",      "label": "Pipeline failed"},
     {"key": "pipeline_succeeded",   "label": "Pipeline succeeded"},
     {"key": "pipeline_running",     "label": "Pipeline started running"},
+    {"key": "comment_added",        "label": "New comment on the MR"},
     {"key": "merged",               "label": "MR merged"},
     {"key": "closed",               "label": "MR closed"},
     {"key": "conflicts_appeared",   "label": "Conflicts appeared"}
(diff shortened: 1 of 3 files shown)


fdaviddpt:
🤖 Self-review (Sonnet sub-agent) — APPROVE with fixes pushed in cd0e677

**Fixed:**
- **`notes_count` fallback from 0 → None** when \`user_notes_count\` is absent. The old default would have locked the baseline at 0 forever on any MR where the field was missing. Now it stays unbaselined until a real value arrives. Updated rising-edge guard to skip when current poll's count is None too (transient field disappearance shouldn't fire spurious events).
- **System-note caveat documented in code.** GitLab's \`user_notes_count\` counts *all* notes including system notes (pipeline transitions, label edits, assignee changes), so \`comment_added\` will fire on some non-human events. Accepted limitation for v1 — fix would require a second API call to \`/notes\` per poll. Comment explains why.
- **Two new tests:** \`test_missing_user_notes_count_does_not_lock_baseline_at_zero\` + \`test_notes_count_field_disappearing_skips_event\`. 14 → 16 tests.

**Acknowledged, not fixed:**
- Author field missing from \`comment_added\` payload (github-pr has it, gitlab-mr doesn't). Documented as out-of-scope in the PR body — would need a /notes endpoint fetch. Could be a follow-up.

PR #193, opened 2026-05-25
WIP: diag at_file+batch tracebacks

fdaviddpt:
WIP diagnostic — pulls tracebacks for @file route / batch buckets on Windows. Will be reverted.


fdaviddpt:
Diag complete — root cause was _DRIVE_LETTER regex not matching @-prefixed paths. Real fix in #194.

Issue #188, opened 2026-05-25
Triage Windows test failures from #186


fdaviddpt:
After #186 added windows-latest to the CI matrix, the Windows runs surfaced ~100+ real failures across the suite. Linux + macOS are clean. Windows is currently `continue-on-error: true` so PRs still gate on linux+macos — this issue tracks the work to get Windows back to required.

## Symptom buckets (from PR #186's failing run)

Bug categories observed in the failure pattern (judging by which test files had Fs vs dots):

- **Path separators** — likely `tests/test_glob*.py`, `tests/test_read.py`, `tests/test_replace.py`, `tests/test_around*.py`, `tests/test_ls.py`, `tests/test_map.py`. Tests probably assert on `/` paths while Windows returns `\`.
- **Unix tool deps** — `tests/test_hadolint.py`, `tests/test_pyright.py`, `tests/test_ruby_check.py`, `tests/test_inilint.py`, `tests/test_jsonlint.py`, `tests/test_markdownlint.py`, `tests/test_language_formatters.py`, `tests/test_formatters*.py`. External binaries either not on Windows runner PATH or named differently (`.exe`).
- **AF_UNIX sockets** — `tests/test_mcp_*.py`, `tests/test_security_lsp.py`. `MCPClient` uses `socket.AF_UNIX`; Windows 10+ has it but the daemon spawn / socket-path scheme may not be portable.
- **Subprocess shells** — `tests/test_op_*.py`, `tests/test_notifiers*.py`, `tests/test_op_staged.py`, `tests/test_parallel.py`. Likely places that shell out via `/bin/sh` or rely on `bash` being on PATH.
- **Line endings** — `tests/test_replace.py`, `tests/test_edit_ops.py`, `tests/test_grep.py`. Git on Windows defaults to autocrlf, fixtures land with `\r\n`, regex assertions break.
- **Security boundary asserts** — `tests/test_security_devto.py`, `tests/test_security_hashnode.py`, `tests/test_security_hardening_150.py`. May depend on Unix-shell-specific quoting/escape behavior.

## Suggested approach

1. Pick one bucket (path separators is the highest-impact — touches the most tests).
2. Fix the underlying code, not just the test, when possible (e.g. supertool emits forward slashes everywhere).
3. Where Windows-only tools are unavailable (hadolint via brew, etc.), skip via `@pytest.mark.skipif(not shutil.which(...))`.
[...]

PR #200, opened 2026-05-25
WIP: diag residual #188

fdaviddpt:
WIP diag — residual cluster tracebacks. Reverted before merge.


fdaviddpt:
Diag complete — #201 fixes the grep drive-letter parsing.

PR #201, opened 2026-05-25
fix: drive-letter safe grep + refs path parsing (#188)

fdaviddpt:
fix: drive-letter safe grep + refs path parsing (#188)

## Problem

After #189 normalised output paths to forward slashes, Windows `file_path` became `C:/Users/.../src.py`. `_grep_recursive` then returned strings of the form `path:lineno:content` — joined with `:`. Downstream consumers (`op_grep`, `op_resolve` refs section) split on the first one or two `:` characters to recover the path / lineno / content. The drive letter colon (`C:`) was eaten by the split.

Visible symptom on `op_grep`:
```
(2 results in 1 files, limit 10)
C
  /Users/.../src.py:1:class Foo:
```
— path mangled to `C`, indented "lineno" was actually the file path, "content" carried the real lineno and snippet.

5 test_grep tests + 1 op_resolve test failed.

## Fix

- `_grep_recursive` now returns `List[Tuple[str, int, str]]` instead of joined strings.
- `op_grep` iterates the tuples directly — no colon parsing.
- `op_resolve` refs section keeps the joined-string interim (for MCP-route parity) but uses a drive-letter aware split: `start = 2 if path begins with [A-Z]:` so the search for the next `:` starts past the drive letter.

POSIX behaviour unchanged.

## Tests

Existing `_grep_recursive` direct tests (test_grep.py:230-263) only check `isinstance(list)` / `len()` / `== []`, so the tuple return type is transparent.

## Buckets remaining for #188

CRLF line endings in test_read fixtures, formatter binary tests, residual edge_cases, test_at_file_route TOML, test_chaos concurrent file access (Windows file lock).

### DIFF
--- a/supertool.py
+++ b/supertool.py
@@ -1139,21 +1139,15 @@ def op_grep(pattern: str, path: str = ".", limit: int = 0,
 
     hits = _grep_recursive(pattern, path, limit, excl)
     count = len(hits)
-    file_count = len({h.split(":", 1)[0] for h in hits if ":" in h})
+    file_count = len({fp for fp, _, _ in hits})
 
     out = [f"({count} results in {file_count} files, limit {limit})\n"]
     current_file = ""
-    for hit in hits:
-        # hits are "path:lineno:content" — split on first two colons
-        parts = hit.split(":", 2)
-        if len(parts) >= 3:
-            fp, lineno, content = parts[0], parts[1], parts[2]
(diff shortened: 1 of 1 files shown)


fdaviddpt:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #207, opened 2026-05-25
test: cross-platform fixes for at_file/dispatch/security suites (#188)

fdaviddpt:
test: cross-platform fixes for at_file/dispatch/security suites (#188)

Six distinct Windows-only failures across 5 files:

1. **`test_dispatch_grep` + `test_dispatch_grep_with_context`** — op output normalises paths to forward slashes (#189). Tests asserted `str(f) + "\n" in out` (backslash on Windows) → use `f.as_posix()`.

2. **`test_at_file_route::TestTomlPayload`** (4 tests) — wrote TOML basic-strings like `path = "C:\Users\runneradmin\..."`. TOML treats `\U` as a Unicode escape so `tomllib` fails at column 13 (the `\U` after `C:`). Use `target.as_posix()` in the payload — forward slashes pass through unchanged.

3. **`test_security_devto::test_track_file_is_fixed_path`** — assertion compared a forward-slash substring against `Path.__str__` which is backslash on Windows. Normalise both.

4. **`test_security_hashnode::test_token_resolution_only_reads_fixed_paths`** — `os.path.expanduser` uses `USERPROFILE` on Windows, `HOME` on POSIX. The test set `HOME` only. Set both.

5. **`test_security_edge_cases::test_op_replace_skips_unreadable_binary_peek`** — `chmod 0o000` is a no-op on Windows. Skipif `win32`.

6. **`test_security_edge_cases::test_atomic_write_recovers_from_write_failure`** — test mocks `fdopen` to return a stub that never closes the fd. Windows can't unlink a file with an open handle, so cleanup leaves the tmp file behind. Skipif `win32` — the assertion is correct on POSIX where unlink-while-open works.

POSIX behaviour unchanged.

## Buckets remaining for #188

test_chaos batch_edit_chain (regex/repr nit), test_huge_batch (slow), test_security_publish_149 path tests, plus a few one-offs.

### DIFF
--- a/tests/test_at_file_route.py
+++ b/tests/test_at_file_route.py
@@ -305,8 +305,9 @@ class TestTomlPayload:
     def test_paste_with_triple_literal_preserves_backslash(self, tmp_path: Path) -> None:
         target = tmp_path / "out.sh"
         spec = tmp_path / "p.toml"
+        # as_posix avoids Windows backslashes being interpreted as TOML escapes.
         spec.write_text(
-            f'path = "{target}"\n'
+            f'path = "{target.as_posix()}"\n'
             "content = '''\n"
             "claude -p \"...\" --permission-mode bypass \\\n"
             "  --disallowedTools \"Grep,Glob\"\n"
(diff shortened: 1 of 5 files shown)


fdaviddpt:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Issue #241, opened 2026-05-28
around:/grep_around: large context can over-fetch — add per-op byte cap


fdaviddpt:
## Summary

`around:` / `grep_around:` with a large context value can over-fetch — a single op produced ~340KB of output (auto-persisted to a file, but it still blows the intended context budget).

## Repro

`./supertool 'around:someSymbol:BIG_FILE.class.php:40'` — with a big context (e.g. 40) on a match inside a large file, the op dumps a very large block in one shot. Combined with other ops in the same batch call, total output ballooned to ~340KB.

## Suggested

Add a per-op byte cap (e.g. ~8–16KB) with a clear truncation footer: `… truncated, N more lines — narrow context (:N) or use between:`. This keeps a single op from ballooning the caller's context while still pointing at how to get more. `between:` is the right tool for whole-symbol chunks, so a hint toward it would help.

Env: supertool 0.14.0.

PR #268, opened 2026-05-31
fix(git-checkout): recover in narrowed-refspec (single-branch) workspaces

### LINKED ISSUE #267: git-checkout fails on repos with a narrowed fetch refspec (single-branch workspaces)
## Summary

`git-checkout:<branch>` reports `ERROR: ref '<branch>' not found even after fetch` for branches that demonstrably exist on the remote. Root cause: the op assumes a standard full fetch refspec (`+refs/heads/*:refs/remotes/origin/*`) so that `git fetch` populates `refs/remotes/origin/<branch>`. In workspaces cloned/configured single-branch, the refspec is narrowed to one branch and no `origin/<branch>` tracking ref is ever created — so the lookup fails even though the branch is fetchable.

## Repro

In a repo whose `remote.origin.fetch` is narrowed:

```
$ git config --get-all remote.origin.fetch
+refs/heads/master:refs/remotes/origin/master
```

```
$ ./supertool 'git-checkout:max/test-coverage-monitors-legalstructure-trait'
FAIL
ERROR: ref 'max/test-coverage-monitors-legalstructure-trait' not found even after fetch.
```

Branch exists on remote:

```
$ git ls-remote --heads origin 'max/test-coverage-monitors*'
f2a16cfec4b2...  refs/heads/max/test-coverage-monitors-legalstructure-trait
```

Plain `git fetch origin` says `1 new refs` but only updates `FETCH_HEAD` — `origin/<branch>` is never created, so `git checkout -B <branch> origin/<branch>` also fails.

## Working manual fallback

Explicit single-ref fetch bypasses the refspec:

```
git fetch origin <branch>
git checkout -B <branch> FETCH_HEAD
```

## Proposed fix

When `origin/<branch>` tracking ref is absent after the initial fetch, fall back to an explicit `git fetch origin <branch>` and check out `FETCH_HEAD` (create/reset the local branch to it). Optionally set upstream afterward. This makes `git-checkout` work regardless of refspec width.

## Context

Hit in a DVSI automation worktree (Kevin-style single-branch checkout). These workspaces routinely have a master-only refspec, so any cross-branch op built on `origin/*` tracking refs will trip on this — worth auditing other git ops (e.g. `git-diverge`, `git-merge`) for the same assumption.

fdaviddpt:
## Problem

`git-checkout:<branch>` fails with `ref '<branch>' not found even after fetch` in workspaces whose `remote.origin.fetch` is narrowed to a single branch (e.g. `+refs/heads/master:refs/remotes/origin/master`). The auto-fetch fallback runs `git fetch --all --prune`, which only updates `FETCH_HEAD` — it never creates `refs/remotes/origin/<branch>` — so the retry `git checkout <branch>` still fails.

Hit in DVSI Kevin-style single-branch automation worktrees.

## Fix

When the retry still hits a pathspec error, fall back to the manual workaround:

```
git fetch origin <ref>
git checkout -B <ref> FETCH_HEAD
```

`-B` is safe here: this block only runs after the first plain `git checkout <ref>` already failed, so a pre-existing local branch (which would have made that succeed) is impossible — `-B` only ever creates the branch.

## Audit

`git-diverge` and `git-merge` do **not** share the bug — neither auto-fetches nor assumes `origin/*` tracking refs; both verify a local ref and tell the user to fetch.

## Tests

Added `test_narrowed_refspec_recovers_via_single_ref_fetch`. 5/5 pass.

Closes #267

### DIFF
--- a/presets/git/checkout.py
+++ b/presets/git/checkout.py
@@ -85,6 +85,19 @@ def main() -> int:
                 print(f"ERROR: ref {ref!r} matches multiple remotes: {joined}. "
                       f"Disambiguate, e.g. git checkout -b {ref} --track <remote>/{ref}")
                 return 1
+        # #267: single-branch / narrowed-refspec workspaces never create a
+        # refs/remotes/origin/<branch> tracking ref, so the #277 path above
+        # finds no match and `git fetch --all` only moved FETCH_HEAD. Fall
+        # back to an explicit single-ref fetch + checkout of FETCH_HEAD.
+        if result.returncode != 0 and ("did not match any" in s or "pathspec" in s):
+            single = _git(["fetch", "origin", ref], timeout=30)
+            if single.returncode == 0:
+                cob = _git(["checkout", "-B", ref, "FETCH_HEAD"])
+                if cob.returncode == 0:
+                    print(f"# (fetched origin {ref}, reset local branch to FETCH_HEAD)")
+                    result = cob
+                    stderr = ""
+                    s = ""
     if result.returncode != 0:
         if "not a git repository" in s:
             print("ERROR: not inside a git repository.")
--- a/tests/test_git_checkout.py
+++ b/tests/test_git_checkout.py
@@ -85,6 +85,41 @@ def test_unknown_ref_recovers_after_auto_fetch(monkeypatch, capsys) -> None:
     assert "→ feature" in out
 
 
+def test_narrowed_refspec_recovers_via_single_ref_fetch(monkeypatch, capsys) -> None:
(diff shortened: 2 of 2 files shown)


fdaviddpt:
Rebased onto current master and dropped the bundled #241 byte-cap commits — those were already merged separately, so this PR now contains **only** the #267 git-checkout fix (checkout.py +13, test +35). Also chained the narrowed-refspec fallback *after* the #277 remote-only-branch path that landed in master since this branch was cut, so both recovery cases are handled: #277 creates a tracking branch when a remote ref exists, #267 fetches + resets to FETCH_HEAD when none does. 5/5 checkout tests pass.

PR #331, opened 2026-06-25
feat(git-status): add :full mode to uncap truncated lists (closes #330)

### LINKED ISSUE #330: No selective-stage op + git-status truncates → forced back to raw git for precise commits
## Problem

When a working tree mixes the changes I want with pre-existing untracked junk (this session: a stray `V19` migration, `db-portage` docs, placeholder images that were there at session start), staging *only* my changes has no supertool path:

- `git-status` is great for the ahead/behind + MR overview, but it **truncates** the file list (`... (4 more)`, `... (25 more)`), so I can't drive precise staging from it.
- There's no `stage`/`add` op with an explicit pathspec + exclusion list.

So I dropped to raw `git add <dirs>` + `git reset <junk-file>` + `git status --porcelain | grep -v <junk>` to build a clean 65-file commit. Worked, but it's the one place the whole MR flow left the supertool lane.

## Proposals

1. `git-status:porcelain` (or `:full`) mode that emits the complete machine-readable list (no truncation) for staging decisions.
2. A `git-stage` op taking include pathspecs and an optional exclude list, e.g. `git-stage:::INCLUDE=src2,tests:::EXCLUDE=src2/Dvsi/Migrations/V19/...`, with a printed cached-set summary for self-review before commit.

## Context

The `mr` op already warns about untracked files NOT auto-included — this would close the loop on the staging side so the clean commit can be built without raw git.

— reported by Max after the autonomous-tabs refactor (needed to exclude 5 pre-existing untracked items from a 65-file commit)

fdaviddpt:
## What

`git-status:full` (alias `:porcelain`) prints the complete untruncated dashboard — every staged/unstaged/untracked file, every diverging branch, every stash — instead of the default `... (N more)` caps.

The **default view stays capped** (20 staged/unstaged, 10 untracked/branches, 5 stashes). The truncation is correct for the common case (ahead/behind + dirty + MR overview); the bug was the lack of an escape hatch. Driving precise staging — e.g. excluding a few pre-existing untracked items from a large commit (the exact case in #330) — needs the full machine-readable list, which previously forced a drop back to raw `git status --porcelain`.

## Changes

- `presets/git/status.py` — `full` flag from arg, gates all five caps
- `presets/git.json` — syntax `git-status[:full]` + description
- `docs/presets/git.md` — table row updated
- `CHANGELOG.md` — Unreleased/Added entry
- `tests/test_git_status.py` — 3 new tests (default truncates, `:full` lists all, `:porcelain` alias). 14/14 pass.

## Note on Proposal 2

Issue #330 also proposed a dedicated `git-stage` op with INCLUDE/EXCLUDE pathspecs. Deliberately **not** building it: once `:full` gives the untruncated list, the existing `git-commit:::MSG:::PATH...` takes explicit paths, covering selective staging without a custom exclude mini-language. Re-scoping the issue to Proposal 1 only.


fdaviddpt:
## Review — APPROVE

Independent review (Sonnet sub-agent, read-only).

**Correctness:** All five lists uncapped uniformly under `:full`/`:porcelain` (`(items if full else items[:cap])` + `if not full and len > cap`) — other-branches, staged, unstaged, untracked, stashes. Default path is byte-identical to before; no regression. `{args}` manifest token is the right choice (matches git-diff/git-push; empty when no arg, so bare `git-status` still works). Arg parsing safe: unknown arg → graceful default, `.lower()` handles case, extra args ignored.

**Docs:** consistent across docstring, `git.json` syntax, `docs/presets/git.md` row, CHANGELOG. Cap numbers (20/20/10/10/5) match code everywhere.

**Test gaps (non-blocking follow-ups):**
- Only the untracked list is tested under `:full`. The other four blocks share the identical one-line pattern (low risk) but aren't directly exercised — a single >20-file staged test would lock them in.
- No exactly-at-cap boundary test (`len == cap` → all shown, no marker). Correct by inspection, but an off-by-one would slip through.

Verdict: **APPROVE** — correct, minimal, uniform, documented. Gaps worth a follow-up, don't block.

PR #385, opened 2026-07-26
feat: add read:PATH:START-END range form and a misread-range note (#382)

### LINKED ISSUE #382: read:PATH:A:B is OFFSET:LIMIT and reads like start:end — add a range form or a heuristic warning
## Problem

`read:PATH[:OFFSET:LIMIT]` is documented correctly. It still reads like `start:end` to anyone who has used `sed -n 'A,Bp'`, and I misused it repeatedly across one session before noticing.

Concretely: `read:pipeline/shell.py:352:372` — intent was lines 352 to 372. Actual: offset 352, limit 372, so it returned lines 353 through 521 (EOF). ~170 lines I did not want. I did the same thing three more times in the same session, including `read:scripts/save-session.sh:425:450` which returned 426→494.

Each of those is a few thousand wasted tokens, and worse, the overshoot is quiet — the output looks plausible, just long, so it does not announce the mistake.

## Suggestions

Either would fix it; the first is cleaner.

**a. Accept an explicit range form.** `read:PATH:352-372` for start-to-end, keeping `:352:372` as offset/limit. Unambiguous at the call site.

**b. Heuristic warning.** Emit a one-line note when the arguments look like a misread range — i.e. `OFFSET + LIMIT` overshoots EOF **and** `LIMIT > OFFSET`. Something like:

```
note: read 169 lines (offset 352, limit 372). If you meant lines 352-372, that is a range — see read:PATH:START-END
```

That combination is rare in legitimate use and near-certain in the misread case.

## Priority

Second of four. Pure token cost, no correctness risk, but it recurs — I hit it four times in one session knowing the signature.

fdaviddpt:
Closes #382.

## The problem

`read:PATH:A:B` is documented as `OFFSET:LIMIT` and behaves that way. It still *reads* like `start:end` to anyone who has typed `sed -n 'A,Bp'` — and the failure is quiet. `read:pipeline/shell.py:352:372` returns lines 353→521, not 352→372: the output just looks long, so nothing announces the mistake. The issue reports four occurrences in a single session, by someone who knew the signature.

## The fix

**`read:PATH:START-END`** — an explicit, inclusive range. Unambiguous at the call site, clamps at EOF, and rejects the shapes that can only be mistakes:

- a trailing `LIMIT` (`read:f:3-5:2`) — the range already bounds the read
- `START < 1`
- `END < START`

`:full` / `:raw` still compose with it. `:A:B` keeps its documented meaning, untouched.

**The note.** Where the old form is used and the arguments look like a misread range, one line is appended:

```
note: read 48 lines (offset 52, limit 72) — those args are OFFSET:LIMIT, not
START:END. For lines 52-72, use read:path/to/file:52-72
```

The trigger is deliberately narrow — `LIMIT > OFFSET` **and** `OFFSET + LIMIT` running past EOF. A real limit is seldom larger than the offset it starts from, so that combination is near-certain in the misread case and rare otherwise. Reading a file's tail (`:90:20`) and any window that lands inside the file stay silent, as does an errored read.

## Implementation

Parsing lives in `dispatch` (`_READ_RANGE_RE`, `\d+-\d+`), which converts the range to the offset/limit `op_read` already takes — `op_read` and `render_file` are unchanged. The note is `_read_range_note`, which uses the existing `_count_lines` helper.

## Tests

`tests/test_read_range_382.py`, 14 cases: the range span (multi-line, single-line, EOF clamp), all three error shapes, `:full` composition, the `:A:B` backwards-compatibility case, and five note cases including the exact `52:72` shape from the issue plus every situation that must stay quiet.

Full suite: 3617 passed, 34 skipped, coverage 87.92%.

## Docs

`--help` header, both `.supertool.json` syntax strings, README input-forms line, `docs/input-forms.md`, and the `docs/operations/{reads,index}.md` tables.


fdaviddpt:
Good catch on the grep composition — that was a real bug, and the error message made it worse by naming a `LIMIT` the caller never wrote.

Fixed in the last commit. The filter is now found by **scanning** the trailing slots rather than indexing `parts[4]`, so every spelling lands in the same place:

- `read:PATH:3-25:grep=X` — the natural one, previously an error
- `read:PATH:3-25:::grep=X` — filter in `parts[5]` once a range is present
- `read:PATH:::grep=X` — the documented form, `parts[4]`, unchanged

Three tests added: the range+grep composition, the triple-colon spelling, and one asserting the range still **bounds** the filter (a match before or after the range stays excluded — the filter narrows within the window, it doesn't widen past it). Plus the backwards-compatibility case for the documented spelling, which had no direct coverage before.

Docs updated in both operations tables and the CHANGELOG entry.

On the two nitpicks:

**Oversized-limit idiom** (`read:file:100:9999:grep=TODO` as "from 100 to EOF"). Agreed it's the note's most plausible false positive. Keeping it: the note is one line, advisory, and names the actual window that was read — which is still true information for that call. The range form now gives that idiom a correct spelling anyway (`read:file:100-9999:grep=TODO` clamps at EOF), which is the outcome the note points at.

**Duplicate `_count_lines` at module scope** — confirmed, pre-existing, and the second definition silently wins. Out of scope here, but a real latent bug rather than a style wart, so I've filed it separately as #388 instead of letting it ride in a review thread.

PR #387, opened 2026-07-26
feat: add an append: op so appending stops costing a discovery call (#383)

### LINKED ISSUE #383: Add an append: op — appending currently needs a wc round-trip plus the inverted-range trick
## Problem

Appending to a file today means:

1. `wc:PATH` to learn the line count (a round-trip that exists only to compute an argument)
2. `replace_lines` with `start = N+1`, `end = N` — the END<START insert form

It works, and the inverted-range trick is documented. But appending is common enough (logs, memory files, changelogs, notes) that it should not cost a discovery call, and `start=887 end=886` reads like a typo to anyone reviewing the command later.

Real example from today, appending to a team-memory markdown file:

```
./supertool 'wc:.claude/remember/team/max/2026-07.md'      # -> 886L
./supertool 'replace_lines:@-'                              # start=887 end=886
```

## Suggestion

`append:PATH` taking the same `@payload` route as the other mutating ops:

```
./supertool 'append:@-' <<'EOF'
path = ".claude/remember/team/max/2026-07.md"
content = '''
## New section
...'''
EOF
```

One call, no line arithmetic, and it should run the same post-write validators as `paste`/`edit`.

Worth considering `prepend:` alongside it, though I have not actually wanted that one.

## Priority

Third of four. Convenience, not correctness — but it saves a round-trip every time.

fdaviddpt:
Closes #383.

## The problem

Appending to a file cost two calls:

```
./supertool 'wc:.claude/remember/team/max/2026-07.md'   # -> 886L
./supertool 'replace_lines:@-'                          # start=887 end=886
```

The first exists only to compute an argument for the second, and `start=887 end=886` — the documented inverted-range insert form — reads like a typo to whoever reviews the command later. Appending is common enough (logs, memory files, changelogs, notes) that it shouldn't cost a discovery round-trip.

## The op

```bash
./supertool 'append:::CHANGELOG.md:::## New section'

./supertool 'append:@-' <<'EOF'
path = ".claude/remember/team/max/2026-07.md"
content = '''
## New section
...'''
EOF
```

Same `@payload` route as the other mutating ops, same post-write validators and formatters, creates the file and its parent directories when missing.

**Trailing newline.** A target whose last line has none gets one first, or the block lands on the end of that line. The receipt says so — silently touching a byte the caller didn't ask about is worth a word — and it matches the file's own convention rather than imposing LF on a CRLF file.

**CRLF.** The file is read with `newline=""`. Reading in text mode translates CRLF to LF, and since the op rewrites the whole file atomically, that would normalise **every line ending in the file** as a side effect of adding one line at the end. Caught by a test, not by inspection.

**Receipt cap.** Two lines of preceding context, then the block, capped at 10 lines with `… (+N more appended lines)`. `append` is the op you reach for with a long changelog entry; echoing it back in full is pure token cost on content the caller already had.

## `hooks_into` — read this before upgrading

`hooks_into` is an explicit list, so every shipped validator, formatter and notifier spec in `.supertool.json` and `.supertool.example.json` gained `"append"` (36 arrays).

**A project config that enumerates its own ops will not validate appended content until `"append"` is added to its `hooks_into` arrays.** Worth doing in the same pass as the upgrade — a mutating op that quietly skips validation is worse than no op at all.

I chose explicit declaration over silently treating `append` as an alias of `paste` at match time: the config contract is that `hooks_into` means what it says, and magic op-aliasing would be invisible in exactly the place someone goes looking.

## `prepend`

The issue mentions it as worth considering. Left out — the issue itself says it was never actually wanted, and an op nobody reaches for is still an op to document, test and keep working.

## Tests

`tests/test_append_op_383.py`, 21 cases: block placement, missing trailing newline on the content and on the target (LF and CRLF), file/parent creation, byte-for-byte preservation of non-UTF-8 bytes and CRLF endings, receipt context and cap, the four error shapes, containment checked **before** `makedirs` (or the directories land outside cwd even though the write is rejected), and the colon-CLI / colons-in-content / `@payload` dispatch routes.

One existing test updated: `test_builtin_defaults_cover_all_write_ops` pins the `@file` registry exactly, so it gains the `append` entry.

Full suite: 3624 passed, 34 skipped, coverage 87.88%.

## Docs

Both config files, the `docs/operations/{edits,index}.md` tables, and every prose list that enumerates the mutating ops (README ×2, input-forms, validators ×2, formatters, notifiers ×2, cursor-witness).


fdaviddpt:
Right on the cursor-witness gap — and worse than a missed enumeration: I *edited the doc* to claim `append` gets the side-by-side diff, which made the PR assert something false. The Python notifier fires correctly; the extension gates on `MUTATING_OPS` and dropped the event entirely, so there was no status bar, no diff, and no `{before_file}` handling.

Fixed in the last commit:

- `extension/src/extension.ts` and the compiled `extension/out/extension.js` — `MUTATING_OPS` now lists `append`
- `notify.py` docstring and `install.sh` setup snippet — the `hooks_into` arrays a user copies into their own config

Also took the three nitpick loops, since they're the exact "enumerated elsewhere" spots that go stale silently:

- `test_validators.py::test_default_ops_cover_all_mutating`
- `test_at_file_route.py::test_registry_populated_after_first_dispatch`
- `test_parallel.py::test_parallel_safe_excludes_mutating_ops` — this one was also missing `paste` and `vim`, so it now exercises `_is_parallel_safe()` for all six rather than four

And the symlink case, which `op_paste` covers and `op_append` didn't: append through a symlink writes to the real target and leaves the link intact.

On the `os.path.isdir` divergence from `op_paste` — intentional, now commented at the guard. `append` is the op most likely to be aimed at a directory by mistake (a `logs/` path with the filename left off), and "is a directory" beats an mkstemp errno for saying so.

Full suite: 3625 passed, 34 skipped, coverage 87.90%.

PR #389, opened 2026-07-26
feat: echo the branch on mutating ops, and diagnose a failed edit (#381, #380)

### LINKED ISSUE #380: DX feedback from a long autonomous session: edit-failure diagnostics, doubled-backslash trap, cwd discovery
Feedback from a long autonomous session (~18h, several hundred supertool calls across a Python/bash repo). No confirmed bugs in the tool — these are DX papercuts, ranked by how much time each cost me.

## 1. `edit` failure gives no diagnostic (cost me the most)

`ERROR: old string not found in <path>` is the whole message. When the payload is close but not exact — one character of whitespace, an escaping discrepancy — there is nothing to work from, and the natural next move is a `read` round-trip that the payload route was supposed to save.

**Suggestion:** on a no-match, print the closest line(s) by similarity, or the longest common prefix of `old` against the file. Even one line of "nearest match at line 142: `...`" turns a three-call debug loop into one.

This bit me ~5 times in one session, always the same shape: TOML literal strings do not process escapes, so `\\302` in a payload is a literal backslash-backslash-3-0-2 and never matches a file containing `\302`. Which leads to:

## 2. Doubled backslashes silently produce broken shell that validators accept

The docs correctly say TOML `'''...'''` preserves backslashes. The trap is that the *result* of getting it wrong is not always a failed match — sometimes it writes. A `\\` at end-of-line in bash is an escaped backslash, not a line continuation, so:

```bash
FOO=$(cmd \\
    --arg)
```

is syntactically valid, passes `bash -n`, passes the `bash-check` validator, and does something entirely different from what was intended. I shipped this into a file and only caught it by reading the bytes back with `repr()`.

**Suggestion:** for `.sh` targets, warn when a written hunk contains `\\` immediately before a newline. It is almost never intentional, and the current validator cannot see it.

## 3. `./supertool` only resolves from the project root

Working in a second checkout, every call needs an absolute path or it fails with `no such file or directory`. `cwd:PATH` exists and solves this — I did not discover it until late, because the failure message says nothing about it.

**Suggestion:** when `./supertool` is invoked and not found relative to cwd, or when a path argument lies outside the project root, mention `cwd:` in the error.

## 4. Validator wording

`bash-check : ok (unchanged)` reads as "the file is unchanged". It means the validator's *result* did not change. Minor, but I double-took on it several times.

## What worked well, so it does not get changed by accident

- `@-` heredoc for single edits is the right default — hook-safe, no temp file, no escaping layer beyond TOML's.
- Post-edit validators with rollback caught real syntax errors several times before they reached a commit.
- Batched reads genuinely collapse round-trips; the honest cost model in the docs (10% cached re-pay, not free) is the reason I trusted it.

### LINKED ISSUE #381: Edit ops should echo the current git branch
## Problem

Two near-misses in one session today, both the same shape: **right file, wrong branch.**

1. Edited `README.md` and `config.example.json` in a checkout that was sitting on a feature branch instead of `main`. Caught it only because I ran `git status` before committing — otherwise unrelated docs would have been swept into an open PR.
2. Tried to edit `scripts/save-session.sh` against a branch that predated a merge. Found out via `ERROR: old string not found`, which is a confusing symptom for "you are on the wrong branch" — my first assumption was that I had the anchor text wrong, and I re-read the file to check.

Neither is supertool's fault. But supertool is the thing that knows, and the fix is nearly free.

## Suggestion

Print the current branch in the footer of every mutating op (`edit`, `replace`, `replace_lines`, `paste`, `vim`, and `batch` once per call):

```
edited scripts/save-session.sh (line 98-127)  [branch: main]
```

supertool already has git awareness — `git-status` is an op — so the information is on hand. A handful of tokens per call, against a class of mistake that is silent until commit time.

Bonus, if cheap: when `edit` fails with `old string not found`, append the branch to the error. That is the exact moment the user is forming a wrong hypothesis about why it failed.

## Priority

Highest of the four I am filing. The other three cost tokens; this one costs correctness.

fdaviddpt:
Closes #381. Closes #380.

## Right file, wrong branch

Two near-misses in one session, both the same shape.

One edited `README.md` and `config.example.json` in a checkout sitting on a feature branch instead of `main` — caught only by running `git status` before committing, which is the difference between "unrelated docs swept into an open PR" and not. The other surfaced as `ERROR: old string not found`, which is a confusing symptom for *you are on the wrong branch*: the first hypothesis is a bad anchor, and the next move is a `read` round-trip to check it.

Neither is supertool's fault. But supertool is the thing that knows.

Every mutating op — `edit`, `replace`, `replace_lines`, `paste`, `vim` — now closes with:

```
edited scripts/save-session.sh (line 98-127)
[branch: main]
```

**On failure as well as success**, which is the issue's bonus ask and the more valuable half: a failed edit is exactly when a wrong-branch hypothesis should already be on the table rather than reached for after re-reading the file.

Implementation notes:

- `git symbolic-ref`, not `rev-parse` — it resolves an **unborn** branch (fresh `git init`, no commit yet), where `rev-parse` fails with `ambiguous argument 'HEAD'`. Detached HEAD falls through to a second call and reports the short sha.
- Cached for the process. A batch of edits pays one subprocess, not one each.
- Any failure — no repo, no git, timeout — yields **no footer**, never an error. The branch is a convenience on a receipt; it must never be the thing that fails an op.
- Added at the single dispatch chokepoint, keyed on `_OP_TARGETS`, so no op function had to learn about git.

## Why the edit failed

`ERROR: old string not found in <path>` was the whole message, so the natural next move was a `read` round-trip — the one the payload route exists to save. Three diagnostics now follow it, in the order they were actually the cause:

**1. Doubled backslashes.** The one that bit ~5 times in a single session. TOML literal strings don't process escapes, so `\\302` in a payload is backslash-backslash-3-0-2 and can never match a file holding `\302`. When collapsing them makes the anchor match, the receipt says so outright:

```
ERROR: old string not found in x.py
  ↳ the file matches with SINGLE backslashes — TOML literal strings ('''...''')
    do not process escapes, so `\\` in the payload is two literal backslashes
```

**2. Whitespace.** A single-line anchor matching some line once whitespace is normalised gets that line number and `check indentation`.

**3. Nearest line.** Otherwise the closest line by similarity, with number and score, when it clears 60%:

```
  ↳ nearest match at line 2 (92%): '    return 1'
```

A wild miss gets **nothing** rather than an invented suggestion. The scan is skipped above 20 000 lines — this is a courtesy on a failure path and must never become the slow part of a failed edit. The backslash check is a substring test, so it still fires on a huge file.

The ambiguous-match error (`found N times`) is untouched: it's a different failure and doesn't get a "not found" hint.

## Not in this PR

The other three items in #380 — the `\\`-before-newline warning on `.sh` writes, the `cwd:` mention in path-not-found errors, and the `bash-check : ok (unchanged)` wording — are a separate concern from edit diagnostics and are coming in their own PR, so #380 stays open until that lands despite the `Closes` above. Happy to drop the keyword if you'd rather close it manually.

## Tests

`tests/test_edit_diagnostics_381_380.py`, 21 cases.

Branch: resolution across checked-out / unborn / detached / non-repo, and the per-process cache (asserted by counting `git` subprocess calls). Footer: on every mutating op, on a failed edit, absent on read ops, absent — rather than `[branch: ]` — when there's no branch.

Diagnostics: each of the three, the wild miss that stays quiet, the size cap, that the backslash hint survives the cap, and that ambiguous-match and success both stay clean.

Every git fixture is built inside `tmp_path` with `git -C tmp_path`, so `conftest._guard_repo_git_state` stays satisfied.

Full suite: 3629 passed, 34 skipped, coverage 87.91%.


fdaviddpt:
All three land. Fixed in the last commit.

**1. Batch footer spam — the worst of the three.** You're right that the issue said "batch once per call" and I gated only on `op in _OP_TARGETS`, which every sub-op satisfies on its way through the recursive `dispatch`. A 50-edit batch printed 50 identical branch lines — precisely inverting the "handful of tokens per call" the feature is justified by.

Now gated on `_DISPATCH_STATE.depth`, which already exists for the recursion guard, so sub-ops stay silent and the batch itself carries one footer. It emits only when the batch actually ran a mutating sub-op, so a read-only batch has none — there's no branch worth warning about when nothing was written.

**2. Plain-mode glyph.** Correct, and the contract is explicit about it. `↳` is now in `_PLAIN_MARKERS` as `->` and both diagnostic sites go through `mark()`. That also fixes the pre-existing unmarked `↳` in the read-edit hint at `_read_edit_hint` — same class, one line away, so I took it rather than leaving a known-broken sibling.

Test added asserting `SUPERTOOL_PLAIN=1` yields `-> nearest match` and no `↳`.

One thing I did **not** change: the em dash in the hint prose. Plain mode's `_PLAIN_MARKERS` covers the status glyphs used for grepping, and existing emitted messages already carry em dashes (`(pre-existing — not from this edit)`). Making my two strings ASCII-only while the sibling column stays multibyte would be inconsistent without actually making the output greppable. If you want plain mode to mean strictly-ASCII output, that's a real change worth its own pass over every message — happy to file it.

**3. `append`.** This one resolved itself in the rebase: #387 landed after this branch was cut, and it added `append` to `_OP_TARGETS` — which *is* the footer gate. So it was correct as soon as the branches met. I've added an explicit test pinning it, plus one asserting `replace_dry` stays footer-free, since you flagged that as easy to break later.

Also confirmed your reading of the `continue_on_error` check: it inspects `_sub_result.split("\n")[1]`, the line right after the header, so a trailing footer can't affect error detection. It's now moot anyway, since sub-ops no longer carry one.

Full suite: 3674 passed, 34 skipped, coverage 88.01%.

PR #390, opened 2026-07-26
feat: rebuild the header for content-heavy ops, and three DX fixes (#384, #380)

### LINKED ISSUE #384: Truncate the op header on batch edits — full old+new strings are echoed twice
## Problem

Every mutating op prints its full arguments in the section header, then prints the diff underneath. For `edit` with substantial content, the old and new strings appear **twice** in the output.

On a 6-op `batch:@-` today — several of the edits carrying 10-20 line code blocks with long explanatory comments — the header lines alone were a large fraction of the response. The header looked like:

```
--- edit:    staging_contents: dict[str, str] = {}
    for path in sorted(globmod.glob(os.path.join(staging_dir, "today-*.md"))):
        ... 20 more lines of OLD ...
:    staging_contents: dict[str, str] = {}
    # Raw file size, captured at read time. ...
        ... 20 more lines of NEW ...
:pipeline/shell.py ---
edited pipeline/shell.py (line 358-376)
    [diff follows, showing the same new content again]
```

The diff below is the useful part and already shows what changed.

## Suggestion

Truncate the op header to something identifying — first ~60 chars of the anchor, plus the path:

```
--- edit: "    staging_contents: dict[str, str] = {}…" -> pipeline/shell.py ---
```

Roughly halves the output of a content-heavy batch with no information lost. If the full arguments are wanted for debugging, a `:verbose` suffix would cover it.

## Priority

Fourth of four. Pure token cost, and it scales with exactly the usage pattern supertool encourages — big batched edits.

fdaviddpt:
Closes #384. Also takes the three remaining items from #380 (the fourth, edit-miss diagnostics, is in #389).

## The header echoed what the diff was about to show

Every mutating op prints its full arguments in the section header, then the diff underneath. For a content-heavy `edit` the old and new strings therefore appear **twice** — and the header comes *first*, before the diff that actually shows the change.

On a 6-op `batch` carrying 10-20 line code blocks, the headers alone were a large fraction of the response. The cost scales with exactly the usage pattern supertool encourages.

Above 160 characters the header is rebuilt from the parsed fields, keeping whatever identifies the **target** and dropping the content:

```
--- edit: "    staging_contents: dict[str, str] = {}… (+609 chars)" → pipeline/shell.py ---
```

Per op: `replace_lines` keeps its line range, `paste`/`append` keep the path and a character count, `vim` keeps its script, `edit`/`replace` keep the anchor and path.

Two deliberate details:

- **The elided count is reported** (`… (+609 chars)`), not trailed off. A silent truncation reads as "that was the whole argument".
- **A multi-line anchor is folded to one line** (`⏎`). Previously a 20-line anchor produced a 20-line header.

Anything under 160 characters keeps its verbatim header, so ordinary short ops are byte-identical to before — which is also why the existing header assertions across the suite still hold.

The `@file` route was already compact (`edit:@p.toml`) and stays that way. The batch sub-op case is the one from the issue: `_sub_arg = ":".join(_sub_parts)` builds the long arg, and that's what now gets rebuilt.

### On `:verbose`

The issue suggests a `:verbose` suffix for the full arguments. I left it out. The arguments are what the caller just sent — they already have them — and the receipt plus diff is the new information. Adding a flag to re-print known content seemed like surface for a case that hasn't come up. Easy to add later if it does.

## #380 — three papercuts

**1. `\\` at end of line in a shell script.** The one with teeth: getting the escaping wrong doesn't always fail to match, sometimes it *writes*. In bash, `FOO=$(cmd \\` followed by a newline is an escaped backslash and then a new command, not a line continuation. It is syntactically valid, passes `bash -n`, passes the `bash-check` validator, and does something entirely different from what was meant. No syntax validator can see it — only the byte pattern can.

Checked at `_atomic_write`, the one place every mutating op passes through, scoped to `.sh`/`.bash`/`.zsh`/`.ksh`. A trailing `\\` is ordinary in Python, C or JSON, and a *single* trailing backslash — the real continuation — stays quiet.

Worth noting where it actually fires: the `@payload` route, where TOML literal strings preserve backslashes verbatim. On the colon CLI, `_decode_escapes` collapses them before the write, so the pattern never reaches the file. That matches the report, which hit it through a payload.

**2. `(unchanged)` wording.** The column reports the delta in the *validator's own result*, but it reads as "the file is unchanged" — the one thing that is definitely false right after a mutating op. Now `(no new errors)`, matching the `(pre-existing — not from this edit)` sibling it sits beside. Four existing tests updated.

**3. `path not found`.** Now ends with `Prefix the call with cwd:PATH to run it from elsewhere.` The failure message is where someone learns the op exists — in the report, `cwd:` went undiscovered until late in a long session precisely because nothing pointed at it.

## Tests

`tests/test_compact_header_384.py`, 15 cases: short and read-op headers stay verbatim; long `edit`/`paste`/`replace_lines` keep their target and drop their content; the elided count is reported; a multi-line anchor becomes one line; the batch sub-op case from the issue; the shell-backslash warning firing plus its four negatives (single backslash, `.py` file, non-shell suffix, and no leak into the next dispatch); and the `cwd:` mention.

Full suite: 3641 passed, 34 skipped, coverage 87.99%.


fdaviddpt:
All three are real. Fixed in the last commit, and issue 1 in particular was a genuine miss on my side — I justified the truncation with "the diff below already shows what changed" without checking that the diff *doesn't render on failure*.

**1. Verbatim header on failure.** The compact form is now computed where it was (while `parts` is in hand) but swapped in at the very end, only when `body` isn't an ERROR. So a failed long `edit` keeps its full arguments, and a successful one gets the short header. That also covers a failed `paste`/`replace_lines`/`vim` — the rule is "nothing was written, so nothing below the header shows what was sent", not "edit specifically".

Two tests added: a long failing `edit` asserting the header is byte-identical to the arg and still contains line 19 of the anchor, and a `paste` to a directory for the non-edit failure path.

**2. Backslash parity.** Correct, and the `.search()` framing is what made it wrong — I was testing for the *substring* `\\` rather than the *shape* of the run. A trailing run of three is one escaped backslash plus a real continuation, i.e. working code, and it was flagged.

Now `(\\+)\r?\n` with a `len(run) % 2 == 0` check. Parametrised test over runs of 1-5: quiet, warn, quiet, warn, quiet.

I also dropped the `[ \t]*` I had allowed between the backslashes and the newline. Trailing whitespace after a backslash breaks continuation in bash regardless of parity, so folding it into this check conflated two different problems.

**3. Rollback leaving a stale warning.** Confirmed — both rollback paths write `pre_content` with a raw `open(path, "wb")`, so they never reach the `_atomic_write` chokepoint where the queue lives. The warning outlived the bytes it described.

The queue is now `(abspath, message)` and both rollback sites call `_drop_write_warnings(path)`. `_atomic_write` also drops any existing entry for a path before re-queuing, so the invariant is "the queue describes what is on disk **now**" rather than "every write that happened" — which fixes the fix-then-rewrite case too, not just rollback. Two tests pin both.

**Nitpicks:** the `append` branch is live after rebasing onto master (#387 landed in between), so it's no longer vestigial. Column width evened up.

Full suite: 3698 passed, 34 skipped, coverage 88.08%.

fdaviddpt:
The `op_replace` zero-match case is a good catch, and it shows my fix was the right shape done the wrong way — I replaced a bad assumption ("the diff shows it") with a bad proxy ("ERROR means nothing was written") instead of asking the real question.

**Now: a write counter.** `_atomic_write` bumps `_WRITE_COUNT`, and dispatch compares it around the op. The header is elided only when bytes actually reached disk.

That also surfaced a second case the prefix test got wrong in the *other* direction, which my own test caught on the first run: `paste` to a directory raises inside `_atomic_write`, and I had put the increment at the top of the function — so a write that never happened counted as one. The counter and the warning queue both moved past the `os.replace`, since both are claims about what is on disk.

Covered now:

- `replace` zero-match with a long `old` → verbatim header, `old` intact
- `replace_dry` → verbatim, always (a preview writes nothing, so its arguments are all the caller has)
- successful `replace` → compact, as the counterpart

**Trailing whitespace.** You're right that the old regex caught `\\  \n` and the parity fix dropped it, and right that the argument points toward catching *more*. It now has its own warning at any parity: a backslash followed by spaces or tabs never continues the line — the escape applies to the space — so the line ends there whatever was intended. That covers the odd-run case (`cmd \ \n`) too, which neither the old nor the new regex caught and which is the more common typo of the two.

Two messages, since they're different mistakes: parity means "you wrote an escaped backslash", whitespace means "you wrote a continuation and killed it with an invisible character".

**Scope note** — agreed the branch is carrying #383/#381/#380 features; that's the rebase onto master as those landed, not new work in this PR. The diff against master should be clean now that they're merged.

Full suite: 3703 passed, 34 skipped, coverage 88.10%.

fdaviddpt:
Took the follow-up rather than filing it — the branch needed a rebase onto master anyway (#391 landed), so it was the same CI cycle either way, and it's the same reproducibility gap as the original finding reached by a different route.

`_retract_write(path)` now replaces the bare `_drop_write_warnings(path)` at both rollback sites: it drops the warning **and** decrements `_WRITE_COUNT`, since both are claims about what is on disk. A rolled-back edit therefore keeps its verbatim header, which is what you'd want — the change didn't stick, so there's no diff to read the arguments from.

Guarded against going negative, with a test for that and one for the retraction itself.

Everything else in your verification matches my reading, including the two quirks you flagged as cosmetic:

- an even run *with* trailing whitespace reports via the whitespace message rather than the parity one — both are true of that line, and the whitespace one is the more actionable of the two (it names the invisible character)
- a file containing both kinds reports one message — I'd rather keep a single line per file than enumerate reasons in a warning that already earns its place by being rare

Full suite: 3713 passed, 34 skipped, coverage 88.26%.

PR #391, opened 2026-07-26
fix: one _count_lines, with the error value as a parameter (#388)

### LINKED ISSUE #388: `_count_lines` is defined twice at module scope — the second silently wins and inverts the first's error contract
## The duplication

`supertool.py` defines `_count_lines` twice at module scope:

**1307** — the documented one, added for the auto-read line cap (#362):

```python
def _count_lines(path: str) -> int:
    """Count lines in a file cheaply, streaming in binary (#362).

    Used by the glob:/grep: auto-read line cap so a file under the byte cap but
    with many lines isn't fully dumped. Returns a large sentinel on read error
    so a file we can't measure is treated as over-cap (skip auto-read)."""
    ...
    except OSError:
        return MAX_AUTOREAD_LINES + 1
```

**2605** — a later, simpler one:

```python
def _count_lines(path: str) -> int:
    """Count lines in a file (fast, doesn't read into memory if big)."""
    try:
        with open(path, "rb") as f:
            return sum(1 for _ in f)
    except OSError:
        return 0
```

Python binds the last definition, so **every call in the file — including the auto-read cap the first one was written for — reaches the second.** The first is dead code that still reads as authoritative, docstring and all.

## Why it matters

The happy paths agree: both count lines equivalently for a readable file, including a trailing line with no final newline (`sum(1 for _ in f)` yields it).

The error path is inverted, and that's the whole point of the first version. Its docstring states the contract: *"Returns a large sentinel on read error so a file we can't measure is treated as over-cap (skip auto-read)."* The definition that actually runs returns `0` — **under** the cap — so an unreadable file is treated as a small file and auto-read is attempted rather than skipped. Exactly backwards from what the comment promises, and the comment is what anyone reading the cap logic will trust.

## Fix

Delete one. Keeping the `MAX_AUTOREAD_LINES + 1` sentinel version (1307) restores the documented behaviour; it needs a check that no current caller relies on `0`-on-error, since that has been the live behaviour for a while and something may have grown around it.

Worth a test either way — the surviving function's OSError branch is the part that differs, and nothing currently pins it.

## How it surfaced

Noticed while reviewing #385, which added a caller (`_read_range_note`). That caller happens to be safe under either definition — it already suppresses its output when the count is `<= 0` — but "safe by accident of which duplicate won" is the same shape as #379, and not a property to leave in place.

fdaviddpt:
Closes #388. Found while reviewing #385.

## The bug

`_count_lines` was defined twice at module scope. Python binds the last one, so **every** call reached the second — including the glob/grep auto-read line cap that the *first* one was written for.

Their error paths are opposites, and that's the whole problem:

```python
# supertool.py:1307 — the documented one
except OSError:
    return MAX_AUTOREAD_LINES + 1   # unmeasurable -> over-cap -> skip auto-read

# supertool.py:2605 — the one that actually ran
except OSError:
    return 0                        # unmeasurable -> under-cap -> auto-read it
```

So an unreadable file was treated as small and auto-read, which is backwards from what the cap exists to do. The docstring stating the correct contract sat right above dead code, reading as authoritative, for months.

## The fix

One definition. Because the callers genuinely want opposite answers, the error value is a parameter rather than a constant somebody has to guess right:

- the auto-read line cap (two call sites) passes `on_error=MAX_AUTOREAD_LINES + 1` and **fails closed**
- `map` and the workspace summary keep the default `0` — a listing wants a truthful zero, not a sentinel rendered as a wildly wrong line count

That's also why I didn't simply delete one: whichever I kept, one caller would have been silently wrong. The existing `test_map.py` assertion that an unreadable file counts as `0` stays green precisely because that behaviour was correct *for map*.

The surviving implementation is the chunked binary reader — cheaper on large files, and it counts a trailing line with no final newline the same way.

## Tests

`tests/test_count_lines_388.py`, 8 cases: counting with and without a trailing newline, the empty file, a count spanning the 64KB chunk boundary, both error values, permission-denied as a second unreadable case, and a guard asserting only one module-scope definition survives — the shape of the original bug.

Full suite: 3656 passed, 34 skipped, coverage 88.10%.


fdaviddpt:
Thanks for the equivalence trace across the chunk boundary and the `\r` / NUL cases — that was the part I most wanted a second pair of eyes on, since a silent off-by-one in `map`'s line counts would have been invisible.

Took the nitpick: `test_only_one_definition_survives` now parses the module and counts module-level `FunctionDef` nodes named `_count_lines`, rather than grepping the source text. You're right that the string form would miss an indented or nested duplicate — which is a plausible next version of this same bug — and could false-positive on a docstring.

Separately, Windows CI caught a bug of mine in the same file: all four Windows jobs errored **during collection** with `module 'os' has no attribute 'geteuid'`. A `skipif` condition is evaluated at collection time, so `os.geteuid() == 0` was evaluated on Windows regardless of the `os.name == "nt"` skipif sitting on the line above it — the guard I thought was protecting it could never run first. `getattr(os, "geteuid", lambda: 1)()` keeps both the root check and the platform. Pushed with the nitpick fix.

PR #395, opened 2026-07-27
fix: nested-batch branch footer, formatter repo opt-in, TOML ''' hint

### LINKED ISSUE #392: A batch nested inside a batch loses the `[branch: X]` footer entirely
Follow-up to #381 / #389. Found while verifying that PR, not a regression it introduced — nested batches were outside its scope.

## Behaviour

The footer is emitted once per call, gated on dispatch depth, with the batch itself carrying it when a mutating sub-op ran:

```python
if getattr(_DISPATCH_STATE, "depth", 1) <= 1:
    if op in _OP_TARGETS or (op == "batch" and _batch_mutated):
        body += _branch_line()
```

`_batch_mutated` is set in the batch loop when `_sub_op in _OP_TARGETS`. `"batch"` is never in `_OP_TARGETS`, so a mutation buried in an *inner* batch never propagates the flag to the outer one.

Outer batch → inner batch → one `edit`:

- the inner batch's own footer is suppressed (depth 2)
- the outer batch's `_batch_mutated` stays `False`, because its only sub-op was `"batch"`

Result: **zero** footers. A file is edited and no branch is reported at all — which is the wrong-branch miss #381 exists to close, in the one shape where nothing warns.

Not double-counting, which is the failure the depth gate was added to prevent. Under-counting to zero.

## Fix

Two options, roughly equal:

**a.** Propagate the flag — treat a `"batch"` sub-op as mutating if its own result carried a mutation. Needs the inner call to report that upward, which the current string return value doesn't do.

**b.** Move the signal to the write chokepoint. `_atomic_write` already knows something was written; a module-level "did this invocation mutate anything" flag, drained at depth 1, makes the answer independent of how deeply the op was nested. That is the same shape as the `_WRITE_WARNINGS` queue added in #390, so the two could share a mechanism.

(b) is the more honest one — nesting depth shouldn't be what decides whether a write gets reported.

## Priority

Low. A batch inside a batch is a rare construct, and the failure is a missing advisory line rather than a wrong result. Worth fixing because the current logic reads as if it handles the case.

### LINKED ISSUE #393: Deferred formatters reformat files the repo never formats
Hit while editing `claude-remember` (a repo with **no** prettier config) from a DVSI shell.

## What happened

A one-line `edit` to `CHANGELOG.md` and another to `README.md` came back with:

```
[validators]
prettier-check: 1 err       (pre-existing — not from this edit)
--- formatters (deferred) ---
  CHANGELOG.md   prettier: ok  (247ms) +22 -22
  README.md      prettier: ok  (163ms) +42 -42
```

126 lines reformatted across two files, none of them mine. The validator itself said *"pre-existing — not from this edit"* and the formatter reformatted anyway.

The repo has no `.prettierrc`, no `prettier` in any manifest, and its own CI does not run prettier. The reformat came entirely from the DVSI `.supertool.json` policy, because that was my cwd.

## Why it matters

The diff is the deliverable. A 2-line change that arrives as 128 changed lines is unreviewable, and in a repo where markdown tables are hand-aligned it is also just wrong. I had to `git checkout` both files and redo the edits with `cwd:` pointed into the target repo so its (absent) policy applied.

## Suggestions, roughly in order of appetite

1. **Do not format a file the edit did not touch the formatting of** — or at minimum, do not format when the pre-edit state already fails the check. "It was already unformatted" is the signal that this repo does not format.
2. **Require evidence the repo opts in.** A formatter runs only if the repo has its config (`.prettierrc`, `[tool.black]`, `.php-cs-fixer.php`, …). Absent that, validate but never rewrite.
3. Failing both, **format only the changed hunk**, not the file.

## Related

`cwd:<other repo>` was the right workaround and I found it by accident. Worth documenting that `cwd:` also switches which `.supertool.json` policy applies — that is a genuinely useful property and not obvious from the op's description ("Set working dir for the whole call").

### LINKED ISSUE #394: TOML payload has no escape hatch for content containing '''
The `@-` / `@file` payload route uses `'''triple-single-quoted'''` for multi-line content, which is the right default — it takes code with zero escaping. But there is no way to pass content that itself contains `'''`.

## Repro

Editing a Python test to add a helper whose body checks for docstring markers:

```
./supertool 'cwd:/path/to/repo' 'edit:@-' <<'PAYLOAD'
path = "tests/test_x.py"
old = '''...'''
new = '''            if stripped.startswith(("#", '"""', "'''", "*")):
                continue'''
PAYLOAD
```

```
ERROR: @file TOML parse error (<stdin>): Expected newline or end of document after a statement (at line 9, column 53)
```

The `"'''"` inside the content terminates the block. There is no documented alternative, and the failure mode is a parse error pointing at a column in the payload rather than at the cause.

Python source that inspects Python source is not exotic — any test that reasons about docstrings hits this.

## What I did instead

Fell back to a `python3 - <<'PY'` heredoc doing a `str.replace`, which works but throws away everything supertool's edit route provides: no single-occurrence check, no validators, no rollback on syntax failure.

## Suggestions

1. **Accept the JSON payload shape for these cases and say so in the error.** The docs mention auto-detection (`starts with { → JSON, else TOML`), so the capability may already exist — the parse error just needs to point at it: *"content contains `'''`; use the JSON payload form"*.
2. Or support a **heredoc-style sentinel**: `new = <<<EOF_MAX ... EOF_MAX`, delimiter chosen by the caller, which is how every other tool solves this.
3. Cheapest: document the collision in `help edit` alongside the existing "TOML preserves backslashes/quotes/newlines" line, which currently reads as though nothing needs escaping.

## Smaller note from the same session

The near-match hint on a failed `edit` is genuinely good:

```
ERROR: old string not found in .claude-plugin/plugin.json
  ↳ nearest match at line 4 (98%): '  "version": "0.8.7",'
```

That turned a two-minute hunt into a two-second one — I had a stray character in the anchor. Please keep it.

fdaviddpt:
Three fixes, all filed from real sessions.

**#392 — nested batch reported no branch at all.** The `[branch: X]` footer is once per call: sub-ops suppressed by dispatch depth, the batch itself carrying the line when a sub-op mutated. That test was a flag set when the sub-op was in `_OP_TARGETS`, and `"batch"` never is — so outer batch → inner batch → edit printed **zero** footers, which is the wrong-branch miss #381 exists to close in the one shape where nothing warns. The signal moved to `_WRITE_COUNT`, bumped past the rename in `_atomic_write`, which every mutating op passes through however deeply nested. `_batch_mutated` is gone.

**#393 — formatters rewrote files in repos that never format.** A one-line edit to a repo with no prettier config came back with 126 reformatted lines, because the policy came from the shell's cwd. A formatter now runs only on evidence: the tool's config file searched from the *edited file's* directory up to **its** repo root (a parent repo's config never opts a nested one in), a manifest naming the tool, or an `env` key carrying the rules (`PHPCBF_STANDARD=PSR12` with no `phpcs.xml` is a legitimate opt-in). Unknown tools are never gated. `requires_config: false` forces always-run; `SUPERTOOL_FORMAT_WITHOUT_CONFIG=1` restores the old behaviour. A gated formatter is **reported** on the receipt rather than skipped silently.

**#394 — content containing `'''` had no escape hatch, and the error hid the two that existed.** The parse error pointed at where the block closed, not at what closed it. It now names them: a `"""basic"""` block or the JSON payload form. Fires on an odd number of `'''` runs, silent otherwise.

### Tests

3731 passed, 34 skipped. New: `test_formatter_optin_393` (10), `test_toml_delimiter_hint_394` (7), plus the nested-batch regression in `test_edit_diagnostics_381_380`.

Seven pre-existing formatter tests changed behaviour under the #393 gate and were updated to state the new contract — three now create the tool's real config marker, four opt out with `requires_config: false` because the glob/hook filtering is what they test. Tightened the name matcher along the way: a spec called `fmt` was matching the `rustfmt` marker table by substring.

Closes #392
Closes #393
Closes #394


fdaviddpt:
### Code review

Found 2 issues:

1. The batch footer now regresses on failure. Switching the test from `_batch_mutated` to `_WRITE_COUNT[0] > _writes_before` means a batch whose only mutating sub-op *fails* gets no `[branch: X]` footer — `_WRITE_COUNT` only increments past `_atomic_write`'s rename. Reproduced head-to-head: a batch with one failing `edit` prints the footer on `master` and prints nothing on this branch. Single ops are unaffected (`op in _OP_TARGETS` still fires). This contradicts this PR's own CHANGELOG line ("A failed mutating op still gets the footer, since that is exactly when a wrong-branch hypothesis is worth having") and the guarantee #389 established. The new test `test_nested_batch_still_reports_the_branch` only covers the successful path, so the gap ships untested.

https://github.com/Digital-Process-Tools/claude-supertool/blob/bb91ac3d8f4e7def78a7e80600de82a0b624cef1/supertool.py#L12492-L12496

2. The new `_toml_delimiter_hint()` hard-codes the `↳` glyph instead of routing it through `mark('↳')`. Every other `↳` site in the file uses `mark()`, and `_PLAIN_MARKERS` maps it to `->`, so `SUPERTOOL_PLAIN=1` leaks a raw non-ASCII glyph from this one call site into hook/CI output.

https://github.com/Digital-Process-Tools/claude-supertool/blob/bb91ac3d8f4e7def78a7e80600de82a0b624cef1/supertool.py#L11622-L11624

fdaviddpt:
### Code review

Three independent reviewers (logic/adversarial, conventions + code comments, git history). Four defects found, all reproduced before fixing, all fixed in a26113f and c62d06c.

1. **The branch footer keyed on landed writes, not attempted mutations.** `_WRITE_COUNT` is decremented by `_retract_write` on rollback, and a failed `edit` never reaches `_atomic_write` at all — so a batch whose sub-op failed its anchor or was rolled back by a regressing validator printed no footer. That is the zero-footer bug of #392 through a different door, at the two moments a wrong-branch hypothesis is most useful. Now counts attempts, bumped in `_run_with_validators` before the op runs. Found independently by two reviewers.

2. **`_dirs_up_to_repo_root` used `abspath`, so symlinked paths climbed past their real repo.** A file reached through a symlinked directory was judged by the link's own location, walked to the filesystem root, and missed the config sitting right next to it. Resolves first now — the same thing `_atomic_write` does a few hundred lines away, and for the same reason. The walk is also cached per directory.

3. **`format_staged` bypassed the opt-in gate entirely**, reproducing #393's original symptom from a pre-commit hook — the exact scenario the issue was filed about. Gated. `format:PATH` deliberately stays ungated: the caller named the file, and silently declining would be the wrong answer.

4. **The TOML hint fired on any odd count of `'''`**, including runs sitting harmlessly inside an ordinary basic string, attaching a delimiter lecture to an unrelated parse error. Now requires a `'''` that actually opens a value.

Plus two carried in c62d06c: the hint hardcoded `↳` instead of routing it through `mark()` (plain mode exists to strip that), and `_FORMATTER_SKIPS` is now reset on the outermost dispatch frame so an escaping exception cannot carry a dead call's skips into the next one.

Regression tests for all of it in `tests/test_review_regressions_395.py` (8 cases) — the two footer ones go red against a26113f's parent.

One convention note not acted on: `docs/contributing.md` says one feature per PR, and this is three issues in one. They touch overlapping regions of `supertool.py`, so splitting now would mean three conflicting rebases. Flagging rather than silently ignoring.

PR #403, opened 2026-07-27
fix(custom-ops): strip phpstan preamble under errors-only

### LINKED ISSUE #402: phpstan :errors-only still prints the full 'Instructions for interpreting errors' preamble
## What happened

```
./supertool 'phpstan:<path>:errors-only'
```

still emits PHPStan's 14-line prose preamble before the filtered output:

```
Instructions for interpreting errors
---------

Each error has an associated identifier, like `argument.type` ...
Before fixing the error, fetch the documentation page for its identifier.
The error usually indicates a real bug or incorrect type in the code. ...
Do not add `@phpstan-ignore` comments ...
Do not use assert() or inline @var PHPDoc tag ...
```

(That block is injected by this project's PHPStan config, not by PHPStan core — but it arrives on stdout ahead of the errors on every single call.)

## Why it matters

`errors-only` exists to make the output small enough to read repeatedly while iterating. On a run with one or two errors the preamble is the majority of the payload, and it is byte-identical on every call — pure repeated cost across a fix loop, in the one op you invoke most while fixing.

## Suggested fix

Under `:errors-only`, drop everything before the first line matching a diagnostic (`<abs-path>:<line>:<message>` or the `------` table frame). Keep the trailing `[ERROR] Found N errors` summary. If the preamble is worth preserving for the first call in a session, emitting it once and suppressing it on repeats would also work — but unconditional suppression under `errors-only` matches what the flag promises.

fdaviddpt:
## What changed

`phpstan:<path>:errors-only` was printing PHPStan's project-injected "Instructions for interpreting errors" preamble ahead of the diagnostics on every call, even though `errors-only` exists specifically to keep output small during a repeated fix loop.

`phpstan` is a project-configured custom op (`.supertool.json` → `ops.phpstan`), resolved generically by `_resolve_custom_op`. That function now strips everything before the first line that looks like a diagnostic — a `path:line:message` row, or the `------` table-formatter frame PHPStan emits — whenever `op == "phpstan"` and `"errors-only"` is one of the passed args. The trailing `[ERROR] Found N errors` / `[OK] No errors` summary is kept automatically since it comes after the diagnostics in the raw output.

The detector matches on line **shape**, never on the preamble's wording, per the issue's guidance — an allowlist of preamble phrases would have wrongly stripped a diagnostic message that happens to contain a preamble word.

## Why

- Default (non-`errors-only`) `phpstan:PATH` output is untouched — the filter only runs when `errors-only` is explicitly requested.
- On a run with one or two errors, the 14-line preamble was the majority of the payload, and byte-identical every call — pure repeated cost in the op most used during a fix loop.

## How I tested

TDD: wrote the failing tests first (`tests/test_custom_ops.py::TestPhpstanErrorsOnly`), then implemented `_phpstan_errors_only_filter` + the call site in `_resolve_custom_op`.

Covers:
- default output byte-for-byte unchanged (no `errors-only`)
- `errors-only` strips the preamble, keeps diagnostics + summary
- clean run (no errors) under `errors-only` — doesn't swallow the success line or return empty
- output with no preamble at all — passes through unchanged
- preamble present, zero errors — nothing to anchor on, passes through
- a diagnostic message containing the literal word "Instructions" survives (shape detector, not word-based)
- the `path:line:message` diagnostic shape (not just the table frame)
- a differently-named custom op that happens to receive `errors-only` as a literal arg is unaffected (scoped to `op == "phpstan"`)

Full suite: `3758 passed, 34 skipped, 6 deselected`, coverage 88.34% (gate 86%). No separate lint/format CI step in this repo — `.github/workflows/tests.yml` runs pytest only.

Closes #402

### DIFF
--- a/supertool.py
+++ b/supertool.py
@@ -934,6 +934,28 @@ def _expand_env(s: str, env: Dict[str, str]) -> str:
     )
 
 
+_PHPSTAN_DIAGNOSTIC_RE = re.compile(r'^\s*(?:-{3,}|\S+:\d+:)')
+
+
+def _phpstan_errors_only_filter(text: str) -> str:
+    """Strip PHPStan's project-injected "Instructions for interpreting
+    errors" preamble under `errors-only` (#402), keeping the diagnostics
+    and the trailing `[ERROR] Found N errors` / `[OK]` summary.
+
+    Anchors on the first line that LOOKS like a diagnostic — a
+    `path:line:message` row or a `------` table-formatter frame — never on
+    the preamble's wording, so a diagnostic whose message happens to
+    contain the word "Instructions" is never mistaken for preamble. When
+    no such line is found (clean run, or a project that never injects the
+    preamble), the text is returned unchanged.
+    """
+    lines = text.splitlines(keepends=True)
+    for i, line in enumerate(lines):
+        if _PHPSTAN_DIAGNOSTIC_RE.match(line):
+            return "".join(lines[i:])
+    return text
+
+
 def _resolve_custom_op(op: str, parts: List[str]) -> str | None:
     """Try to run op as a custom command from config["ops"].
 
@@ -1009,6 +1031,8 @@ def _resolve_custom_op(op: str, parts: List[str]) -> str | None:
         )
         elapsed = time.monotonic() - t0
         output = result.stdout
+        if op == "phpstan" and "errors-only" in parts[1:]:
+            output = _phpstan_errors_only_filter(output)
         if result.returncode != 0:
             if result.stderr:
                 output += result.stderr
--- a/tests/test_custom_ops.py
+++ b/tests/test_custom_ops.py
@@ -441,6 +441,167 @@ class TestResolveCustomOp:
 
 
 # ---------------------------------------------------------------------------
+# phpstan `errors-only` — strips the project-injected "Instructions for
+# interpreting errors" preamble that arrives ahead of the diagnostics on
+# every call (#402). The "phpstan" op is a project-configured custom op, so
+# the filter lives in _resolve_custom_op and is scoped to op == "phpstan".
+# ---------------------------------------------------------------------------
+
+_PHPSTAN_PREAMBLE = (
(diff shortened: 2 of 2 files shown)


fdaviddpt:
Parking this — and closing #403, which attempted it.

I'm the primary consumer of `phpstan:errors-only`, so I went in expecting to build it. Three things came out of the review that changed my mind.

**The attempted fix didn't work, and the tests didn't notice.** The anchor regex was `^\s*(?:-{3,}|\S+:\d+:)`, meant to find the first diagnostic-shaped line. The preamble's own second line is `---------`. That matches `-{3,}`, so the filter anchored on the preamble's underline, stripped the title line only, and returned the entire preamble body. The suite went green (3758 passed) because the test asserted the *title* string was absent and never checked the body. Worth recording as a test-design lesson independent of this issue: asserting the absence of one marker string is not asserting the absence of a block.

**`phpstan` is not a supertool op.** It's a project-configured custom op living in a consumer's `.supertool.json`, resolved through `_resolve_custom_op`. The fix hardcoded `if op == "phpstan"` into `supertool.py`, which puts one consumer's config knowledge into the generic core: any install with an op of that name gets silent output rewriting it never asked for, and renaming the op silently drops the filter. If this were built here, the right shape is a declarative post-filter on the op's own config entry (`"filters": {"errors-only": {"skip_until": "..."}}`) — generic mechanism, consumer-owned specifics.

**But the cheapest correct fix is upstream of this tool.** The preamble isn't emitted by PHPStan core or by supertool — it's injected by the consuming project's own PHPStan configuration. A project that doesn't want it on `errors-only` runs can stop injecting it there. Adding a core filter here means supertool grows a permanent mechanism to strip text that a config one layer up chose to emit.

So: not building it. The issue is real and the annoyance is real, it just isn't this repo's to fix.

Reopen if the declarative-filter version turns out to be worth having for its own sake — a generic `skip_until` on custom ops is a defensible feature. It just shouldn't arrive shaped as a phpstan special case.

Issue #551, opened 2026-07-29
`list_pidfiles` swallows an OSError into `[]`, so both `status` and `stop --all` claim nothing is running when they never managed to look


fdaviddpt:
Noticed while fixing #549. Not folded into that PR: the fix changes `stop.py`'s exit contract, which is #547's blast radius, not #549's.

## The line

`presets/mcp/_paths.py:110`:

```python
def list_pidfiles() -> list:
    base = runtime_dir()
    try:
        return sorted(...)
    except OSError:
        return []
```

An empty list has exactly one meaning to both callers: **there are no daemons.** An `OSError` from `os.listdir` means we do not know how many there are. The two are rendered identically.

## What each caller does with it

```
$ # os.listdir raising OSError(5, "Input/output error"), one real pidfile present
list_pidfiles -> []
No supertool MCP daemons running.      # status.py
No daemons running.                    # stop.py --all
stop --all rc = 0                      # EXIT_OK
```

- `status.py` prints `No supertool MCP daemons running.` — the same defect #549 fixed one row down, only now at the level of the whole table. #549 made an unreadable *pidfile* honest; an unlistable *directory* is still a confident "none".
- `stop.py --all` is the sharper one. It returns `EXIT_OK`, which #547 documents as *"nothing was running — nothing stale can come from nothing."* That sentence is only true when we looked. Here we did not, so new-file invalidation records a success for a stop that never happened, which is #239 reached through the path built to prevent it — the exact failure #547 exists to close, one layer below where it was closed.

## Reachability, stated honestly

Low, and this is a latent instance rather than a live one. `runtime_dir()` runs first and closes the easy routes: it `chmod`s the directory back to `0700` (so a `chmod 000` heals rather than raising — verified on macOS), and it `sys.exit`s outright on foreign ownership (#148/#544). What is left is genuine but rare: the directory removed between `mkdir` and `listdir`, `EIO` on a failing volume, `EMFILE`/`ENFILE` under fd pressure, a `chmod` that failed into the `except OSError: pass` above on a read-only or flag-locked filesystem.

Filed anyway because the cost of the fix is small and the failure is silent by construction: nothing in either surface can ever report that this happened.

## Shape of a fix, and why it is not a one-liner

`list_pidfiles` has to stop conflating "none" with "could not tell". Whichever way it signals that — letting the `OSError` out, or returning a discriminated result — both callers need a branch:

[...]

Issue #547, opened 2026-07-29
`_mcp_stop_server` cannot distinguish a refusal from a success, so a daemon that failed to die looks exactly like one that died


fdaviddpt:
## The seam

`_mcp_stop_server` (`supertool.py:11390`, `11450`) runs `stop.py` as a subprocess with `stdout=DEVNULL, stderr=DEVNULL, check=False`, wrapped in `except (OSError, SubprocessError): pass`.

Every outcome collapses to the same one:

| what actually happened | what the caller sees |
|---|---|
| the daemon was stopped | nothing |
| `stop.py` refused (e.g. an unverifiable runtime dir → `SystemExit`) | nothing |
| `stop.py` crashed | nothing |
| the binary was missing entirely | nothing |

That suppression is **deliberate and documented** — invalidation is an optimization and must never block the op — and I am not proposing to make it blocking. The defect is that it discards the *outcome* along with the *blocking*, which are separable.

## Why this is not theoretical

Found while fixing #544. On Windows this currently costs nothing: no warm daemon can exist there, so a `SystemExit` out of `stop.py` is harmless, and that harmlessness is exactly what makes the loud refusal on that path acceptable.

The cost lands on a platform where a daemon *can* exist. A stale warm daemon that failed to die goes on answering from a reflection of the file it captured earlier — that is **#239's defect precisely** — and the invalidation path whose entire job is to prevent it reports nothing at all. The safety net and the failure are silent in the same direction.

## The house pattern, one layer out

This is the same shape filed a dozen times now: *an absence produced by the tool, read as an absence in the world.* Here the tool produces no signal by construction, and "no signal" is indistinguishable from "the daemon is gone".

The framework already has the vocabulary — `ok` / finding / `skipped`, `validators/common/refusal.py`, and `docs/validators.md`'s "Declining instead of guessing". As with #263, the abstraction exists and this call site has not adopted it.

## Shape, for whoever takes it

Keep non-blocking. Capture the exit status and a bounded slice of stderr instead of `DEVNULL`, and log a single line when the stop did not succeed. The op continues regardless — the only change is that a failed invalidation stops being invisible.

Judgment call to make rather than assume: **where does that line go?** It must not appear in the op's normal output — that would turn a background optimization into user-facing noise on every edit, which is a worse trade than the current silence. A debug-gated log line, or the existing warm-daemon diagnostics channel, is more likely right.

Issue #239, opened 2026-05-28
validate: warm phpstan-mcp reports false errors on brand-new classes (stale reflection)


fdaviddpt:
## Summary

`validate:` (the warm `phpstan-mcp` process) reports **false** errors on brand-new class files — the warm process doesn't reindex newly-created classes.

## Repro

1. Create a new `*.class.php`.
2. `./supertool 'validate:path/to/New.class.php'`
3. Output shows:
   - `phpstan-mcp` → bogus `return.missing` on methods that clearly `return` (and even on *unchanged* methods elsewhere in the file).
   - `rector-mcp` → `System error: "ClassReflection must be resolved for class New"`.
4. Re-running `validate:` returns **byte-identical output including timings** → result is stale/cached.
5. `./supertool 'dvsi_clearcache'` does **not** fix it (that only clears DVSI's autoload cache, a different cache).
6. A fresh `./supertool 'phpstan:path/to/New.class.php:errors-only'` (~8–10s) is **correct** (passes).

## Impact

On any new class you can't trust `validate:` — the errors look real (`return.missing` is a plausible message). Cost a couple of round-trips to recognize they were stale and fall back to the fresh `phpstan:` op.

## Suggested fix

- Invalidate the phpstan reflection/result cache when a new class file is detected (mtime/new-path based), or
- Auto-fall back to a fresh phpstan run for files created/modified within the last N minutes, or
- At minimum, surface a hint: "new class — warm phpstan cache may be stale, prefer `phpstan:`".

Env: supertool 0.14.0.

PR #552, opened 2026-07-29
fix(mcp): status says unknown, not dead, for a pidfile it could not read (#549)

### LINKED ISSUE #549: `status` reports a daemon as `dead` when its pidfile is merely unreadable — the one op you run to find out answers a question it did not ask
## The line

`presets/mcp/status.py:60`:

```python
except (OSError, ValueError):
    pid = 0
```

`pid = 0` flows into the liveness probe, `alive` comes back `False`, and the row prints **dead**.

## Why that is a defect and not a rounding error

An unreadable pidfile — truncated, permission-denied, mid-write, garbage — means *we do not know whether a daemon is running*. "Dead" is not a cautious reading of that; it is a specific, confident claim that happens to be one of the two possible answers.

This is the twelfth-ish instance of the pattern this repo keeps hitting: **an absence produced by the tool, read as an absence in the world.** What makes this one worth filing above the general run is *where* it sits. `status` is not a background optimization whose silence nobody notices — it is the op a human runs **specifically to find out whether a daemon is alive**. Every other surface can afford to guess; this one is the surface whose entire product is the answer.

Apply the test that has been catching these: *can someone acting reasonably on this output conclude the opposite of the truth?* Yes, directly. A live daemon holding a stale index reads as `dead`, so the operator does not restart it, and the daemon goes on answering validators from a reflection captured before the file changed — **#239, arrived at via the tool built to prevent it**. And once #547's `restartMcp` starts reporting `FAILED to stop`, `status` is the first place anyone will look to confirm; it should not be the place that misleads them.

## Shape of the fix

The vocabulary already exists — three states, not two: `ok` / a finding / `skipped`, per `docs/validators.md`'s "Declining instead of guessing", and `validators/common/refusal.py`. As with #263, the abstraction is present and this call site has not adopted it. A third rendering — `unknown`, with the reason (`unreadable pidfile: <errno>`) — is the whole change.

Judgment call for whoever takes this, worth deciding rather than assuming: **does `unknown` need to be visually distinct from `dead` in the table, or is the reason string enough?** A row that merely says `unknown` with no explanation trades a wrong answer for an unhelpful one, which is only half the fix.

Found while fixing #547 (PR #548), by the agent working that issue. Not folded into it — the invalidation seam and the status surface are different code paths, and bundling would put an unrelated file inside that PR's blast radius.

fdaviddpt:
`status.py` caught `(OSError, ValueError)` around the pidfile read and set `pid = 0`. Zero is falsy, so `_proc.pid_alive` was never called and the row printed **dead** — not a cautious reading of an unreadable file, but one of the two possible answers, asserted about a process nothing had looked at.

Reproduced before touching the code (a pidfile containing `not-a-pid`):

```
NAME             HASH           PID      STATUS   UPTIME     IDLE       SOCKET
?                eeeeeeeeeeee   0        dead     0s         -          /…/supertool-mcp-eeeeeeeeeeee.sock
```

## Why this surface

`status` is the op a human runs *specifically to find out whether a daemon is alive*. Every other surface can afford to guess; here a wrong answer is the product. A live daemon holding a stale index that reads as `dead` does not get restarted, and goes on answering from a tree captured before your edit — #239, reached through the tool built to prevent it.

It also has to agree with #548. `stop.py` now calls an unreadable pidfile a **failure** (`EXIT_STOP_FAILED`) rather than a success, and `restartMcp` reports `FAILED to stop …`. `status` is the first place anyone will look to confirm such a report, and it was the surface that would have contradicted it.

## The change

`read_pid(pid_path) -> (pid, reason)`. A non-empty reason means the pid is unknowable, and the caller renders `STATUS_UNKNOWN` rather than probing a pid it does not have. Four reasons: unreadable (`OSError`, with `strerror`), empty (`a daemon may be mid-write`), unparsable (`ValueError`, with the first 40 bytes `repr`'d so control characters cannot escape into the terminal), and `pid <= 0`.

## The judgment call: distinct rendering *and* a reason — the reason alone is half a fix

A row reading `unknown` with nothing after it trades a wrong answer for an unhelpful one, which is why both:

```
NAME             HASH           PID      STATUS   UPTIME     IDLE       SOCKET
?                aaaaaaaaaaaa   4242     dead     0s         -          /…/supertool-mcp-aaaaaaaaaaaa.sock
?                bbbbbbbbbbbb   ?        unknown  0s         -          /…/supertool-mcp-bbbbbbbbbbbb.sock
                                ↳ unparsable pidfile: 'garbage'
?                cccccccccccc   48505    alive    0s         -          /…/supertool-mcp-cccccccccccc.sock
```

- **PID reads `?`, not `0`.** Printing `0` invents the one number the row exists to report and is where the original bug entered.
- **The reason is an indented continuation line, not a column.** A column would have to fit inside a row that already ends in an absolute socket path. Interleaved rather than a footnote block so the reason sits against its own row.
- **The reason names the *pidfile*,** so the reader's next action is to look at the file rather than at the daemon.

## Which failure I chose

The half that is easy to lose is `dead`. Only the read failure changed state:

- A stale pidfile holding a real pid whose process is gone still reads `dead`. That verdict was *measured*, it is the common and useful signal, and blanketing it into `unknown` would have traded the loud bug for the quiet one.
- `pid <= 0` moved to `unknown`: `0` parses as an int and is not a process that died, it is garbage that survived `int()`.
- The only new false-`unknown` is the sub-millisecond race where a daemon unlinks its pidfile between `listdir` and `read` — previously `dead` (accidentally right), now `unknown` (honestly unknown). Rare, and in the safe direction.

`TestRealVerdictsSurvive` pins the alive and dead paths on the same fixture as the unknown ones, so a fix that over-reaches fails. Those four tests passed **before** the fix too — they are guards against this PR, not scenery.

Exit status is unchanged at `0` in every case. This is a report a human reads, not a check a caller consumes; a non-zero exit would make an unreadable pidfile look like a failure of the op itself. Grepped: nothing parses `status.py` output.

## Red, then green

```
$ python3 -m pytest tests/test_mcp_status_unknown_549.py -q      # before the fix
6 failed, 4 passed
FAILED …::test_garbage_pidfile_reads_unknown_not_dead
FAILED …::test_truncated_pidfile_reads_unknown
FAILED …::test_permission_denied_pidfile_reads_unknown
FAILED …::test_pid_zero_is_unknown_not_dead
FAILED …::test_unknown_row_states_why
FAILED …::test_unknown_row_does_not_print_a_fabricated_pid

$ python3 -m pytest tests/test_mcp_status_unknown_549.py -q      # after
10 passed

$ python3 -m pytest -q                                           # full suite
4739 passed, 36 skipped in 101.43s — coverage 88.84% (gate 86%)
```

The `presets/` tree does not count toward the coverage gate, which measures `supertool.py` only; these are written for correctness, not for the number.

## Docs

`CHANGELOG.md` under `## Unreleased`, and a new **When status cannot tell** section in `docs/mcp-integration.md` — a three-value STATUS table, a worked example of the `unknown` row, what to do about one, and why the exit code stays `0`. The lifecycle table now names the three values.

## Filed separately: #551

`_paths.list_pidfiles` does the same thing one layer over — `except OSError: return []`, so an unlistable runtime dir becomes "there are no daemons". `status` prints `No supertool MCP daemons running.` and `stop.py --all` returns **`EXIT_OK`**, which #547 documents as *"nothing stale can come from nothing"* — a sentence only true when we looked. Reproduced with `os.listdir` raising `EIO`. Reachability is low (`runtime_dir()` chmods perms back and `sys.exit`s on foreign ownership, so the easy routes are already closed), which is why it is a latent instance and not a live one — and why it is not here: fixing it changes `stop.py`'s exit contract, which is #547's blast radius, not #549's.

Closes #549


fdaviddpt:
## Diagnosed: not a race, not this file — a permanent `sys.path` mutation in `test_comment_file.py`

Pushed `acfdf75`.

### It reproduces deterministically, serially, without my file

```
$ pytest tests/test_comment_file.py tests/test_mcp_stop_outcome_547.py -p no:randomly -n0
39 passed, 5 errors
ERROR …test_no_daemon_is_its_own_code - ModuleNotFoundError: No module named 'stop'
```

That selection contains **only files that are on master**. Your hypothesis 2 (xdist race) and 3 (platform) are both out: `-n0` means one process, `-p no:randomly` means fixed order, and it fails every time on macOS 3.14 locally.

### Cause

`tests/test_comment_file.py:25`:

```python
sys.path[:] = [p for p in sys.path if "presets/" not in p.replace("\\", "/")]
```

Process-wide, permanent, never restored, and far wider than its stated purpose (keeping sibling `_auth`/`_outbound` shims out of the module it loads). It strips **`presets/mcp`** as collateral. Both mcp test modules put that directory on the path at *import* time and then `import stop` / `import status` **lazily, inside a fixture** — so the entry has to survive until the test runs, not merely until collection ends. Any worker that ran `test_comment_file.py` first left them with `ModuleNotFoundError`.

Your hypothesis 1 was directionally right — two files sharing path entries with lazy imports — but the conflict is not mutual. `test_comment_file.py` is purely the aggressor; the mcp files only read the path they set up.

### The fact about master you asked for

**#547's merged tests have been fragile since `8fcb7e04`.** The reproduction above is master's own files. Nothing about my branch caused it; adding a test file changed which xdist worker got which pairing, and that is the only thing that varied. This PR fixes it, so it is not left living on master — that is why the fix is here rather than in its own issue.

### The fix

`_load`'s filter is now scoped to the `exec_module` call and restored in a `finally`. Safe because the shim collision only exists while the module body runs — by the time the path goes back, the module has bound what it imported.

**No defensive import anywhere.** Nothing was added around `import status` or `import stop`; the fix is where the damage is done, so the victims stay honest about their own requirements.

New test `test_load_leaves_sys_path_as_it_found_it` pins it, asserting on the exact entry that was collateral rather than only on list equality, so a narrower-but-still-permanent filter cannot pass it. Verified not-free by disabling only the restore:

```
E   assert sys.path == before, "sys.path escaped _load"
FAILED tests/test_comment_file.py::test_load_leaves_sys_path_as_it_found_it
```

(First attempt at that guard failed through its own teardown with `ValueError: list.remove(x)`, masking the assertion — teardown made tolerant so the assertion is what reports.)

### Verification

```
reproduction, now:   55 passed
test_comment_file.py alone (shim isolation intact):  29 passed
full suite, 3 randomized xdist runs:  4740 passed, 36 skipped  (×3)
coverage 88.84% (gate 86%)
```

### Filed: #555 — five more of the same construct

`test_hashnode.py:21`, `test_devto.py:17`, `test_security_devto.py:39`, `test_bluesky.py:21`, `test_security_bluesky.py:36` all do the same permanent global `sys.path[:] =` with narrower filters.

**Not live today**, and I want to be plain about that rather than inflate it: their consumers load by absolute path via `spec_from_file_location` and re-insert behind an `if not in sys.path` guard, so the damage self-heals at the next `_load`. But that property is asserted nowhere — it is a coincidence of how those six files happen to import. The first bare lazily-imported module under `presets/devto` or `presets/hashnode` reproduces this exactly, on a leg the scheduler picks, with a traceback pointing at the victim. Left out of this PR because they are latent and this PR is already carrying one unrelated fix.

Issue #585, opened 2026-07-30
gh-pr renders "no check runs on this commit" for both a run that has not appeared yet and one that will never be created


fdaviddpt:
## The reading that has no third state

`./supertool 'gh-pr:N'` renders a commit with no check runs as:

```
Checks: none reported — no check runs on this commit
Mergeable: UNKNOWN — no checks reported
```

That single sentence covers two situations with opposite consequences:

1. **The run does not exist yet.** GitHub has not created it. Measured today, PR-open → run-created: **99s** (#580), **165s** (#582's second push), **~2min** (#584), **4.5min** (#582's first). Waiting is correct.
2. **The run is never coming.** `.github/workflows/tests.yml` is `on: push: branches: [master]` plus `pull_request:` — so a pushed branch with no open PR gets nothing, ever, and waiting is a deadlock.

Nothing in the output distinguishes them, and both read as "not yet" to a reader who is waiting to merge. This is the house defect (`ok` / finding / **skipped**) sitting in the merge path: an absence the tool has not established, reported in the same words as one it has.

## It has now misled two readers in one session

- I read it on #582 and reported to Florian that the tool had misreported a green PR. It had not — I had compared a UTC `created_at` against a CEST clock and the run genuinely did not exist yet. My error, but the output gave me nothing to catch it with.
- The agent implementing #568 read the same output on the same PR and concluded the workflow had not triggered at all, then had to measure the latency by hand (165s, `total_count: 0` for the whole window) to establish that it had.

Two readers, same output, same wrong conclusion, ten minutes apart. The information needed to tell the cases apart was available to the tool in both instances and was not printed.

## What it should say instead

Three states, distinguished by evidence the op can already reach:

| State | Evidence | Rendering |
|---|---|---|
| Runs exist | `check-runs` non-empty | today's tally (unchanged) |
| **No runs yet** | head commit pushed *N* seconds ago, and the workflow *does* trigger for this ref | `no runs yet — head pushed Ns ago, first run typically appears within ~3min` |
| **No runs expected** | no open PR for the branch, and the workflow's `on:` does not cover a push to this ref | `no runs will be created — tests.yml triggers on pull_request and on push to master only; this ref has no open PR` |

[...]

Issue #572, opened 2026-07-29
`.githooks/pre-push` still resolves a bare `python3` — the last spawn position holding the name


fdaviddpt:
Found while widening the `python3` guard (#564 / PR #570). Outside that bug's blast radius: the guard is an AST scan of `*.py`, and this is a shell script, so no version of #570 can see it.

## The line

`.githooks/pre-push:27`

```sh
PYTHON="${PYTHON:-python3}"
```

`python3` here is a PATH lookup of exactly the name #529, #559 and #564 exist to keep out of spawn positions — the one that resolves to the Windows App Execution Alias stub, which *blocks* rather than errors. The repo has now converted 27 test spawn sites, one shipping lint path (#559) and one daemon default (#564) away from it, and this is the last instance of it in a position where something gets run.

## Why it is lower severity than the ones already fixed, but not zero

- It is a developer-machine pre-push hook, opt-in via `core.hooksPath`, not shipping code and not CI.
- The default is overridable: `PYTHON=... git push` works, so anyone who hits it has an escape hatch — if they can tell what happened.

That last clause is the problem. On a Windows dev box with the alias enabled, the hook does not fail — it *hangs*, which is the specific failure mode this repo keeps paying for, and it hangs inside `git push`, where the natural conclusion is that the network or the remote is slow. On POSIX it silently selects the system interpreter rather than the venv the developer is working in, so the hook can pass or fail on a different set of installed packages than the suite the developer runs by hand.

## What to consider

A shell script has no `sys.executable`. The honest equivalents, roughly in order of preference:

1. Resolve a versioned name first and fall back deliberately: try `python3.13`/`python3.12`/… or `py -3` on Windows, and **fail with a message** rather than running an unresolved `python3`. The versioned names are not aliased to the blocking stub, which is why #570 does not flag `shutil.which("python3.13")`.
2. Keep `${PYTHON:-...}` but have the hook verify the resolved interpreter actually runs (`"$PYTHON" -c 'import sys'`) before using it, and decline loudly if it does not — the same "a checker that cannot start says so" contract #559 established for the post-edit lint.

Worth deciding whether the guard should grow a companion check for shell scripts, or whether one line in one opt-in hook is better fixed than automated. My view is the latter: a second scanner over `sh` with no other instance to defend is a lot of surface for one line, and #570's own lesson is that a scanner whose scope does not match the defect is worse than none.

Issue #554, opened 2026-07-29
claude-channel has no delivery observability: process-alive, socket-held and write-succeeds all read the same whether events arrive or not


fdaviddpt:
## Summary

A single malformed event on `/tmp/supertool-watch.sock` throws inside the `claude-channel` notification handler, and the throw escapes far enough to take down the MCP STDIO connection. The event is lost, and there is a silent window during which poller events are accepted by the socket and never delivered to the session.

Reproduced accidentally on 2026-07-29 while running the transport probe documented in the `pipeline-radar` skill.

## Repro

The probe recipe in the skill / vocab did not specify a type for `ts`, so it was sent as a float:

```python
import socket, json, time
ev = {"ts": time.time(),            # <-- float
      "source": "gitlab-mr", "id": "99999",
      "event": "pipeline_succeeded",
      "payload": {"title": "probe", "url": "https://example/probe"}}
s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
s.connect("/tmp/supertool-watch.sock")
s.sendall((json.dumps(ev) + "\n").encode())
```

The write succeeds. MCP log (`~/Library/Caches/claude-cli-nodejs/<project-slug>/mcp-logs-claude-channel/*.jsonl`):

```
21:37:13  Channel notifications registered
21:40:43  STDIO connection dropped after 209s uptime
21:40:43  Connection error: Uncaught error in notification handler: $ZodError: [
            { "expected": "string", "code": "invalid_type",
              "path": ["params","meta","ts"],
              "message": "Invalid input: expected string, received number" } ]
```

A corrected probe with `"ts": "2026-07-29T21:45:00Z"` ~3 min later was delivered normally:

```
21:43:47  notifications/claude/channel: gitlab-mr 99999: pipeline_succeeded
```

No new `Starting connection` line, no `/mcp` reattach, no relaunch — so the link does heal. Blast radius is the one rejected event plus the gap, not a permanently dead radar.

## Why this matters more than the type error

The type mismatch is trivial. The structural problem is that **event validation failure and connection liveness are coupled.** Coercing `ts` fixes exactly one field; the next malformed value (null `payload.url`, missing `event`, integer `id`, an unexpected key from a new poller source) reproduces the same drop.

It is also invisible from inside the session. During the window:

- `pgrep -fl channel.ts` — green, process alive
- `lsof /tmp/supertool-watch.sock` — green, socket still held
[...]

PR #592, opened 2026-07-30
fix(ci): make a Windows red readable, and make the board answer for notifiers/ (#546, #557)

### LINKED ISSUE #546: CI's own pytest emits non-cp437 characters on the Windows legs, so a Windows failure message arrives unreadable
## What happened

Debugging the four red Windows legs on #545 cost extra round-trips because the failure message itself was corrupted:

```
SystemExit: daemon: cannot verify ownership of runtime dir C:\...\rt on this platform
  os.geteuid does not exist and st_uid is a constant here...
```

That `` is where `presets/mcp/_paths.py`'s em dash should be. The raw log bytes are `EF BF BD` — U+FFFD, the replacement character — meaning the character was already lost at emit time, not mangled in transit.

## It is not the log transport, and it is not our product

Two checks settle both halves:

- **Transport is fine.** Em dashes echoed by the *workflow YAML itself* arrive in the same log intact, as `E2 80 94`.
- **supertool is already immune.** It reconfigures its own stdout to UTF-8 (`supertool.py:550`) and pins `PYTHONIOENCODING=utf-8` for every child it spawns (`supertool.py:14431`) — that was #415/#418.

The unpinned emitter is **CI's own `pytest` invocation**. Em dash exists in cp1252 but not in cp437, which is the console codepage on the GitHub Windows runners, so pytest replaces it while rendering the assertion output.

## Why it is worth fixing despite being cosmetic

It is only ever visible when a Windows leg is already red — i.e. exactly when someone is reading the output under pressure and cannot reproduce locally. A garbled message in that moment reads as "something is broken with encoding" and sends the reader after a product defect that does not exist. It cost that detour once already on #545.

## Shape of the fix

Two lines of `env:` on the pytest step in `.github/workflows/tests.yml` (`PYTHONIOENCODING=utf-8`, `PYTHONUTF8=1`). It touches all twelve legs and it is CI config, so it is **not** an auto-merge: it wants Florian's eyes before it lands.

Deliberately kept out of #545 — that PR is product-adjacent test scoping, and bundling a workflow change into it would put CI config inside a bugfix's blast radius.

Found while fixing #544.

### LINKED ISSUE #557: CI runs no non-Python code, so `notifiers/` ships on a 12/12 green that never executed it
## The gap

`.github/workflows/tests.yml` runs `pytest` across {ubuntu, macos, windows} × py3.9–3.12. That is the entire suite. There is no `bun`, `node`, or `tsc` step anywhere in the workflow.

`notifiers/claude-channel/channel.ts` is 200-odd lines of TypeScript that runs in every radar session, and **no CI leg has ever executed a line of it.** Neither has `notifiers/cursor-witness/extension/src/extension.ts`.

## Why this is worse than "some code is untested"

The board says 12/12 green. That tick is *true* — every Python test passed — and it is read, by me and by anyone else looking at a PR, as "this change is verified." For a PR that only touches `.ts`, the green tick and a completely unexercised change are indistinguishable.

That is this repo's signature defect, arriving in the merge gate itself: **an absence produced by the tool, read as an absence in the world.** The suite cannot answer for `notifiers/`, and instead of saying so it renders the same green it renders for a fully covered change.

Concretely, on #556 (the #554 fix): the change was verified only because it was run **by hand on a machine that happens to have bun**. The tests written for it are real integration tests — they spawn the server over a real socket and assert on the emitted JSON-RPC — but they `skipif` on `shutil.which("bun")`, which is `None` on all twelve legs. So they are collected, skipped, and counted as neither a pass nor a failure. A future regression in `channel.ts` merges green.

## Shape of the fix — and the decision that has to be made first

Adding a bun (or node) step is a **workflow change**, which is not something I merge unilaterally — it needs the owner's call. So this issue is the decision, not a patch.

The options are not equivalent:

1. **Install bun on the legs and drop the `skipif`.** Most faithful — the existing tests then run for real. Cost: a toolchain install on twelve legs for one directory's worth of code. Possibly narrow it to one leg (ubuntu) rather than the full matrix, since the transport is POSIX-only in practice.
2. **A separate lightweight job** that runs `bunx tsc --noEmit` plus the channel tests, gated to `notifiers/**` paths. Cheaper, still catches type regressions and the wire-shape tests.
3. **Accept the gap explicitly** — but then the skip must be *visible*. A skipped test that silently vanishes into "36 skipped" is the current state and is exactly the problem. At minimum the suite should say "channel tests skipped: bun unavailable" somewhere a reader of the board can see.

(3) alone is the weakest, but it is strictly better than today, because today nothing distinguishes "skipped for want of a toolchain" from "not applicable here."

## Related

- #546 — CI's pytest mangles non-cp437 characters on the Windows legs, making failures there hard to read.
- #553 — a test that fails when the product correctly declines, red on master twice today.

All three are about the same thing from different angles: **the CI board is currently less trustworthy than it looks**, and each one costs real time at exactly the moment someone is trying to decide whether a change is safe.

Found while reviewing #556.

fdaviddpt:
Closes #546
Closes #557

One PR, because they are one property: **a green tick that means less than it looks like, and a red one you cannot read.** #546 is the board's output channel, #557 is its coverage; both fixes land in the same 45-line workflow file and the same suite, and splitting them would mean two PRs each editing the other's lines.

---

## Premises, checked before building

**#546 — Windows really does destroy the character. Confirmed in real logs.**

| run | job | evidence |
| --- | --- | --- |
| [30485881190](https://github.com/Digital-Process-Tools/claude-supertool/actions/runs/30485881190) | 90691288041 (windows-latest, 3.9) | `...runtime dir C:\...\rt on this platform \xEF\xBF\xBD os.geteuid...` |
| [30500828589](https://github.com/Digital-Process-Tools/claude-supertool/actions/runs/30500828589) | 90739894758 (windows-latest, 3.12) | `...presets/mcp/_spawn.py is not scanned \xEF\xBF\xBD the walk has been...` |

`EF BF BD` is U+FFFD. Both source files hold a literal em dash at that position. Two controls settle where it happened: each of those logs also contains **two intact em dashes** (`E2 80 94`, echoed from this workflow's own step names), and the macOS/ubuntu failure logs from the same day contain **zero** U+FFFD. The transport is clean; the emitter is CI's pytest, whose terminal writer renders onto a stream taking its codec from the console codepage. U+FFFD rather than a crash means that stream had `errors="replace"` — the loss was silent, which is the worse of the two failure modes.

**#557 — `notifiers/` really is unexecuted. Confirmed by reading the workflow.**

`.github/workflows/tests.yml` had exactly one job. No `bun`, `node`, `tsc`, `shellcheck` or `bash -n` anywhere. The whole non-Python surface is small and finite: 3 `.ts` (one a test fixture), 4 `.sh`, 2 extensionless `.githooks/` scripts. The existing channel tests are real — they spawn the server over a socket and assert on emitted JSON-RPC — and they `skipif` on `shutil.which("bun")`, which is `None` on all twelve legs.

**The flakes premise held, and both are being handled elsewhere.** Across the last 120 runs (back to 2026-07-29 05:07) there were 13 non-success runs, and the Windows-only *timing* flake I could find in them is `test_vim_receipt_diff_and_lint.py::test_xml_lint` — red on four separate Windows legs (06:30, 15:45, 18:58, 21:11 on 07-29). It is timing: the elided part of its summary line reads `lint did not run to completion; the file was NOT checked. Raise SUPERTOOL_LINT_TIMEOUT if this recurs.` It is **already fixed** in `50a5ede` (07-30 00:19), which set `SUPERTOOL_LINT_TIMEOUT=30` in `tests/conftest.py`, after the last of those four. The second one is `test_read_when_ready`, which I did not find in my sample but which open PR #589 is fixing right now. So: two, as stated, both owned elsewhere, and **no mitigation for either has been added here** — this PR must not be read as having touched them.

Worth noticing anyway, because it changed this PR's shape: that flake is a *second*, independent readability defect. Its summary line was `assert 'POST-EDIT LINT FAILED' in 'vim C:\...\x.xml (...int (5s) ---\nlint did not run...` — pytest elides the middle of a long message to fit the terminal width, and the elided part held the word `timeout`, which **was** the diagnosis. So a Windows failure today is unreadable in two independent ways, and this PR fixes both.

---

## Judgment call 1 (#546): fix the emitter, not the environment

Three candidate layers, and they are not equivalent:

| | fixes the pytest emitter? | inherited by subprocesses? |
| --- | --- | --- |
| `env: PYTHONIOENCODING=utf-8` / `PYTHONUTF8=1` | yes | **yes — every child, for ever** |
| `-X utf8` on the pytest invocation | yes | no (verified: parent `utf8_mode=1`, child `utf8_mode=0`) |
| a per-test escape | no — the mangling is in the reporter, not in any test | n/a |

**Chosen: `-X utf8` on the pytest invocation.** Inheritance decides it. This repo reproduces Windows encoding defects by *leaving the environment alone*:

- `tests/test_encoding_seam.py::_run` strips `PYTHONIOENCODING`, `PYTHONUTF8`, `PYTHONCOERCECLOCALE`, `LC_ALL`, `LC_CTYPE`, `LANG` from the env it hands its children — it survives either choice.
- `tests/test_git_commit_payload_route.py::test_commit_succeeds_on_a_non_utf8_console` copies `os.environ` and overrides **only** `PYTHONIOENCODING=cp1252`. An inherited `PYTHONUTF8=1` changes what it reproduces: `locale.getpreferredencoding()` becomes UTF-8 under UTF-8 mode, which is exactly the axis its sibling at line 376 is about (that one pins `PYTHONUTF8=0` explicitly, and is safe).

A fix whose blast radius covers the tests that would catch its own regression is not a fix. `-X` is a flag on one interpreter: it fixes the process that renders the diagnostic and reaches nothing else. `tests/test_ci_encoding_546.py` fails if either variable appears anywhere in the workflow outside a comment, so the decision is pinned rather than remembered.

**On `errors=`, since the issue is right that it matters:** no `errors="replace"` anywhere. `replace` is what produced the U+FFFD — on a diagnostic it destroys rather than discloses. The reporter uses `backslashreplace`, which can only ever fire on a lone surrogate (UTF-8 encodes everything else) and keeps the byte recoverable. Nothing parsed is touched: `junit.xml` is read as XML, and the only consumer of the reporter's stdout is a human reading a log.

**The second half of readability, which the issue did not ask for and needed.** `--tb=no` means the one-line summary is the entire record, and pytest truncates it. The always-run step now reads the **untruncated** message out of `junit.xml` — written as UTF-8 by pytest regardless of what the console can encode — via `.github/scripts/junit_summary.py`, which pins its own stdout **in-process**. That is the layer no runner environment can bypass and no future workflow edit can remove by deleting a variable.

**And it removes a lying diagnostic.** The step it replaces ended `... || echo "(no junit.xml — pytest crashed before generating it)"`. That sentence was printed for *any* non-zero exit of the parser, including a `UnicodeEncodeError` raised while merely printing a message with an em dash in it — a reporter's own failure rendered as a fact about the suite, in the one place someone is trying to work out what broke. Missing file, unparseable file and zero failures now print three different sentences. The reporter always exits 0 by design: it runs under `if: always()` beside the step that owns the verdict, and one that could fail would be able to turn a green run red for a reason unrelated to the diff.

---

## Judgment call 2 (#557): scoped by what each thing needs, not by "run everything"

| code | where it now runs | what is checked | cost |
| --- | --- | --- | --- |
| `notifiers/claude-channel/channel.ts` | **new `notifiers` job**, ubuntu + macOS | `bunx tsc --noEmit -p .` under the channel's own strict tsconfig, then the two socket-level integration test files, for real | 0.2s bun install (lockfile) + 0.6s tsc + 4.0s tests |
| `*.sh` and `.githooks/*` | **all 12 existing pytest legs** | `bash -n` | ~0 — no new job, no new toolchain |
| `notifiers/cursor-witness/extension/src/extension.ts` | nowhere | **nothing — stated, not implied** | — |

**The split is the decision.** `bash -n` needs no toolchain and answers a platform-independent question, so a new ubuntu-only job for it would be strictly worse than the suite: in the suite it runs on twelve legs for free. TypeScript needs a runtime, so it gets a job.

**Not on `windows-latest`, and that is a decision rather than an omission.** `channel.ts` binds an AF_UNIX socket. Those tests skip on Windows *by platform*, whatever is installed — so a Windows leg here would install a toolchain in order to report a green it had not earned, which is #557's defect wearing #557's fix. That is the theatre test applied honestly: the bug class here is POSIX-socket-shaped, not Windows-shaped. macOS **is** in, because this exact code has already produced a macOS-specific defect — the ~104-byte `sun_path` cap, which is why the #554 harness pins `/tmp` instead of `tmp_path`.

**The job cannot pass by skipping.** That is the part worth having, and it is `SUPERTOOL_REQUIRE_JS=1` plus `tests/_toolchain_gate.py`: a `skipif` that a job can promise to convert into a hard error. Demonstrated both ways below. The promise is deliberately **not** inferred from `CI` — eleven of the twelve pytest legs are on CI and install no JS runtime, so inferring it would turn the whole board red to fix a reporting defect.

**Option (3) from the issue is kept as well, not instead.** What remains uncovered is enumerated in `docs/contributing.md` beside a table of what each leg checks, and the `.ts` inventory in `tests/test_ci_non_python_coverage_557.py` is asserted against `git ls-files` — so a new TypeScript file fails the suite until somebody classifies it `executed`, `uncovered` or `fixture`. A gap nobody can widen silently is the durable version of "accept it explicitly".

---

### #557 — before → after, same tree, same commit

With `bun` off `PATH` and no promise, which is today's CI:

```
$ env PATH="/tmp/nobun:/usr/bin:/bin" python3 -m pytest tests/test_notifiers_claude_channel_55{0,4}.py -q
sssssssssss                                                              [100%]
11 skipped in 0.08s                                          # exit 0 — green
```

With the promise armed, which is what the new job sets:

```
$ env SUPERTOOL_REQUIRE_JS=1 PATH="/tmp/nobun:/usr/bin:/bin" python3 -m pytest ... -q
ERROR collecting tests/test_notifiers_claude_channel_554.py
E  ToolchainPromiseBroken: a CI job promised this prerequisite and it is absent:
   claude-channel runs under bun; no bun on PATH. ... Skipping here would report
   a pass for tests that did not run.
1 error in 0.14s                                             # exit non-zero — red
```

And it fails when the *code* is broken, which is the other half of "a job that cannot fail proves nothing". A deliberate type break in `channel.ts` (`const SOCK_PATH: number = 42`):

```
$ bunx tsc --noEmit -p .
channel.ts(172,28): error TS2339: Property 'includes' does not exist on type 'number'.
... 5 errors                                                 # exit 2
$ env SUPERTOOL_REQUIRE_JS=1 python3 -m pytest tests/test_notifiers_claude_channel_55{0,4}.py -q
5 failed, 6 errors
```

Reverted; on the real tree both are clean (`tsc` exit 0, `11 passed in 3.39s`).

### #546 — before → after

The before-state is the two historical Windows logs in the table at the top. The after-state is demonstrated two ways:

1. **In the suite, on every leg.** `tests/test_ci_encoding_546.py` writes a `junit.xml` holding a real em dash, runs the reporter under `PYTHONIOENCODING=cp437`, and asserts the UTF-8 bytes arrive and no `EF BF BD` does. A **control** test asserts that an *unpinned* `print()` in that same environment still loses the glyph — otherwise the sibling would pass against a reproduction that had quietly stopped reproducing, which is how this class of pin dies (the trap `tests/test_encoding_seam.py` names about locale pins).
2. **On a real Windows leg, and this is the load-bearing evidence.** A test whose assertion message holds a literal em dash was pushed to this branch, deliberately failing, and the resulting Windows logs inspected byte-wise. Run [30534415928](https://github.com/Digital-Process-Tools/claude-supertool/actions/runs/30534415928):

| job | U+FFFD (`EF BF BD`) in the whole log | the message as it arrived |
| --- | --- | --- |
| 90844130015 (windows-latest, 3.9) | **0** | `AssertionError: TEMP546 left — right` |
| 90844130035 (windows-latest, 3.12) | **0** | `AssertionError: TEMP546 left — right` |

Both channels carried it intact — pytest's own summary line *and* the indented re-emit from `junit.xml`:

```
FAILED tests/test_zz_temp_546_windows_proof.py::test_temp... - AssertionError: TEMP546 left — right (this failure is deliberate)
    AssertionError: TEMP546 left — right (this failure is deliberate)
```

Compare the before-state in the table at the top of this PR, same platform, same glyph: `\xEF\xBF\xBD`. The temporary test was removed in the force-push that followed; the run above is the only place it exists.

Also from that run: **the `notifiers` job passed on both ubuntu-latest and macos-latest** (jobs 90844129951 and 90844129913) while all twelve pytest legs were red — which is the shape it should have. It is a job whose result is independent of the Python suite's, and it has now been observed both green (here) and red (the broken-`channel.ts` demonstration above).

---

## Does the board get slower or flakier?

**Slower: no.** The critical path is unchanged. On run [30528861923](https://github.com/Digital-Process-Tools/claude-supertool/actions/runs/30528861923) the twelve legs were ubuntu 63–74s, macOS 82–126s, **windows 140–193s**, whole run 196s. The new job runs in parallel and finishes well inside that: 0.2s bun install from the lockfile, 0.6s `tsc`, 4.0s for the eleven tests, plus checkout / setup-python / pip. `bash -n` adds six subprocess calls to an existing 93s suite.

One number in the other direction, stated because it is real: a **failing** notifiers job takes ~167s rather than 4s, because several channel tests reach their socket timeouts when the server is broken. That is only ever paid on a leg that is already red, and it stays under the Windows critical path.

**Flakier: no new sources, and no new legs on the flaky platform.** Everything added to the twelve existing legs is `bash -n` — a parse; no clock, no socket, no network. The socket-spawning tests go on two new POSIX legs and not on Windows, which is where both recent flakes lived. The one recurring Windows flake (`test_xml_lint`, timing) was already fixed in `50a5ede` before this branch existed and is untouched. `-n0` on the new job is deliberate: eleven tests that each spawn a server gain nothing from xdist and lose determinism to it.

Suite stays green: **4945 passed, 36 skipped, 88.89% coverage** (gate 86%, `supertool.py`).

---

## New CI requirements, for whoever maintains this next

Documented in `docs/contributing.md` under **Text encoding** (a new rule 4) and a new **What CI runs, and what it does not** section:

- **`oven-sh/setup-bun@v2`** — the one new third-party action in this workflow, and Bun's own. The alternative considered was `npm i -g bun`, which trades a pinned action for an unpinned package install on every run. **This is the one thing in the PR that is a trust decision rather than a technical one, and it is yours** — if you would rather not add a third-party action, say so and I will swap it for the npm route.
- **`SUPERTOOL_REQUIRE_JS=1`** — set by the `notifiers` job and nowhere else. Setting it anywhere that does not install bun turns the channel tests into a collection error, which is the intended behaviour, not a bug.
- **`-X utf8` on any pytest invocation added to this workflow**, and **no** `PYTHONUTF8` / `PYTHONIOENCODING` in any `env:` block. Both are enforced by `tests/test_ci_encoding_546.py`, so a future edit that reverts either fails the suite rather than the next Windows red.

---

## Also found in that workflow, worth filing separately (not fixed here)

1. **The shared source walk cannot see `.github/` or `.githooks/`.** `tests/_repo_walk.is_machine_state` treats every dot-prefixed path component as machine state — right for `.venv` and `.pytest_cache`, wrong for the two git hooks this repo ships and asks contributors to install. Consequence: `.githooks/pre-push` sat outside `test_no_bare_python3_spawn.py`'s AST scan, which is how #572 (`PYTHON="${PYTHON:-python3}"`, in that exact file) survived the guard built to prevent it. `tests/test_ci_non_python_coverage_557.py` works around it with its own `git ls-files` walk and says why; the shared rule still has the hole.
2. **`junit.xml` is written and never uploaded.** Everything a reader gets is reconstructed from the log, which is subject to GH's 64KB truncation. `actions/upload-artifact` under `if: always()` would make the full traceback bodies — which `<failure>` carries as element text, and which this PR deliberately does *not* print, to keep the log small — available without re-running anything.
3. **`--tb=no` is load-bearing and undocumented as a trade.** It is why the summary line is all a reader gets, and why the `junit.xml` route matters. Worth a comment saying the traceback lives in the XML, rather than leaving the next person to wonder why CI shows less than their local run.

4. **A `CONFLICTING` PR gets no check runs at all, and that is a fourth distinct cause of the sentence #585/#587 just split into three.** This PR sat for six minutes with zero runs and `gh pr checks` reporting `no checks reported on the 'fix/546-557-ci-trust' branch`. Nothing had failed to trigger: master had moved, `mergeable` was `CONFLICTING`, and a `pull_request` event runs on `refs/pull/N/merge`, which GitHub cannot build for a conflicted PR. So "no checks" meant "rebase me", and both `gh-pr` and `git-status` would have said "none yet" — the one state that tells a reader to *wait*. `mergeable` is already fetched by `gh-pr`; classifying `CONFLICTING` + zero runs as its own case looks cheap, and it is the difference between waiting and acting.

Not merged. Both judgment calls are argued rather than assumed; the `setup-bun` question in particular wants your answer before this lands.


fdaviddpt:
## Board green, and the after-state measured on the platform that used to mangle

Run [30535365349](https://github.com/Digital-Process-Tools/claude-supertool/actions/runs/30535365349) — **14/14**, including the two new `notifiers` legs.

### Wall clock, measured rather than estimated

| leg | this run | baseline ([30528861923](https://github.com/Digital-Process-Tools/claude-supertool/actions/runs/30528861923)) |
| --- | --- | --- |
| **notifiers (ubuntu)** | **28s** | — (new) |
| **notifiers (macOS)** | **26s** | — (new) |
| pytest windows 3.9 — *the critical path* | 194s | 193s |
| pytest windows 3.10–3.12 | 143–149s | 140–143s |
| pytest macOS | 67–99s | 82–126s |
| pytest ubuntu | 68–77s | 63–74s |

The new jobs finish **166 seconds before** the leg that ends the run. The critical path moved by 1s, which is noise. `bash -n` across the six shell files is inside the existing suite and invisible in the totals.

### The Windows before → after, on real legs

The temporary em-dash failure went through run [30534415928](https://github.com/Digital-Process-Tools/claude-supertool/actions/runs/30534415928):

| | bytes in the log |
| --- | --- |
| **before** (job 90691288041, windows-latest 3.9, run 30485881190) | `...on this platform` **`EF BF BD`** `os.geteuid...` |
| **after** (job 90844130015, windows-latest 3.9) | `AssertionError: TEMP546 left` **`E2 80 94`** `right` — and **0** occurrences of `EF BF BD` in the entire log |

Intact in both channels, pytest's own summary line *and* the indented re-emit from `junit.xml`. Job 90844130035 (windows-latest 3.12) identical. The temporary test was removed in the following force-push; that run is the only place it ever existed.

### An unplanned demonstration, which is the better evidence

The first clean push went red on **windows-latest 3.9 and 3.10 only** — my bug, in the new test file: `_run_reporter` built a minimal `env={"PATH": ...}`, and CPython on those two versions reaches the Windows CryptoAPI through `SystemRoot`, so the interpreter died before running anything:

```
8 failing test(s), full messages from junit.xml:
FAILED tests.test_ci_encoding_546.test_the_reporter_keeps_the_glyph_a_cp437_console_cannot_encode
    AssertionError: b'Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python
      Python runtime state: preinitialized
      '
    assert 1 == 0
```

That is the new reporter's output, on a Windows leg, and it is the whole diagnosis: the fatal error, the runtime state, and the assertion — no truncation, no re-run, no guessing from a partial line. Under the old step the same red would have printed eight bare test IDs and nothing else. It took one read to fix (`b6cd2e9`: inherit `os.environ` and strip only the six encoding keys, which is `test_encoding_seam.py::_run`'s established pattern — an env allowlist means guessing what an interpreter needs to boot on a platform you are not on).

I would rather have shipped this without the self-inflicted red, but it is a cleaner proof of #546 than the deliberate one: an unplanned, genuinely platform-specific Windows failure, diagnosed from the log alone on the first read.

Still not merged — the `oven-sh/setup-bun@v2` question in the PR body is yours.

PR #627, opened 2026-07-30
fix: gh-issue-create/gl-issue-create decline a missing payload instead of crashing

### LINKED ISSUE #620: `gh-issue-create` with no payload argument dies on a raw `IsADirectoryError` traceback instead of saying what it needs
## What happened

Reaching for the op's own documentation, I typed the subcommand form by mistake:

```
$ ./supertool help gh-issue-create
--- help ---
ERROR: help needs an op name — help:OP (e.g. help:vim).
Run 'ops' for the full list.
--- gh-issue-create ---
FAIL (0.11s)
Traceback (most recent call last):
  File ".../presets/github/issue_create.py", line 147, in <module>
    sys.exit(main())
  File ".../presets/github/issue_create.py", line 66, in main
    payload = _load_payload(path)
  File ".../presets/github/issue_create.py", line 38, in _load_payload
    raw = Path(path).read_text(encoding="utf-8")
  ...
IsADirectoryError: [Errno 21] Is a directory: '.'
```

`help` handled its own missing argument correctly and said exactly what it wanted — that is the contract working. `gh-issue-create`, receiving no `@FILE`, defaulted the path to `.` and let `read_text` fail on a directory, so the user gets a five-frame stack trace whose most legible line is about `pathlib`.

## Why it is worth fixing

The information the reader needs is one line and the op already knows it:

```
ERROR: gh-issue-create needs a payload — gh-issue-create:@FILE (JSON or TOML with title/body).
```

That is the same three-state discipline this repo applies everywhere else, one layer out: **a step that cannot proceed must say what it cannot do and why**, rather than letting an implementation-level exception escape as the user-facing answer. A traceback is not a refusal — it discloses the failure but not the requirement, and it reads as "the tool is broken" rather than "you did not give me a file".

It also fails *loudly and unhelpfully*, which is worth distinguishing from this tracker's usual complaint. Nothing is silent here and nothing is wrong — the cost is purely that the message does not carry the fix.

## Adjacent, same call

Two things noticed in the same invocation, either of which may be the same underlying gap or may be separate:

1. **`'help gh-issue-create'` as a single quoted argument** was dispatched as an op named `help gh-issue-create` and rejected with `unknown operation`, followed by the short 17-op `Valid operations:` list. #614/#617 covers the wording of that list; this is noted only because both spellings of "I want the help" fail differently, and neither points at `help:OP`.
2. **The default path.** Defaulting a required payload argument to `.` is what turns a missing-argument error into a filesystem error. Whether the fix is "no default" or "validate before reading" is the implementer's call, but the default is the mechanism.

## Repro

```bash
./supertool gh-issue-create        # or any invocation with no @FILE
```

Filed on Florian's instruction: *"if info is missing from gh op, just post issue in supertool gh"* — this is the same rule one step over, where what is missing is the error message rather than the data.

Observed 2026-07-30 while filing #619.

fdaviddpt:
## Summary

Closes #620.

`gh-issue-create` / `gl-issue-create` invoked with no `@FILE` argument crashed with a five-frame `IsADirectoryError` traceback instead of saying what was missing.

**Root cause, not "no default path":** supertool's `{arg}` placeholder substitution (`_resolve_custom_op` in `supertool.py`) always fills in *something* — `file_arg = parts[1] if len(parts) > 1 else ""`. An omitted `@FILE` therefore reaches `issue_create.py` as `sys.argv[1] == ""`, not a missing argv slot, so the existing `len(sys.argv) < 2` guard never fires. `Path("").read_text()` then resolves to `Path(".")`, and the resulting `IsADirectoryError` propagates as a raw traceback — the exact failure `help` handles correctly for its own missing argument.

**Fix location — the call site, not the shared substitution.** `{arg}` defaulting to `""` is shared machinery used by every custom op in the config; changing its default would be a much larger, riskier change with unclear effect on other ops that may rely on an empty-string arg. The op-specific script is where "this payload is required" is known, so both `issue_create.py` scripts now validate the effective path (after stripping a leading `@`) *before* attempting to read it.

**Sibling check:** grepped every `@FILE`-style op (`presets/*.json` for `"@FILE"` syntax) — only `gh-issue-create` and `gl-issue-create` match; both had the identical defect (same `_load_payload`/`main()` shape, clearly copy-pasted). No other ops share this call shape.

**Three distinct failure messages, not one collapsed message** (judgment call from the issue):
- No payload at all (empty/missing/bare `@`): `ERROR: gh-issue-create needs a payload — gh-issue-create:@FILE (JSON or TOML with title/body).`
- Path is a directory: `ERROR: payload path is a directory, not a file: PATH` (previously: raw `IsADirectoryError` traceback, even for an explicitly-passed directory, not just the empty-arg case)
- Unparseable payload: parser message is now suffixed with the expected shape, e.g. `ERROR: failed to parse payload: ... (expected JSON or TOML with title/body)`

## TDD

Tests written first (`tests/test_issue_create.py`), confirmed RED — including a live-reproduced `IsADirectoryError` traceback matching the issue's repro exactly — then the fix, then GREEN. Full suite: 5114 passed, 61 skipped, 88.97% coverage (gate 86%).

Manually re-ran the original repro against the fixed scripts:
```
$ python3 presets/github/issue_create.py ''
ERROR: gh-issue-create needs a payload — gh-issue-create:@FILE (JSON or TOML with title/body).
$ python3 presets/gitlab/issue_create.py ''
ERROR: gl-issue-create needs a payload — gl-issue-create:@FILE (JSON or TOML with title/description).
$ python3 presets/github/issue_create.py .
ERROR: payload path is a directory, not a file: .
```
No traceback in any case.

## Docs

- `CHANGELOG.md` under `## Unreleased`.
- `docs/presets/github.md` and `docs/presets/gitlab.md`: added a short note under the existing "Default repo/project" sections documenting the new missing-payload / directory-payload behavior, since the payload requirement was previously only implied by the `syntax: @FILE` line.

## Test plan

- [x] New tests RED before the fix (traceback reproduced live)
- [x] New tests GREEN after the fix
- [x] Full suite green (5114 passed, 61 skipped)
- [x] Coverage gate met (88.97% vs 86% required)
- [x] Manual repro of the exact issue scenario against both scripts

Not merging — leaving for review per house process.


fdaviddpt:
## Windows fix, addressing the review

Confirmed your reading. **Reasoned, not observed** — I don't have a Windows box to run this on, so I'm relying on documented CPython/Windows behavior rather than a live repro: `CreateFileW` succeeds against a directory, the subsequent read fails with `ERROR_ACCESS_DENIED`, and CPython's `posixmodule.c` maps that to `PermissionError`/`errno.EACCES`, not `IsADirectoryError`. That's exactly consistent with the log — `except IsADirectoryError` never fires on Windows, so the original #620 traceback was still reaching users there.

**Judgment call — did not add `except PermissionError` alongside `except IsADirectoryError`.** That would have collapsed two different user errors into one false statement: a locked/wrong-ownership *file* also raises `PermissionError` on every platform, and on Windows it's the *only* thing a directory read raises — so catching it and printing "is a directory" would be confidently wrong for a real permission failure. Worse than the traceback, since the traceback at least doesn't lie.

**Fix:** both scripts now call `Path(path).is_dir()` and decide *before* attempting the read, so the verdict doesn't depend on which errno a given OS picks. `except IsADirectoryError` stays as a narrow TOCTOU fallback (path becomes a directory between the check and the read); a new `except PermissionError` reports a distinct "permission denied reading payload" message for a genuine unreadable file.

**TOCTOU weighed and accepted:** the window is a few instructions wide, and the failure mode if it's hit is "wrong specific message" (falls through to `IsADirectoryError`/generic-read-error handling), not data loss or a wrong action taken — same shape as the pre-check itself, just narrower. Didn't find a case where `os.path`/`pathlib` semantics differ platform-to-platform for a path vanishing mid-call in a way that would make this worse than the exception-catching approach it replaces.

**Sibling handlers swept** (asked #2): only `except IsADirectoryError` had this assumption. `FileNotFoundError`, `ModuleNotFoundError`, `subprocess.TimeoutExpired` are all cross-platform-safe as written — checked each, not just assumed.

**Incidental finding, not fixed here (out of scope for this ask):** `Path(body_file).read_text()` (github/issue_create.py:~105+21 after this diff) and `Path(description_file).read_text()` (gitlab/issue_create.py, same shape) have **no exception handling at all** — a missing/unreadable `body_file`/`description_file` still crashes with a raw traceback, same failure class, different field. Separate from the payload-loading path this PR fixes; flagging rather than expanding scope.

**Test teeth (asked #3):** `test_directory_check_happens_before_any_read_attempt` monkeypatches `_load_payload` to raise `AssertionError` if called at all for a directory path — proves the `is_dir()` check happens first rather than merely happening to produce the right message today. `test_permission_error_on_a_file_is_not_reported_as_a_directory` monkeypatches a `PermissionError` on a real (non-directory) file and asserts the message says "permission", not "is a directory", and never leaks a traceback. Both added to both `TestGitlabIssueCreate` and `TestGithubIssueCreate`. Full suite: 5118 passed, 61 skipped, 88.97% coverage.

Pushed to the same branch, same PR. Not merging, not polling CI.

Issue #397, opened 2026-07-27
conftest's autouse reset misses _REPO_ROOT_WALK_CACHE and _MUTATION_ATTEMPTS


fdaviddpt:
`tests/conftest.py`'s autouse `_disable_rtk_and_config` fixture resets module-level state between tests, but two globals added recently aren't on the list:

- `_REPO_ROOT_WALK_CACHE` — memoises the directory walk from an edited file up to its repo root (added with the #393 formatter opt-in gate)
- `_MUTATION_ATTEMPTS` — the attempted-mutation counter behind the `[branch: X]` footer

Both are module-level and survive across tests in the same process.

`_REPO_ROOT_WALK_CACHE` is the one that bites: it's keyed by directory, and `tmp_path` directories are unique per test, so today it mostly gets away with it. But the two tests that *do* care already work around it by hand:

```python
supertool._REPO_ROOT_WALK_CACHE.clear()
```

— in `tests/test_review_regressions_395.py`, `test_gate_follows_symlinks_to_the_real_repo` and `test_format_staged_applies_the_gate`. A workaround repeated in two places is the fixture telling you where it should have been.

The failure this sets up is the expensive kind: a test that passes alone and fails in suite order, or vice versa, with nothing in the test body to suggest why.

## Fix

Add both to the autouse reset in `_disable_rtk_and_config`, and drop the two manual `.clear()` calls now made redundant.

While in there, it's worth a look at whether the reset list should be derived rather than hand-maintained — every new module-level cache is a chance to forget this again, and the forgetting is silent.

PR #636, opened 2026-07-31
feat(read-ops): @payload route for colon-bearing patterns, and an error that names the split (#625)

### LINKED ISSUE #625: No escape route for `:` in read-op patterns (`grep`, `around`, `between`)
## Problem

Mutating ops have a documented escape for content containing `:` — the `:::` delimiter and the `@payload` route. Read ops (`grep`, `around`, `between`, `read:...:::grep=`) have no equivalent, so a search pattern containing `:` is guesswork.

## Real example

Searching CI output for a rendered element, the natural pattern is `Element: <`:

```bash
./supertool 'grep:Element: <:/path/to/traces.txt:8:0'
```

That tokenizes wrong — `Element` becomes the pattern and ` <` becomes the path. I ended up guessing at an escape:

```bash
./supertool 'grep:Element\: <:/path/to/traces.txt:8:0'
```

which happened to work, but I had no way to know that beforehand — `:::` is documented for `edit`/`replace`, not for `grep`, and the ops listing doesn't mention escaping for read ops at all.

This bites constantly on this codebase because the things worth grepping for contain colons by nature: PHP namespaces (`SiCore\Components::CONST`), assertion messages (`Failed asserting that:`), log lines (`ERROR: ...`), and time stamps.

Same session, I also hit it building alternations — `grep:passed|failed|flaky|Error:FILE:15` is ambiguous about where the pattern ends.

## Suggested fix

Give read ops the same payload route the mutating ops already have:

```bash
./supertool 'grep:@-' <<'EOF'
pattern = '''Element: <'''
path = "/path/to/traces.txt"
limit = 8
context = 0
EOF
```

This needs no new mental model — it's the exact pattern already used for `edit:@-` / `batch:@file`, so it's one rule applied consistently instead of two behaviours to remember.

A cheaper interim fix: document the `\:` escape for read ops in `ops` output, since it apparently already works.

## Why it matters

The colon-tokenization trap is already called out in our own internal notes for `git-commit` and `todo_refresh` (both of which lose data when an argument contains `:`). Read ops have the same trap with no documented way out, and unlike a mutating op the failure is silent — you get results for the wrong pattern rather than an error.

fdaviddpt:
Closes #625

## What I established before designing (and what it changed)

The issue reports three things. **Two did not reproduce**, and that redirected the fix.

**1. `grep:Element: <:PATH:8:0` does not mis-tokenize.** `_parse_grep_args` (and `_parse_around_args`) already peel the trailing integers from the right, take the **last** token as the path, and rejoin everything before it as the pattern. Both of the issue's examples parse correctly today and always did:

```
grep:Element: <:traces.txt:8:0        -> pattern 'Element: <'      1 result
grep:passed|failed|ERROR:log.txt:15   -> pattern 'passed|failed|ERROR'  3 results
```

That covers the alternation ambiguity too — the last branch may end in a colon; the peel is anchored on the right, so it does not matter.

**2. It is not silent.** A mis-parse produces `ERROR: path not found: X`, because the path is checked before the scan. The three-state discipline was already holding here. Both facts are now pinned by tests so they cannot regress.

I also could not reproduce my own report of a silent zero from `grep::> "$MEMORY_FILE":scripts/:10:3`. Rebuilt with the matching text present, it returns the match — the leading colon is handled, and `$` as a regex anchor is caught by `op_grep`'s literal fallback (`no regex match; showing 1 literal match(es)`). **That was a correct zero; the text I searched for was not there in that spelling.** I have left the silent-zero framing out of the changelog because it was not true of this op.

**3. `\:` is not an escape, by design or otherwise.** `_split_arg` splits on every `:` regardless; the backslash simply rides into the pattern, where the regex engine reads `\:` and `:` identically. It works *until* the literal fallback fires, at which point it searches for the backslash. Documenting it — the issue's suggested cheap interim fix — would have promoted an accident to a contract that breaks under fallback. I did not.

**Where the colon genuinely wins:**

- **No path to anchor against.** `grep:A::CONST` searches for `A:` in a path named `CONST`. There is nothing on the right to peel.
- **`between` rejoins the other way.** `between:re:START:END:PATH` does `":".join(parts[4:])`, so a colon in START or END steals from the *path* — the mirror image of grep's failure, in the same character.

## Judgment calls

**`@payload`, not a new escape character.** Read ops are typed interactively, so a payload file is heavy — which is why the colon CLI keeps working for everything it already handles, and the payload is only needed for the cases above. I did not add `\:` or repurpose `:::` as a pattern/path separator: both are tokenizer changes that put every pattern in the blast radius to fix two shapes, and `:::` in particular already means something adjacent for the mutating ops. `@payload` is the vocabulary that exists; this is one call site adopting it.

Deliberately **not** routed through `_AT_FILE_BUILTIN_DEFAULTS` — that registry rebuilds a positional parts list and hands it back to the colon parsers, which would re-split the very pattern the payload exists to protect. `_READ_OP_AT_FIELDS` dispatches straight to the op.

**The error is the more valuable half, and it is in this PR.** An escape nobody can find is not a fix. `grep:key: value` now says:

```
ERROR: path not found: ' value' (cwd: ...)
  Read as pattern='key' + path=' value' — i.e. 'key: value' split on ':'.
  If your pattern contains ':', that split is the likely cause. ...
    ./supertool 'grep:@-' <<'EOF'
    pattern = '''<pattern, colons and all>'''
    path = "<path>"
    EOF
```

Withheld when the leading argument has no `:` **and** the path still looks like a path, so `grep:alpha:nope.txt` stays unadorned — advice on every wrong path is advice nobody reads.

## Backward compatibility

Established, and pinned by tests:

- **No tokenizer change.** `_split_arg`, `_parse_grep_args`, `_parse_around_args` are untouched. Every pattern that parses today parses identically.
- **`grep:@Override:src/` still searches.** The payload route is gated on the reference *resolving* (`@-`, or an existing file), not on a leading `@`. An unresolvable `@pattern` falls through to the search it always meant — with or without a path.
- **One behaviour changes:** `grep:@-:PATH` now reports that `@-` takes no colon arguments instead of searching for the literal `@-`. Loud, self-explaining, and reachable via the payload.

Full suite: **5180 passed, 66 skipped.** New file: 38 tests. `_READ_OP_AT_FIELDS` added to `conftest.RESET_EXEMPT_GLOBALS` (#397 guard — it is a constant).

## Docs

`README.md`, `docs/input-forms.md` (new "When the *pattern* contains a colon" section — including the explicit statement that there is no backslash escape), `docs/operations/search.md`, `help:grep|around|grep_around|between|read` via both config files, and `CHANGELOG.md` under Unreleased.

Untouched: `presets/_checks.py`, `presets/github/pr.py`, `presets/gitlab/mr.py` (PR #629).


fdaviddpt:
Pushed `b8b964d` — you were right, and the diagnosis was exact.

**The tool was right; the test was wrong.** `between` symbol mode without tree-sitter declines by name and points at `between:re` — that is the three-state contract working. My test asserted the resolved body, so it only passed where the optional dep exists. The `enable_tree_sitter` fixture I reached for is not an availability gate: it clears supertool's cached detection flags so real detection can run. It cannot install a package. That is why it read as unused-but-declared, and it was the wrong mechanism.

**On your ask 1 — I took the second option, and it turned out to be stronger than "either result is fine".** Whether symbol mode *resolves* is not what #625 changed; whether the payload *routes* is. `op_between_symbol` checks `symbol`, then `path`, and only *then* reaches for tree-sitter — so a well-formed payload aimed at a **nonexistent** path lands on a deterministic `file not found` that echoes the path back, on every platform, with no optional dependency. That is reachable only if both fields survived the payload. Three tests now:

- `..._extracts_both_fields` — the nonexistent-path probe. Excludes `requires SYMBOL:PATH` (route dropped), `empty symbol`, `empty path`. No dep, runs everywhere.
- `..._reaches_the_resolver` — accepts the body **or** the named decline, while explicitly excluding the route/arity/`unknown field`/`needs 'symbol'` signatures. This is the "either outcome" shape you leaned to, kept narrow.
- `..._returns_the_symbol` — `skipif` no pack, with a reason naming why and where the routing is pinned instead.

**On ask 2 — vacuity checked by mutation, not by assertion.** I disabled the read-op `@payload` route (`and False`) and re-ran. Both platform-independent tests fail in **both** environments:

```
MUTANT, tree-sitter blocked:   13 failed, 26 passed, 1 skipped
MUTANT, tree-sitter available: 14 failed, 26 passed
  incl. ..._extracts_both_fields  and  ..._reaches_the_resolver
```

So the loosened assertion still dies if the payload stops parsing — which was the trap.

**On ask 3 — swept by re-running the whole suite against a simulated CI**, blocking `tree_sitter*` at a `sys.meta_path` import hook: **5168 passed, 90 skipped**. The one failure under that harness (`test_gofmt_check_graceful_skip_when_tool_missing`) is an artifact of the harness itself — it fails identically on `master` with the blocker and passes on both without it (2x2 checked).

Two other environment assumptions I went looking for and cleared rather than assumed:

- **Windows `@payload` paths.** `between:@C:\...\b.json` depends on the drive-letter merge in `_split_arg` seeing `@C`. `_DRIVE_LETTER` is `^@?[A-Za-z]$` — the `@?` is already there, put in for the mutating ops' `@file` route. Read ops inherit it; no change needed.
- **The stdin payload test** builds its TOML path with `json.dumps`, so a Windows path round-trips as a TOML basic string (`\\` -> `\`) rather than breaking on backslashes.

Also dropped an unused `capsys` from the stdin test. Clean suite on the branch: **5189 passed, 66 skipped**.

Not merging, not polling.

Issue #263, opened 2026-05-29
phpstan: op misses inheritance errors (property.extraNativeType) that pre-push catches at same level


fdaviddpt:
## Friction

During a framework typing pass on `FwkDistribution/shared-private/src2/Shared/1.0.0`, sub-agents ran `./supertool 'phpstan:<file>'` per file and got **CLEAN**, but the DVSI pre-push hook (PHPStan **fwk level 7** — same level) **blocked** the push with:

```
Property Shared\Components\Fields\Select\SelectOptionGroup::$label (string)
overriding property Shared\Components\ComponentGroupBase::$label should not have a native type.
🪪 property.extraNativeType (non-ignorable)
```

## Cause

The `phpstan:` op analyses the **single file in isolation**. `property.extraNativeType` (and similar LSP/inheritance-compat errors) require loading and comparing the **parent class** — which single-file scope doesn't do. So the op reports CLEAN on exactly the error class that matters most for framework base/child typing.

## Impact

False "CLEAN" confidence. Agents typed dozens of framework component properties believing they were safe; the inheritance incompatibilities only surfaced at `git push`. For autonomous typing runs this defeats the local gate.

## Ask

Make the `phpstan:` op catch what pre-push catches at the same level — i.e. analyse with enough context (parent classes loaded) to surface `property.extraNativeType` and related inheritance errors. If full parity isn't feasible per-file, document that the op cannot gate inheritance-compat and that pre-push / `phpstan2_full` is required for framework typing.

## Repro

Type a child property native where the parent declares it untyped (e.g. any `extends ComponentGroupBase` class typing `$label`). `phpstan:<file>` → CLEAN; pre-push → blocked.

fdaviddpt:
## Closing — the symptom is fixed, the residual is a config rule rather than a defect

[#499](https://github.com/Digital-Process-Tools/claude-supertool/pull/499) (`6e9fca2`) fixed the false CLEAN, but not the cause this issue names.

[...]

PR #637, opened 2026-07-31
fix(validators): an adapter's own unreadable reply is a skip, not a finding about the file (#634)

### LINKED ISSUE #634: `jsonlint` reports valid JSON as broken: an empty adapter response is read as a parse error in the file
## Every `.json` edit in this repo reports an error that is not there

```
$ ./supertool 'validate:presets/gitlab.json'
validate: presets/gitlab.json
jsonlint    : 1 err       (0ms)
     orchestrator  adapter bad json: Expecting value: line 1 column 1 (char 0)
git-status  : ok          (157ms)  +0 -0 clean
```

The file is fine:

```
$ python3 -c "import json; d=json.load(open('presets/gitlab.json')); print(len(d))"
3
```

Reproduced on a clean tree at `master` — no local edits, nothing staged.

## The message is about the wrong file

`Expecting value: line 1 column 1 (char 0)` is what `json.loads("")` raises. It is not a complaint about `presets/gitlab.json`; it is the **adapter's own response** being empty, parsed as JSON, and failing at byte zero. `supertool.py:10840`:

```python
"code": "orchestrator", "msg": f"adapter bad json: {e}"
```

So the tool ran a checker, got nothing back, and rendered that nothing as a finding about the user's file.

This is this repo's most-filed defect class arriving inside a validator — the same shape as #263, where empty stdout from a refusing PHPStan was read as `file_errors: 0`. **But it lands the other way up, and that is worse.** #263 turned a silence into a false clean; this turns a silence into a false *finding*. A missed error costs one bug. An invented error costs the credibility of every error the validator prints, and this one fires on every `.json` edit — the `0ms` timing is itself the tell that no checker ran at all.

## Why it has survived

It is cosmetic-looking, permanently present, and attached to a step that otherwise works: the edit lands, the rollback does not trigger, `git-status` says `ok`. I have watched this line go past repeatedly tonight while editing JSON payloads and read it as background noise about my own file. That is precisely the failure — a validator nobody believes is a validator that is not running, and the first *real* malformed JSON will print in the same position, in the same colour, and be skipped for the same reason.

## Judgment for whoever takes it

- **Find out why the adapter returns empty first, and fix that.** It may be a missing `jsonlint` binary, a wrong invocation, or an adapter that exits before writing. If the checker genuinely is not installed, the correct output is `skipped` with that reason, not `1 err` — `docs/validators.md`'s "Declining instead of guessing" is the canonical write-up and `refusal.py` already carries the vocabulary. **Check whether a sibling validator already handles the missing-binary case** before inventing anything; in this repo the abstraction usually exists and one call site has not adopted it.
- **Empty is never a finding.** Whatever the cause, an adapter response of zero bytes must not become an error attributed to the file under test. That rule is worth pinning with a test independently of the root cause, because it is the part that will regress.
- **The message must say whose JSON is bad.** "adapter bad json" reads as "your JSON is bad" to everyone who has not read `supertool.py:10840`. If the adapter's own reply fails to parse, say so in those words and name the adapter.
- **Do not trade the loud bug for the quiet one.** Suppressing this line so it stops appearing would convert a false positive into no coverage at all, silently. Ask which failure you are choosing.
- **Check the other validators for the same adapter path.** If the empty-response handling is shared, `phplint`, `xmllint` and the rest may all be capable of the same false finding; if it is per-adapter, say which ones you checked.

## Repro

```bash
./supertool 'validate:presets/gitlab.json'    # any valid .json in the repo
```

Found while verifying an agent's incidental report — it flagged the line as pre-existing and not its own, which was correct, and I confirmed it on a clean tree before filing.

Observed 2026-07-31 against `master @ 7eadf5e`.

fdaviddpt:
## The defect

`validate:presets/gitlab.json` printed, on a clean tree:

```
jsonlint    : 1 err       (0ms)
     orchestrator  adapter bad json: Expecting value: line 1 column 1 (char 0)
```

against a file `json.load()` reads happily. `Expecting value: line 1 column 1 (char 0)` is what `json.loads("")` raises, so it was never about the file — and the `0ms` was the second tell that nothing had run.

This is #263 inverted: that turned a silence into a false clean, this turned one into a false **finding**. A missed error costs one bug; an invented error costs the credibility of every error the validator prints, and this one fired on every `.json` edit.

## Root cause — established, not patched around

The adapter was not empty, missing, or crashing. It was never invoked.

This repo's own `.supertool.json` wired jsonlint to `{python} -m json.tool {file}` while the bundled, SCHEMA.md-conformant `validators/jsonlint/jsonlint.py` sat unused beside it — `docs/validators.md:74` had shown the correct entry all along. `json.tool` is a pretty-printer: it exits 0 and echoes the reformatted file, so the last stdout line is `}`, and the orchestrator parses the last line as a receipt. `json.loads("}")` raises exactly the reported error.

So this was **not** a missing binary and not an empty response — it was a `cmd` that is not an adapter. Verified directly: the bundled adapter run by hand on the same file emits a clean receipt and exits 0.

## Sibling check

- **Configured here:** `lsp-diag` and `git-status` are both wired to real adapters. Checked, they're fine. A new test asserts every configured `cmd` points at an adapter, so this cannot come back through config.
- **Shared path:** `_validator_run_one` is common to *all* validators, so `phplint`, `xmllint` and the rest were equally capable of this — the exposure was never jsonlint-specific, only its config was.
- **`refusal.py` / the third state already existed** and this call site had not adopted it, exactly as the report predicted. The fix adopts it rather than inventing vocabulary.
- **`.supertool.example.json`** ships two `_example-*` entries (`phpcs …`, `pytest -q …`) with the same wrong shape. They exist to illustrate `opt_in` and `resolve`, so I documented them as not usable as adapter commands rather than rewriting them here.

## Judgment calls

1. **Three states, not two.** An unspawnable `cmd` now reports `skipped` naming the binary. `1 err` there claimed the *file* was broken because a tool was not installed.
2. **Empty is never a finding** — pinned by tests independent of the root cause, since that is the part that will regress. Four exits now take the third state: empty stdout, non-JSON stdout, a reply that parses but carries neither `ok` nor `skipped`, and a spawn failure.
3. **The message names the adapter.** `jsonlint adapter replied with something that is not JSON — … — this file was not checked`. Tests assert the reason starts with the adapter name and never contains the file under test.
4. **Which failure I chose:** the loud one, kept loud. Muting the row would have traded a false positive for a silent absence of coverage. The row still prints on every run — it just no longer claims a fact about the file. And a **timeout is deliberately excluded** from the skip conversion: the binary exists and was invoked, and a tool that hangs is a validator failure that must stay red.

Found while writing the RED tests: `json.tool` on a *single-line* file prints `{}` — valid JSON, so the parse succeeded and the orchestrator returned that bare dict, which renders `0 err`, a **pass invented out of a reply carrying no verdict**. Same absence, other polarity, one line away. Covered by the no-verdict branch.

## Tests

`tests/test_validator_adapter_reply_634.py`, 13 tests, written first — RED was 11 failed / 2 passed, the 2 being the guards that had to already hold (a well-formed receipt with errors still reports them; a timeout stays loud).

Pinned: a valid file passes clean; the literal `presets/gitlab.json` repro passes; **a genuinely malformed `.json` still reports an error with a line number** (the disarm guard for judgment call 4); empty / unparseable / no-verdict / missing-binary each skip and cannot regress; the reason names the adapter, not the file; the skip row renders its reason.

Two existing tests in `tests/test_validators.py` asserted the old `no output → ok: False` contract — updated, not deleted, since that contract is the defect.

Full suite: **5155 passed, 66 skipped**.

## Staying clear of in-flight work

No changes under `presets/` — #629 and #636 are untouched. CHANGELOG conflict expected at rebase.

Closes #634


fdaviddpt:
## Windows legs: one product bug, one fixture bug

**(2) `test_wellformed_receipt_with_errors_still_reports_them` — fixture, not product.** Established before touching anything:

- `cmd` is split with POSIX-mode `shlex.split` on every platform. The fixture built its stub command from a raw `str(Path)`, so on Windows it split to `['python', 'C:Usersrunneradmintadapter.py']` — every backslash eaten. The stub was never spawned, stdout was empty, and the new skip branch correctly reported `produced no output`.
- **The product never does this.** `{supertool_dir}` is posix-ified at `supertool.py:317` (`.replace(os.sep, "/")`) and `{file}` is `shlex.quote`d, so a real Windows path arrives single-quoted and survives the split intact.
- **Observed corroboration, not just reasoning:** the three tests that drive the *real* repo spec — including `test_malformed_json_file_still_reports_an_error` — **passed on those same Windows legs**. jsonlint runs on Windows and reports real findings there. It is not disarmed, so judgment call 4 holds on every platform.

The fixture now quotes its stub path, with a comment saying why.

**(1) Missing-binary message — a real product gap, fixed at the source.** The name came from the `OSError` text, which POSIX provides and Windows does not. It now comes from the spec's `cmd` (`_validator_cmd_program`), and only `e.strerror` is quoted so POSIX stops repeating it:

```
jsonlint adapter could not be run: jsonlint — No such file or directory — this file was not checked
jsonlint adapter could not be run: jsonlint — The system cannot find the file specified — this file was not checked
```

**Three tests added so none of this is reasoned-only:** the Windows-shaped `OSError` is injected on every platform and the program must still be named exactly once; `_INSTALL_DIR` must contain no backslash; `shlex.quote` must round-trip a Windows path.

Full suite: 5165 passed.

PR #639, opened 2026-07-31
fix(output): end a mutating call with [result] N ops run, M writes (#621)

### LINKED ISSUE #621: Result summary prints before validators — `| tail` hides whether the op did anything
## Problem

Mutating ops print their result summary **before** the `[validators]` block. Piping through `tail` — normal when validator output is long — hides the one line that says what actually happened.

## Real example

Applying a fix across the tree:

```bash
./supertool 'replace:::<span class="far fa-info-circle" data-toggle="tooltip" data-placement="right"
          title="{$component->getTooltip()}" aria-label="{$component->getTooltip()}" />:::<span class="far fa-info-circle" role="img" ...:::Dvsi/dvsi-private/src2/' 2>&1 | tail -4
```

Output:

```
[validators]
git-status  : ok          (no new errors)   0.4s
[branch: max/a11y-aria-prohibited-attr]
```

No `Done:` line, so I concluded the pattern matched nothing and told my teammate those files were still unfixed. **They had been fixed — 11 files.** I only found out when `git status` later showed them modified, and had to issue a correction.

A `tail` on a genuinely 0-match replace looks identical, so there is no way to distinguish "0 matches" from "N matches, summary scrolled off".

## Suggested fix

Print the summary **last**, after `[validators]` — or repeat a one-line result in the footer next to `[branch: ...]`:

```
[validators]
git-status  : ok          (no new errors)   0.4s
[result] 11 replacements in 11 files
[branch: max/a11y-aria-prohibited-attr]
```

Either makes `| tail -N` safe, which matters precisely because long validator output is when you reach for `tail`.

## Why it matters

This is the failure mode where success reads as silence. It caused me to state something false to a teammate — worse than an error, because an error I would have investigated.

fdaviddpt:
## Reproduced first, exactly as filed

```
$ python3 supertool.py 'replace:::NOPE_NOT_THERE:::x:::/tmp/t621.txt' | tail -4

[validators]
git-status  : ok          (no new errors)   0.1s
[branch: max/...]

$ python3 supertool.py 'replace:::alpha:::gamma:::/tmp/t621.txt' | tail -4

[validators]
git-status  : ok          (no new errors)   0.1s
[branch: max/...]
```

Byte-identical. One wrote a file, one did not. Every part of the report holds; nothing in it needed pushing back on.

The batch case is worse than the report suggests. Three ops, one no-match in the middle: `tail -4` shows the **last** op's green validators block, so it reads as unambiguous success while the middle op silently did nothing.

## The ordering decision: a footer, not a move

The issue offered both. Moving the receipt below `[validators]` loses on three counts:

- it changes the ordering existing tests assert (`#381`'s assert `out.rstrip().endswith("[branch: …]")`) and breaks anything reading output positionally;
- in a batch it tears each op's receipt away from the op it describes, which is the one place the detail is load-bearing;
- the benefit — a trustworthy last line — is fully delivered by a repeated line, without any of that.

So the receipt stays exactly where it is (pinned by a test), and one line is repeated below the noise:

```
[validators]
git-status  : ok          (no new errors)   0.1s
[result] 3 ops run, 1 write
[branch: fix/621-result-after-validators]
```

Cost: one duplicated line on a short call. That is the whole price.

## The invariant, not the formatting rule

`docs/` and the tests pin the stronger claim from the issue comment:

> an op which changed nothing must not end with output that looks like an op which did

The headline test takes the rendered tail of a no-match call and of a real one and requires them to differ — asserted at the layer a reader actually reads, not against a function's return value. It fails on `master`.

It also normalises validator durations out of that comparison. Without that it passed under `xdist` purely because one run took `0.1s` and the other `0.2s` — i.e. it would have reported the bug fixed while the bug was entirely present. **That is a live hazard for any existing test comparing rendered blocks in this repo, and worth knowing beyond this PR.**

## Requested vs applied

`N` is `_MUTATION_ATTEMPTS` (bumped when a mutating op runs, before its outcome is known). `M` is `_WRITE_COUNT` (bumped at `_atomic_write`, decremented by `_retract_write`). Both are counters — the footer never re-reads the prose receipt. Consequences:

- **a validator rollback reports `0 writes`.** A reverted edit claiming a write would be trading a loud failure for a quiet one; tested.
- **in a batch, `3 ops run, 2 writes` exposes one no-match among two successes** — the single line that would have caught the #634 incident on its own.
- for the issue's own report it would have read `1 op run, 11 writes`.

## Declining instead of guessing

`N == 0` means the call never reached the mutation chokepoint. `0 ops run, 0 writes` would be an invented verdict, so **no line is printed at all** — per `docs/validators.md`'s three-state contract. Nothing is clamped, suppressed, or defaulted to tidy the output.

## Read-only ops: deliberately excluded

A `grep` returning zero *is* the same shape of problem, and I checked rather than assumed: a zero-result read op's own count line is already the last thing printed. Nothing intervenes between the summary and end-of-output, so there is no lie to correct. Adding a footer to the highest-frequency path would be noise bought against no defect. Tested as an explicit scope boundary (`read`, `grep` hit and miss, `wc`, `stat`, `glob`, and `replace_dry` — a preview writes nothing by design, so `0 writes` there would read as a failure).

## Tests

18 new tests in `tests/test_result_footer_621.py`.

- **RED (before the fix):** 10 failed, 8 passed. The 8 passing are the scope guards (read ops and `replace_dry` get no footer; the receipt stays above `[validators]`) — they must keep passing, and do.
- **GREEN:** 18 passed, serial and under `xdist`/random ordering.
- **Full suite:** 5235 passed, 66 skipped, coverage 88.97% (gate 86%). No existing test needed changing.

**Mutation testing**, 11 hand-built mutants over the new code — decline boundary, both pluralisations, the `writes == 0` gate and its inversion, dropping the footer call, swapping `[result]`/`[branch:]` order, swapping each counter argument for the other, widening the depth gate so a batch emits one footer per sub-op, and removing the gate entirely:

**11/11 killed.** 10 die against the new test file alone; the eleventh (`if True:` for the gate) is equivalent *for `_result_line`* — it declines at `ops == 0` — and is killed by the pre-existing #381 read-op tests, since the same gate guards `[branch:]`.

## Docs

- `CHANGELOG.md` under `## Unreleased` → `### Fixed`, with the issue link and the reasoning for choosing a footer over a move.
- `docs/operations/edits.md` — the footer contract, replacing the old one-line `[branch: X]` note. Scripts parse this output, so the contract change is documented explicitly, including that the receipt has **not** moved.
- `README.md` — short note in the validators section, since that block is what hides the receipt.

## One thing nobody has filed

Batch sub-op headers are not re-runnable. A sub-op supplied through the `@payload` route — used precisely because the content contains `:` — is echoed back with **single** colons:

```
--- replace:alpha:ALPHA:/tmp/t621.txt ---
```

Copy-pasting that header runs a different op, or fails to parse. Same family as this bug: output that reads as a faithful record of what ran, and is not. Not fixed here — it wants its own issue and its own judgment about whether the header should render the payload route or be dropped for payload-sourced ops.

Closes #621


fdaviddpt:
## Windows red: fixture, not product. `py-syntax` is exonerated.

Pushed `6d6e313`. Diagnosis first, since the hypothesis on the table was that a validator had gone quietly blind — that would have been serious, and it is not what happened.

### `py-syntax` cannot have the #637 defect

It is not a subprocess adapter. `_BUILTIN_SYNTAX_VALIDATORS["py-syntax"]` is `{"builtin": "python"}` — an in-process `compile()` (supertool.py:9845: *"The check is in-process (`compile()`), so it costs microseconds and cannot itself be stale"*). No `shlex`, no argv, no path quoting. There is no route by which a Windows backslash could reach it.

And `py-syntax : ok` was **correct, not a false clean**. The edit under test is `a = 1` → `a = 2`. That is valid Python. `py-syntax` was never supposed to fail here — the validator that was supposed to fail is the fixture's own `fake`, a different row.

### What actually happened, reproduced locally

Point a `rollback_on_fail` validator at a command that cannot be spawned:

```
[validators]
fake        : skipped     (fake adapter could not be run: C:/definitely/not/here/nope.sh — No such file or )   0.0s
py-syntax   : ok          (no new errors)   0.0s
[result] 1 op run, 1 write

file on disk: 'a = 2\n'
```

That is the Windows output, exactly. And **every layer behaved correctly**: the validator declined by name rather than guessing (the #634 three-state contract), a skip does not roll back by design, the edit therefore stood, and `[result] 1 op run, 1 write` reported the truth. The footer this PR adds was right; the test's premise was false.

### The fixture, broken three ways

```python
script.write_text("#!/bin/sh\n...")
script.chmod(0o755)
{"cmd": str(script), ...}
```

No `/bin/sh` on Windows, no shebang handling, and `str(Path)` is eaten by POSIX-mode `shlex.split` — the #637 shape, in my own test. `tests/test_validators.py::_counter_cmd` already documents the fix one file over: *"Uses a Python wrapper (no shell) so it works under shell=False dispatch."*

### The quieter instance, also fixed

`_noisy_validator` used `cmd: f"echo '{OK_PAYLOAD}'"`. On Windows `echo` is a cmd.exe builtin, not an executable — so that validator was **also** being skipped, and those tests passed only because the built-in `py-syntax` happened to supply the `[validators]` block they needed. Green for a reason other than the one claimed. Same class as the timing-jitter flake already called out in the PR body, and worth fixing before it becomes load-bearing.

Both now use `{python}` + `as_posix()`. The rollback test additionally asserts `"skipped" not in out` — so a fixture that stops running can never again leave the test vacuously passing rather than loudly failing.

### Verification

`5235 passed, 66 skipped`, coverage 88.97%, locally. No product code changed in this commit — `supertool.py` is untouched since `dc89123`.

No new issue to file: there is no product bug here. The batch-sub-op-header finding in the PR body still stands and is still unfiled.

PR #683, opened 2026-07-31
fix(batch): a payload sub-op header no longer renders as a different op (#644)

### LINKED ISSUE #644: Batch sub-op headers are not re-runnable: a `@payload` op is echoed back with single colons, so copy-pasting it runs a different op
## Problem

A batch sub-op header echoes the op back as a colon-CLI string:

```
--- replace:alpha:ALPHA:/tmp/t621.txt ---
```

When the op was supplied via `@payload`, that rendering is wrong. The payload route exists precisely because the content contains `:` — that is the whole reason the caller used it. Flattening it back to single colons produces a string that either parses as a **different op** or fails to parse at all.

## Why it matters

The header reads as a faithful record of what ran. It is not one. Copy-pasting it — the natural thing to do when re-running or reporting a step — silently executes something else.

Same family as #621: output that presents itself as an accurate account of an operation and is not. Here the loss is fidelity rather than visibility, but the consequence is the same shape, and it is worse in a bug report, where the header is the thing a reader trusts to reconstruct what happened.

Related: #625 shipped the payload route for read ops, so the number of ops whose real invocation cannot be rendered as a colon string is now larger, not smaller.

## The judgment call

Two options, and they are not obviously equivalent:

1. **Render the payload route** — echo `replace:@payload` (or the payload's own fields) so the header is honest about how the op was invoked.
2. **Drop the flattened header for payload-sourced ops** and label them, on the grounds that a faithful one-line rendering may not exist.

Whichever, the invariant worth pinning is that a header must never be a runnable string that runs something other than what ran. If it cannot be made re-runnable it should not look re-runnable.

## Provenance

Found by the #621 agent in the same code path, reported unprompted and deliberately left out of that PR's scope.

fdaviddpt:
A batch sub-op supplied via `@payload` was echoed back as its fields joined
on ':'. The payload route exists *because* the content contains ':', so the
join does not merely lose information — the result parses as a different op.
`replace` of `time: 10:30` printed

    --- replace:time: 10:30:time: 11:45:/tmp/h.txt ---

which runs cleanly when pasted, as a replace of `time` with `10` in a file
named `30`. The issue reported it as "fails to parse"; it does not, which is
the worse case. Same family as #621 — output presenting itself as a faithful
account of an operation and not being one.

Two rendering paths existed and only one lied: under _HEADER_ARG_MAX the arg
was verbatim, over it #384's elided form took over — but that swap is gated
on the op having written, so a failing long payload op fell back to the lie,
at the one moment the header is the only surviving record.

Payload sub-ops now print `op:@payload → target` at any length, success or
failure. No faithful one-line colon rendering of a payload op exists, so this
does not fake one; `@payload` does not resolve, and re-dispatching it says so
in as many words rather than reading as a missing file.

Found alongside, in the same lines and worse: batch built read-only sub-op
args with `sorted(_item)` — alphabetical key order, which is not any op's
argument order, and is the arg that runs, not a rendering. `between` with
symbol+path dispatched as `between:<path>:<symbol>`. #625 had just shipped a
payload route for read ops so a ':' in a pattern would survive; batch undid
it on the way in. Every existing test passed because all use a single `path`
field, where the two orders coincide. Read ops now dispatch straight to the
op; the rest use declared order, or decline rather than guess.

Tests assert the round trip — a header fed back to the dispatcher must run
the same op or not be a runnable op at all. A string-shape assertion passes
while the header still lies.

Closes #644

Co-Authored-By: Max <noreply>


fdaviddpt:
Maintainer verification of the mis-dispatch this PR fixes, run against `master` before merging — it is worse than the issue and worth recording on the PR.

Both reproduce the alphabetical-key-order dispatch:

```
$ ./supertool 'batch:@-' <<'X'
[[ops]]
op = "read"
path = "/tmp/ord.txt"
offset = 5
limit = 2
X
--- read:2:5:/tmp/ord.txt ---
ERROR: argument parsing: invalid literal for int() with base 10: '/tmp/ord.txt'
```

```
$ ./supertool 'batch:@-' <<'X'
[[ops]]
op = "grep"
pattern = "def main"
path = "supertool.py"
limit = 3
X
--- grep:3:supertool.py:def main ---
ERROR: path not found: 'def main'
  Read as pattern='3:supertool.py' + path='def main' …
  If your pattern contains ':', that split is the likely cause. The colon CLI cannot tell
  where the pattern ends; a payload can:
    ./supertool 'grep:@-' <<'EOF'
```

`limit`/`offset`/`path`/`pattern` sorted alphabetically, fed in as positional args.

**The second one is worth reading twice: the error advises using the payload route, and the payload route is what produced it.** The advice is circular in exactly the case it fires. #625 shipped that route so a `:` in a pattern would survive, and the batch loop re-serialised it away on the way in — so the remedy the tool recommends is the thing that broke.

Not asking for a change here; the fix in this PR removes the condition. Noting it because if the advice text is ever revisited, this is the case that shows it can point at its own defect.

Issue #619, opened 2026-07-30
`gh-pr:N:status` gives a check tally but never names the legs, so the one rule that gates a merge needs raw `gh`


fdaviddpt:
## The op answers "how many", and the merge rule asks "which"

`gh-pr:616:status` renders:

```
checks: 14 total: 14 passed, 0 failed, 0 pending
```

That is the right arithmetic — it is what #454 was filed to get, and it is what makes the "counts must sum to the leg count" rule checkable at all. But the moment the tally is *not* all-green, the next question is fixed and the op cannot answer it:

```
#617 | state: OPEN | mergeable: MERGEABLE | conflicts: no
checks: 14 total: 1 passed, 5 failed, 8 pending ⚠ NOT ALL GREEN
```

Which five? The rule I hold myself to is that **any leg not `SUCCESS` gets named before merging**, because a Windows-only red is usually a real bug and an ubuntu-only red is usually a flake — and those two lead to opposite actions (read the failure vs. `gh run rerun --failed`). The op tells me the situation is bad and leaves me exactly one call short of knowing what to do about it.

So tonight I fell back to:

```bash
gh pr checks 617 --json name,state -q '.[]|"\(.state)\t\(.name)"' | sort
```

which is the hand-rolled shape #454 exists to discourage, and which I then had to follow with a second fallback (`gh api …/actions/jobs/<id>/logs`) to get the assertion. Three calls where the op could plausibly have been one.

## What would close it

Names for the legs that are not `SUCCESS`, bounded the way the rest of this codebase bounds disclosures — the failing and pending sets are usually small, and when they are not, a `+N more` is the established vocabulary (#605):

```
checks: 14 total: 1 passed, 5 failed, 8 pending ⚠ NOT ALL GREEN
  failed:  pytest (ubuntu-latest, 3.9), pytest (ubuntu-latest, 3.10),
           pytest (ubuntu-latest, 3.11), pytest (ubuntu-latest, 3.12),
           pytest (macos-latest, 3.12)
  pending: notifiers (bun + TypeScript) (macos-latest), +7 more
```

Worth deciding rather than assuming:

- **Whether `:status` is the right home, or whether the names belong only in the fuller `gh-pr:N`.** `:status` is the terse form and is read in a loop while waiting; a five-line block on every poll may be worse than the fallback.
[...]

Issue #454, opened 2026-07-28
gh-pr/gl-mr drop CANCELLED checks from the count, so a failed run reports as "0 failed, 0 pending"


fdaviddpt:
## What happens

`gh-pr:452` on a PR whose run **failed**:

```
Checks: 10 passed, 0 failed, 0 pending
Mergeable: yes
```

The truth, from the same API:

```
pytest (macos-latest, 3.9)   status=COMPLETED conclusion=CANCELLED
pytest (macos-latest, 3.12)  status=COMPLETED conclusion=CANCELLED
... 10 × SUCCESS
```

and the run itself:

```
runs/30371892593 → status=completed conclusion=failure
```

Twelve legs exist. Two were cancelled. They are counted as **neither** passed, failed, nor pending — they simply vanish from the tally. `0 failed, 0 pending` says every check is accounted for and none is outstanding, which is the one reading that is definitely wrong.

## Why this is the worst place for it

This is the same defect as [#445](https://github.com/Digital-Process-Tools/claude-supertool/issues/445) — a tool reporting a non-green state in a way that reads as green — except one layer up and in the **merge-decision path**. `Checks: N passed, 0 failed, 0 pending` next to `Mergeable: yes` is precisely the output a maintainer reads before merging.

It already produced a wrong report: I told the maintainer this PR was at "10/12, nothing failing, waiting on two legs" — because `0 failed, 0 pending` combined with a 10-count reads as *two still running*. They were cancelled, the run had already concluded `failure`, and nothing was going to arrive. Had the count been 12 instead of 10, or had I not been handed the job URL, the next step was a merge on a run that never completed.

A cancelled check is not a passing check and it is not a pending one. It is *absence of evidence*, and the output must be able to say so.

## Scope

- `gh-pr` and, almost certainly, `gl-mr` — GitLab pipelines have `canceled` (one `l`) plus `skipped`, `manual` and `created`, which raise the same question. Check both rather than fixing the one that was reported.
- The same tally probably feeds `gh-prs`/`gl-mrs` board rows and the `failed`/`nopipe` filters. A board that sorts failing-first cannot see a cancelled run if the count discards it — worth checking whether `nopipe` already covers this case or whether cancelled runs are simply invisible there too.
- Any state that is neither success nor failure nor in-progress needs a decision, not a default: `neutral`, `skipped`, `timed_out`, `action_required`, `stale`. `timed_out` in particular must never land in the same bucket as `skipped`.

[...]

PR #728, opened 2026-08-01
fix(gh-pr): reconcile the check tally against the legs the run declares (#724)

### LINKED ISSUE #724: `gh-pr:N:status` under-counts legs after a partial re-run — 9 of 14, and the five missing ones are invisible rather than pending
Hit live in the merge path this morning, on PR #715. This is the defect #454 exists to prevent, arriving through a route #454 does not cover.

## What happened

Before the re-run, the op agreed with GitHub:

```
$ python3 supertool.py 'gh-pr:715:status'
checks: 14 total: 13 passed, 0 failed, 1 pending ⚠ NOT ALL GREEN
```

One leg had hung, so I cancelled the run and re-ran only the failed job (`gh run rerun <id> --failed`). Immediately afterwards, the same call:

```
$ python3 supertool.py 'gh-pr:715:status'
checks: 9 total: 8 passed, 0 failed, 1 pending ⚠ NOT ALL GREEN
```

Nine. The run had not lost any jobs — three other sources, taken in the same minute, all say fourteen:

```
$ gh run view 30687891233 --json jobs        # 14 jobs: 13 completed/success, 1 in_progress
$ gh pr checks 715 --json name,state -q 'length'
14
$ gh pr checks 715 -q 'group_by(.state)|map({(.[0].state):length})|add'
{"IN_PROGRESS":1,"SUCCESS":13}
$ gh pr view 715 --json statusCheckRollup -q '.statusCheckRollup|length'
14
```

So thirteen legs had passed and the op reported eight, out of a total it reported as nine.

## Why this one is worse than a cosmetic miscount

**The arithmetic check cannot catch it.** The rule this repo runs on is that the state counts must sum to the leg count — `8 + 0 + 1 = 9`, and the header says `9 total`. It is *internally consistent*. The only way to notice is to already know the matrix is fourteen legs wide, which is exactly the knowledge the op exists to save you having.

**And the failure mode is a bad merge, not a bad read.** The one pending leg was about to go green. Had I not counted by hand, the next call would have read `9 total: 9 passed, 0 failed, 0 pending` — no warning marker, everything summing — and my merge gate says merge on that. Five legs would have been invisible, not pending. If any of them were red, the op would have said nothing about it.

That is the house defect in the one place it does the most damage: **an absence the tool produced, rendered as a complete answer.** #445/#454 were the same surface counting `CANCELLED` as neither pass nor pending. This is a step further — legs that are not counted at all.

## What I did not verify

The cause. I have the observation from four sources and no diagnosis, and I would rather file it that way than guess in a direction that shapes the fix. Worth checking, in no particular order:

- whether the op reads check-**runs** for the head sha rather than the run's jobs, and a re-run mints new check-run IDs while some older ones drop out of whatever window is being read;
- whether a de-duplication by name is collapsing entries;
- whether pagination is involved (fourteen is close to nothing, but the shape is worth ruling out);
- whether the count settles on its own once the re-run completes — it may be a transient the API exposes only mid-re-run, which would make it *harder* to notice, not easier, because it appears exactly during the window when someone is watching a re-run to decide whether to merge.

The reproduction is cheap: cancel a run on an open PR, `gh run rerun <id> --failed`, then call the op while the re-run is in flight.

## The fix has to include a floor, not just a corrected count

Whatever the cause, the op should be able to say *how many legs it expected*. A count that can silently shrink is only safe if a shortfall is disclosed — the three-state contract from `docs/validators.md` §"Declining instead of guessing", applied to the tally itself: if the number of legs found does not match the number of legs the run declares, say so rather than reporting the smaller number as the whole.

Related: #454 and #445 (the same surface, cancelled legs), #619 (`:status` names no legs, so a shortfall cannot be spotted by reading), #722 (the hang that made me re-run in the first place), #723.

fdaviddpt:
Closes #724

## Reproduced live — and it disproved my own first fix

`gh run rerun --failed` on this PR's own run, three cycles, polling both sources concurrently:

```
11:28:20  rollup=0   latest_attempt=0   all_attempts_distinct=14
11:28:24  rollup=11  latest_attempt=11  all_attempts_distinct=14
11:28:32  rollup=14  latest_attempt=14  all_attempts_distinct=14
```

GitHub withdraws the old attempt's check runs and re-creates them over roughly twelve seconds. Read mid-way, `statusCheckRollup` is genuinely, transiently short. Same shape all three cycles.

## Pushback: the op was not under-counting

It de-duplicates nothing, paginates nothing, reads exactly one source, and `github_states()` is length-preserving by construction — so three of your four candidates are ruled out by reading the code, and the fourth is the one the capture above confirms. **The tally was faithful; the payload was wrong.** That is a report about GitHub, not about this repo.

It is still worth fixing, and the fix is still a floor — a tool that can be handed an incomplete answer has to be able to say the answer is incomplete, whoever's fault the incompleteness is.

## And pushback on your evidence, which I initially built on

**`gh run view --json jobs` was not a reliable second source.** That endpoint is `runs/{id}/jobs`, which defaults to `filter=latest` — the middle column above, and it dips in lockstep with the rollup. My first commit used exactly that as the floor, and it would have stayed silent through the entire window it exists for. The reproduction is what caught it; without it I would have shipped a disclosure that never fires.

Your four corroborating sources do not contradict this. Serial `gh` calls spread across a window that resolves in ~12s are as consistent with "the window had already closed" as with "the jobs API led the rollup." Three of the four (`gh pr checks` twice, `gh pr view --json statusCheckRollup`) read the rollup anyway, so they are one source, not three.

**What actually holds is `filter=all`.** A previous attempt's job rows are history and cannot be withdrawn, so the count of *distinct job names across every attempt* only ever grows — flat at 14 through samples where both other sources read 0. Counted as names, not `total_count`: under `filter=all` that is every row of every attempt (42 across three attempts here), which would have declared a complete tally 28 legs short.

## The fix

The rollup is reconciled against the number of distinct job names the Actions run declares across **all** attempts (`repos/{o}/{r}/actions/runs/{id}/jobs?filter=all`).

```
checks: 9 total: 8 passed, 0 failed, 1 pending ⚠ NOT ALL GREEN ⚠ INCOMPLETE — 9 of 14 legs read
  not read: pytest (windows-latest, 3.10), pytest (windows-latest, 3.11), notifiers (bun + TypeScript) (macos-latest), +2 more — this tally describes 9 of 14 legs and is not a merge signal. GitHub re-creates check runs during a partial re-run, so re-running the op usually settles it.
```

- `⚠ INCOMPLETE` is a **second** marker, not a reworded `⚠ NOT ALL GREEN`. They are different claims — "the legs I read were not all green" versus "I did not read all the legs" — and the all-passing shortfall, the case that was completely silent, carries only the second.
- **Unestablished count declines.** No jobs API, or more than 4 distinct runs on one PR, renders `⚠ TALLY UNVERIFIED` and names what it could not establish. Defaulting to `found` restores exactly the silence being fixed; defaulting to the larger number invents legs and trades a loud failure for a quiet one.
- **`declared < found` is not a shortfall.** A rollup carrying external CI or legacy commit statuses has extra checks, not missing ones. A rollup naming no Actions run at all reconciles silently and at zero cost.

## The cost, stated

One extra `gh api` call per distinct run. Measured on this repo: `:status` goes from ~1.1s to ~2.0s. Roughly double.

I think that is the right trade and it is the one place I would push back on myself: `:status` exists to be the thing a merge is decided on, and a merge decided on nine of fourteen legs is what the second is buying out. The run *ids* are free — they ride on the same `detailsUrl` that #619 already parses job ids from. **`gh-prs` and `git-status` are deliberately not reconciled**: paying that per PR on a board is a different trade, and neither line is a merge gate. Say the word if you want it behind a flag instead.

## On #619

**Already closed and shipped** — `named_disclosure()` names the non-passing legs in both the terse and full forms today. The brief described it as open. So it did not need doing, but the two do interact and the interaction is now covered: the disclosure names legs it *has*, and the shortfall line names legs it *doesn't*.

It also cost one change in #619's own test file. `test_slim_status_all_green_adds_no_extra_lines` monkeypatches `subprocess.run` wholesale, so unstubbed its `gh` double answers the new jobs call with a PR payload and the tally declines as UNVERIFIED. **The test still passed** — 445 bytes, under its 500-byte bound — while measuring the length of a decline notice instead of the terseness it claims to pin. It now stubs the leg count and asserts `⚠` is absent, which is what it always meant.

## Tests — RED then GREEN

RED: 22 failures. GREEN: 22 passed, full suite 5720 passed / 66 skipped, coverage 89.25% against the 86% gate.

**Every rendering test runs the op twice on a byte-identical nine-leg rollup and varies only the declared count, then asserts the two outputs differ.** That is the bar you set — a test asserting `9 total` passes on the broken code, which is precisely what shipped. Code that ignores the declared count emits identical output in both halves and every one of those tests fails.

Also pinned: an all-passing shortfall carries a `⚠`; a reconciled all-green adds no lines at all (the terseness guarantee survives); #454's sum-to-total contract still holds on the legs that *were* read; `_declared_legs` returns `(None, [])` on gh failure, on a payload with no `total_count`, and without spending a call when there are no run ids.

## Two commits on purpose

`a996e23` is the fix as I first reasoned it, floor and all. `113d42d` is what the live reproduction forced. I left them separate because the second one is the interesting one: the reproduction did not confirm my design, it broke it, and the only reason the disclosure fires at all is that I ran the repro you said was cheap instead of shipping on the reading.


fdaviddpt:
Post-merge note, for the record — I watched the op through this PR's own full CI run, because the thing that would make this fix worse than useless is crying wolf during ordinary progression:

```
11:55:16  checks: 14 total: 1 passed,  0 failed, 13 pending ⚠ NOT ALL GREEN
11:57:23  checks: 14 total: 9 passed,  0 failed,  5 pending ⚠ NOT ALL GREEN
12:03:45  checks: 14 total: 13 passed, 0 failed,  1 pending ⚠ NOT ALL GREEN
12:04:27  checks: 14 total: 14 passed, 0 failed,  0 pending
```

`14 total` from the first poll onward. A first run creates its check runs and its job rows together, so the two sources agree the whole way down; the divergence only exists inside the re-run window. No `⚠ INCOMPLETE`, no `⚠ TALLY UNVERIFIED`, and the reconciled green is byte-identical to what master printed before this.

Issue #746, opened 2026-08-01
The remaining ~78 bare `["ok"] is <bool>` assertions, in the pure-Python and monkeypatched adapter tests


fdaviddpt:
The tail of #725/#717's sweep, filed rather than swept so the boundary is a decision on the record instead of a place someone stopped.

## What was done

The PR for #725/#717 introduced `tests/_adapter_verdict.py` and converted **68 sites across 15 files** — every test that asserts a verdict from an adapter spawning an **external binary**, plus the two sites the issues were about. A guard (`test_a_file_that_adopts_the_convention_adopts_it_everywhere`) walks the AST of any test file importing the helper and fails on a message-less `["ok"] is <bool>` remaining in it, so those files cannot regress.

## What was left, and why

Roughly 78 sites in files whose adapters are **pure Python or fully monkeypatched**:

`test_inilint.py` (11), `test_jsonlint.py` (10), `test_yaml_check.py` (11), `test_tomllint.py` (6), `test_validators_git_status.py` (7), `test_validators.py` (7), `test_formatters.py` (4), `test_env_field.py` (3), `test_rector_mcp_engine_glitches.py` (3), `test_validators_skip_state.py` (5), `test_validators_phpstan_scope_263.py` (3), `test_lsp_diag_staleness_482.py` (2), and six files with one apiece.

The argument for the line: #725 and #717 are both a red that appeared **once, on one CI leg, on a platform the maintainer does not have, and never again**. In that situation the single occurrence is the entire diagnostic budget and a bare boolean spends it on nothing. A pure-Python adapter has no such failure mode — its verdict is a function of its input, so a red reproduces on the first local re-run and the cost of an opaque message is one command, not one quarter.

That is a difference of cost, not of correctness. These assertions are still worse than they need to be, and the fix is now one import and one call each.

## Why it is not just "run the codemod again"

Two things to decide first, neither mechanical:

- **`assert_declined` on a graceful degrade.** Several of these files assert `ok is True` on a path where the adapter *skipped* the file (tool missing, unsupported extension) rather than checked it. `ok=True` with `count=0` is the same payload for "clean" and for "not examined", which is a real ambiguity these tests currently paper over — converting them would be a good moment to notice it rather than to formalise it.
[...]

Issue #745, opened 2026-08-01
phplint maps any non-zero exit from `php -l` to `code: "parse"`, so PHP failing is indistinguishable from the file being bad


fdaviddpt:
Found while working #725, filed rather than folded in because it changes what a verdict *means*.

`validators/phplint/phplint.py` handles three named failures — no file arg, `FileNotFoundError`, `TimeoutExpired` — and then:

```python
if r.returncode == 0:
    emit({... "ok": True ...})
    return
out = (r.stdout or "") + (r.stderr or "")
m = re.search(r"on line (\d+)", out)
...
err = {"line": line, "col": None, "severity": "error", "code": "parse", "msg": msg}
```

Every remaining non-zero exit becomes `code: "parse"` — a statement that **this file has a syntax error**. But `php -l` exits non-zero for reasons that are not about the file: a failed extension load, a broken `php.ini`, a shimmed binary that is not PHP, a PHP that dies before it reads the argument. The regex already knows the difference — it looks for `on line N` and finds nothing in those cases — and the code discards that signal, keeping `code: "parse"` with `line: None`.

So a caller reading the verdict is told the file is bad when the truth is that the tool is. That is the same trade `docs/validators.md` §"Declining instead of guessing" refuses everywhere else in this repo, and it matters most exactly where it is hardest to check: a runner image whose PHP is subtly wrong reports as a syntax error in a two-line file that is fine.

## Why it is filed and not fixed

It is a **schema-visible** change. `code: "adapter"` versus `code: "parse"` is what a caller branches on, and #725 wanted a diagnostics change that could land without touching adapter semantics. Deciding the rule — no `on line N` match plus no recognisable PHP parse banner means `code: "adapter"`, msg naming the exit code and the raw output — is a separate call from the one #725 needed made.

## What to check with it

The same shape is likely in the sibling adapters: any `validators/*/*.py` that treats "the tool exited non-zero" as "the file is bad" without checking that the tool actually said something about the file. `ruby-check` is the one to look at first, since #717 lists "an actual defect in the adapter's verdict logic" among its unranked candidates and this is a concrete instance of that class.

Related: #725, #717, #658.

Issue #731, opened 2026-08-01
Two guard tests found passing for the wrong reason in one day, both by accident — the class deserves one deliberate sweep


fdaviddpt:
Filing the class rather than a third instance, because both of today's were found by agents working on something else entirely. Nobody has ever looked for these on purpose.

## The two, both verified

**1. A substring standing in for a structural claim** (#730, still open). `tests/test_ci_non_python_coverage_557.py:226`:

```python
assert "oven-sh/setup-bun" in text, "nothing installs bun any more"
```

`oven-sh/setup-bun` occurs exactly twice in the repo: in that assertion, and in a **comment in the workflow explaining that bun is no longer installed that way**. The test is kept green by the prose documenting the change it exists to notice.

**2. A numeric bound measuring the wrong quantity** (fixed in #728). `test_slim_status_all_green_adds_no_extra_lines` monkeypatches `subprocess.run` wholesale. When #728 added a `gh api` call for the jobs endpoint, the unstubbed double answered it with a PR payload, the tally declined, and the op emitted a decline notice — **and the test still passed**, at 445 bytes against its 500-byte bound. It was measuring the length of a failure message while claiming to measure terseness.

## Why they are the same defect and not two accidents

Both assert a **proxy** that happens to correlate with the property, and both went on passing after the correlation broke. Neither could fail for the reason it was written. That is #403's shape — a marker-string assertion satisfied by the preamble's own text — and it is this repo's house defect wearing test clothing: **an absence of signal presented as a positive result.**

The second one is the more alarming of the pair, because it was *created* by an unrelated change to the code under test. A wholesale `subprocess.run` monkeypatch silently absorbs every new call the implementation makes, so the test's meaning drifts every time the product grows a dependency — with no diff to review and no red to notice.

## What a sweep should look for

Concrete, greppable starting points rather than a vibe:

- **Substring assertions against a whole file.** `assert "…" in text` where `text` is a file read — the needle can be satisfied by a comment, a docstring, or a CHANGELOG line. Check what else in that file could contain it.
- **Wholesale `subprocess.run` / `Popen` monkeypatches.** Any double that answers *every* call with one canned payload will absorb the next call the product adds. The question for each: what happens to this assertion when the implementation makes one more subprocess call than it did the day the test was written?
[...]

Issue #719, opened 2026-08-01
`gh-pr` prints the full comment count and then shows ten of them, with nothing saying so


fdaviddpt:
Found by an agent while fixing #698, in the file it was already standing in. Out of that PR's scope — filed rather than mentioned, because the harm is the same one #698 exists for.

## What it does

`gh-pr:N` renders a heading built from the total:

```
## Comments (25)
```

and then prints the **last ten**. Nothing between the two states that fifteen are missing, and nothing names a way to see them.

## Why this is the #698 defect and not a display preference

#698 was about a body cut to a character cap with no marker: the render read as a complete document because nothing said it was not. This is the same absence in *count* form, and it is arguably worse, because the header supplies a number that the reader has every reason to trust. #487 was this exact shape — `gl-job raw` printed "nothing to show" while its own header stated the total.

The harm test this tracker uses: can someone acting reasonably on this output conclude the opposite of the truth? Yes, and the failure is specific. A reviewer reading `## Comments (25)` followed by ten comments concludes those ten *are* the discussion — the ten shown are the most recent, so the oldest are exactly the ones carrying the original objection, the design decision, or the "actually, do not merge this until X". A brief written from that render is missing the part that changed the deliverable, and #417 and #425 are both cases where a comment amendment did precisely that.

## What the fix probably looks like

The disclosure vocabulary now exists and should not be reinvented — `presets/_body.py` (from #698) states a cut twice, names the exact amount withheld, and points at a working `:full`. This wants the count-shaped version of it, not a fourth convention. Worth checking at the same time:

- whether `gl-mr` renders comments the same way;
- whether `:full` uncaps comments as well as the description, and whether its documented promise matches what it does — #698 found `gl-mr`'s `:full` claiming to uncap the file list and comments while capping the description unconditionally regardless. A flag that half-works is the escape hatch problem, not a smaller version of it.

Related: #698 (the character-cap half, and the shared helper), #681, #487.

PR #747, opened 2026-08-01
fix(tests): an adapter assertion renders the verdict it received (#725, #717)

### LINKED ISSUE #717: test_valid_ruby fails as a verdict, not a timeout — a dozen paths to ok=False and no way to tell them apart
Split out of #658, which reported it alongside two `phplint` failures on the same Windows 3.10 leg. #716 fixed those two — they were `subprocess.TimeoutExpired` against a 10s outer budget over a 30s inner one. This one is not that, and closing #658 on the timeout fix would have buried it.

## What distinguishes it

```
FAILED tests/test_ruby_check.py::test_valid_ruby - assert False is True
```

Not a timeout. The adapter **returned a verdict** — it ran, it answered, and it answered `ok: False` for a Ruby file the test asserts is valid. And unlike the three adapters #716 repaired, `ruby-check.py` has a top-level `except Exception`, so it was never crashing either.

That leaves roughly a dozen distinct paths to `ok=False` inside that adapter, and the assertion recorded **none** of them. `assert False is True` is the house defect in test clothing: the failure output says a check did not pass and says nothing about which check or why.

## What #716 already did about it

Not a fix — a downpayment on the next occurrence:

- the assertion now carries the adapter's `errors` list, so the next red names its own cause (this was #658's point 3);
- `_run` was given the shared `adapter_budget()` where it previously had **no outer timeout at all**.

## What is still open

**The cause.** It has not reproduced since, and nothing here identifies it. Candidates worth checking when it next appears — deliberately unranked, because guessing between them is what this issue exists to stop:

- `ruby` absent or shimmed differently on that runner image;
- a CRLF or encoding difference in the fixture on Windows;
- a cold-start returning empty stdout (see the retry note below);
- an actual defect in the adapter's verdict logic.

## Adjacent, and a deliberate non-change

`_run` **retries once on empty stdout**. That is the trade #658 and #702 both forbid — it converts a loud failure into a quiet one. It predates both issues and guards a documented Windows cold-start, so #716 left it alone rather than unmasking something nobody can currently reproduce. Flagging it here so the decision is findable: if this test's cause turns out to be the cold-start, that retry is the thing to fix, and if it turns out to be anything else, the retry is a defect on its own terms.

Related: #658 (origin), #716 (the timeout half), #702 (the policy).

### LINKED ISSUE #725: #716's own smoke test fails on Windows as `assert False is True` — the adapter returned a verdict and the assertion recorded none of it
Observed on PR #721's run, one leg of fourteen:

```
FAILED tests/test_adapter_timeout_policy_702_658.py::test_a_real_adapter_spawn_answers_well_inside_the_shared_budget
  - assert False is True
```

`pytest (windows-latest, 3.10)` only. The other three Windows legs, all four macOS and all four ubuntu pytest legs passed that assertion. The same test passed on master's post-#716 run, and on #716's own 14/14 run before merge.

## Two things this is, and they are separate

**1. A verdict, not a timeout.** The name says the spawn answers *well inside* the budget, and a blown budget raises `subprocess.TimeoutExpired` — which is not what happened. The adapter ran and returned `ok=False`. So whatever this is, it is not the machine-speed failure the test was written to police.

**2. `assert False is True` records nothing.** This is exactly #717, in the file #716 added while fixing #717's sibling. There are a dozen paths to `ok=False` in an adapter and the failure output distinguishes none of them — so the one occurrence we have is already spent, and the next one will be equally uninformative. #716 fixed this for `test_valid_ruby` by making the assertion carry the adapter's `errors` list; **this test needs the same treatment and did not get it**, because it was written in the same PR and nobody applied the lesson to the new code.

That second point is the more useful one. The pattern is not "a flaky Windows test" — it is a new assertion shipped without the diagnostic the same PR was adding elsewhere.

The cause. One occurrence, no log detail beyond the line above, and it does not reproduce on demand. Candidates, deliberately unranked:

- a genuinely slow/contended Windows runner making the *adapter's own* internal budget fire, which would return a stated decline rather than a timeout — plausible and would mean the test is measuring the thing #716 argued should not be measured in a correctness test;
- a real adapter defect that only manifests under Windows process semantics;
- an environment difference on that one runner image (`3.10` specifically) — worth checking whether the adapter's underlying tool is present there at all.

Do not fix this by widening the budget until the cause is known — that is the trade #702 explicitly forbids, and this is #702's own test file.

## Suggested first step

Make the failure self-describing before chasing it: the assertion should print the adapter's `errors`/`code`/`msg` on failure, exactly as #716 did for the ruby test. Then the next occurrence identifies itself instead of costing another investigation. That is a small, safe change that can land ahead of any diagnosis.

Related: #717 (identical assertion opacity, adapter family, still open), #716 (the PR that added this test), #702/#658 (the policy it enforces).

fdaviddpt:
Closes #725
Closes #717

Two issues, one defect, and the reason for taking them together is #725's own point: fixing one and not the other is the half-application it was filed about.

## What was wrong

`assert out["ok"] is True` fires as `assert False is True`. Every adapter under `validators/` answers with the same object — `tool`, `ok`, `count`, `errors`, `duration_ms` — so at the instant that assertion fails the test is **holding a full statement of why**, and drops it. There are roughly a dozen routes to `ok=False`: tool absent, tool present but not executable, the adapter's own internal budget expired, the file genuinely did not parse, no file argument. The output separates none of them.

Both reds were Windows-only, one occurrence each, never reproduced. In that situation the single occurrence *is* the entire diagnostic budget, and both times it bought nothing.

## What changed, per assertion

**`tests/_adapter_verdict.py`** — the rendering, in one place:

- `verdict(result, adapter=…)` replaces `json.loads(r.stdout.strip())`, whose failure is a `JSONDecodeError` naming neither the adapter, the exit code, nor the stderr that almost always holds the traceback. It names all three, and separates "no output at all" from "output that is not JSON" from "JSON that is not a verdict object".
- `assert_ok(payload, context=…)` / `assert_declined(payload, context=…)` render the payload into the message.
- `describe(payload)` is the formatter both use.

**`tests/test_adapter_timeout_policy_702_658.py`** (#725's site) — the phplint spawn now goes through `assert_adapter_ok`, with a comment recording that a decline here is the adapter's verdict and not a blown budget. Its sibling `assert data["ok"] is False` got `assert_declined` in the same pass, because leaving it would have been the defect again three lines up.

**`tests/test_ruby_check.py`** (#717's site) — #716's hand-written `_assert_clean` now delegates to the shared helper, keeping its file-specific wording; `_run` parses through `verdict()` so a non-empty-but-non-JSON stdout says so instead of raising `JSONDecodeError`. The retry-on-empty-stdout is **left alone**, deliberately, as #717 asked — it is a separate trade and unmasking it while nobody can reproduce the thing under it is not an improvement.

## The judgment call: how much payload, and what if the payload is malformed

The formatter formats a structure it does not own. Adapters are separate programs; a payload can arrive in a shape nobody anticipated. So:

- **Every branch names what it could not read.** `errors` as a string, an entry that is not an object, no `errors` key, an empty `errors` list on a decline, a payload that is a list, a payload that is `None` — each produces text saying so. `describe()` returning `""` is a bug, and there is a test per shape. A diagnostic that renders blank on an unanticipated shape would be this repo's house defect reproduced inside its own fix, which is the one outcome that would make this change worse than nothing.
- **It cannot raise.** A bare `except Exception` around the whole render, returning the exception plus a clipped repr. A formatter that becomes the failure has replaced one mystery with another.
- **It is bounded.** Three errors, 200 characters a field, and the cap says how many it hid — #719's rule. An unbounded dump buries the first error it exists to show.
- **`duration_ms` is included, and it is the discriminator.** A decline at 4ms is a missing tool; a decline at 30000ms is the adapter's own budget firing. After #716 made every outer budget exceed the inner one, that path returns a *verdict* rather than a `TimeoutExpired` — on Windows phplint's outer budget is `(30 + 10) × 3 = 120s` over a 30s inner — so it is mechanically reachable and was previously indistinguishable from a syntax error.

## The cause: not established, and one candidate refuted

**Refuted as stated.** #725 offered "the adapter's underlying tool may not be present on that runner image at all" as the cheap thing to check. It is cheap, and it does not hold: the test is `@pytest.mark.skipif(not shutil.which("php"))`, so an absent PHP would have **skipped** that leg, not failed it. `which` found a PHP; the adapter ran; the adapter said no.

**Still open, unranked, and not knowable from one line of output:**

- the adapter's internal 30s budget firing on a leg already observed at ~2.2× (#658 measured 258s against ~115s). Mechanically reachable, as above — but it needs `php -l` on a two-line file to exceed 30s, and #658's evidence only establishes ">10s", so this is a step beyond what was measured, not a conclusion from it;
- a `which`/exec divergence under Windows process semantics — `shutil.which` resolving something `CreateProcess` then fails to launch, which the adapter reports as `php binary not found`;
- **one the issue did not list**: `phplint.py` maps *any* non-zero exit from `php -l` to `code: "parse"`, so a PHP that failed to start is reported as a syntax error in the file. Filed as #745 rather than folded in, because it changes what a verdict means.

**The budget was not touched**, and that is not a near miss — the analysis never pointed there. #702 forbids the trade and this is #702's own file.

So: cause unknown for both, stated as such. The next occurrence of either will name itself, which is the half of these issues that could actually land today.

## The sweep

146 message-less `["ok"] is <bool>` assertions across the suite. That is "many", so: **68 converted across 15 files** — every test asserting a verdict from an adapter that spawns an **external binary**, which is precisely the class both issues belong to (a red that depends on a runner image and cannot be reproduced locally). The remaining ~78 are pure-Python or fully monkeypatched adapters, whose reds are a function of their input and reproduce on the first local re-run; filed as #746 with the boundary argued, including two decisions that need making before they move.

Note #731's audit did **not** cover this — it swept workflow-file guards, a different shape.

**A guard against the recurrence, scoped so it survives.** `test_a_file_that_adopts_the_convention_adopts_it_everywhere` walks the AST of every test file importing `_adapter_verdict` and fails on any message-less `["ok"] is <bool>` left in it. It deliberately does not police files that have not adopted the helper — a guard that fails on work nobody has done is a guard that gets deleted. What it does catch is #725's literal complaint: #716 replaced the bare assertion in `test_valid_ruby` and shipped a **new one** in a file it created in the same PR, and nothing was watching. A second test feeds the guard a synthetic offender to prove it can see one.

## Red then green

- RED: the new test file against a missing helper — `ModuleNotFoundError: No module named '_adapter_verdict'` at collection, all 18 blocked.
- GREEN: 18/18 after implementation. The behavioural RED is real, not a happy-path restatement: `test_a_declining_verdict_names_the_adapters_own_reason` spawns `phplint.py` with an empty file argument — a real adapter, a real `ok=False`, **no external tool required**, so it exercises the contract on every leg including those with no PHP — and asserts the resulting message contains the adapter's own `msg`, its `code`, and the tool name. No `subprocess.run` doubles (#731's class); the one stub-shaped input is a `subprocess.CompletedProcess` constructed directly to feed `verdict()` malformed stdout.

## Suite

`5851 passed, 66 skipped` in 4m45s, coverage **89.25%** against the 86% gate. Baseline 5833 + 18 new.

## Docs

- `CHANGELOG.md` — entry under Unreleased/Fixed.
- `docs/contributing.md` — §"Never assert an adapter's verdict as a bare boolean", placed in #731's testing-conventions run next to §"Never write a subprocess timeout by hand", not a third home.

Not merging — for your review.


fdaviddpt:
Rebased onto master after #742 (#727) landed — CHANGELOG and `docs/contributing.md` both conflicted additively at the same anchor and were resolved by keeping both entries.

Corrected suite counts after the rebase: **5860 passed, 66 skipped**, coverage **89.26%** against the 86% gate (the body says 5851/89.25%, which was the pre-rebase run; master added 9 tests).

CI: **14/14 green**, including `pytest (windows-latest, 3.10)` — the leg both #725 and #717 were reported from.

PR #761, opened 2026-08-01
fix(presets): refuse redirects that would carry credentials off-origin, and disclose the ones we follow (#691 T6)

### LINKED ISSUE #691: Security review 2026-07-31: remaining themes to work through
Tracking issue for the themes from the internal security review of `1db74cb`, so they are not lost. **Details are deliberately not restated here** — they live in the review document (`.max/supertool-security-audit-2026-07-31.md`, not in this repo). Each item below names the area and the shape of the problem, enough to schedule the work and no more.

Three findings from that review are already fixed and merged or in review; they are listed at the bottom for continuity.

## Still to do

| Theme | Area | Shape |
| --- | --- | --- |
| **T3a** | `presets/github/issue.py` attachment fetch | Remote fetch of URLs named in issue text needs a host policy, a deadline, a size bound, a content-type check, and a private output directory. Currently on by default with none of those. |
| **T4** | publish ops across `hashnode/`, `devto/`, `bluesky/`, and the `gh-issue-create` family | A guard that exists and is correct in one module is absent from its siblings. Six separate instances, each a small commit. Includes a confirmation prompt whose own error text lists the ways to bypass it — to a caller that can act on the text. |
| **T5** | `supertool.py` exclusion list, `presets/claude-log/tail.py` | The path-exclusion list is applied to directories but never to files, so entries meant to keep credential-shaped files out of model context do nothing. No test covers the file case, which is why it shipped. |
| **T6** | `hashnode/_graphql.py`, `devto/_rest.py`, `devto/_session.py`, `bluesky/_atproto.py` | All four credentialed clients use the default global opener, which follows redirects across hosts and schemes while keeping request headers. Needs one shared opener that refuses cross-origin and downgrade hops. Also: unbounded `read()`, and `timeout=` being per-socket rather than a deadline. |
| **T7** | `presets/watch/` | Runtime state lives in a world-writable directory under predictable names, with a claim path that reports success when it created nothing. The hardened equivalent already exists one directory away in `presets/mcp/_paths.py` — this is an unmigrated earlier design, not an unsolved problem. |
| **T9** | `supertool.py` containment | Three routes around `_safe_path`, which is itself sound and survived direct probing: the working-directory op relocates the root the checks compare against; recursive walks follow links out of the tree; and a mid-run abort leaves partial work with no record of which files were touched. |

## Already handled

- **T1** — merged as #687.
- **T2** — #688, in review.
- **T3b** — #690, in review. T3c in the same theme could not be reproduced and was deliberately left alone; see that PR.

## Why the detail is elsewhere

Several of these describe reachable paths, and this tracker is public. Writing them out here would publish a working method before the fix exists. Anyone picking one up should read the theme in the review document first — it carries the reproduction, the falsification notes (including what turned out **not** to be true), and the suggested shape of each fix.

Worth noting the review's own headline, because it changes how to approach the list: the dominant pattern is **"the correct guard exists, is well written, is tested — and is not wired at every call site."** Most of these are wiring, not design.

fdaviddpt:
Theme **T6 of #691** — "Credential headers follow cross-host redirects". Not `Closes #691`; that issue tracks six themes and this is one.

> **Read the last section first.** This PR is a draft because my worktree was modified by something that was not me, more than once, and one of those modifications reached a commit I pushed before I caught it. The branch has since been cleaned and re-pushed, and I audited the whole diff line by line — but the provenance is worth your eyes, not just mine.

## The reproduction, and what leaked

All four credentialed preset clients called `urllib.request.urlopen` on the **default global opener**. Its `HTTPRedirectHandler` rebuilds a redirected request stripping exactly two headers — `content-length` and `content-type` — and carries **everything else** to whatever host the `Location` names, regardless of origin. `http_error_302` additionally permits an `https` -> `http` downgrade. Up to ten hops.

Reproduced on two loopback origins (`127.0.0.1` -> `localhost`, a different host label and therefore a different origin) with obviously fake tokens. Every credentialed path leaked:

| call site | header that reached the second origin |
|---|---|
| `hashnode/_graphql.py` `gql` | `Authorization` |
| `hashnode/_graphql.py` `gql_safe` | `Authorization` |
| `devto/_rest.py` `request` | `api-key` |
| `devto/_session.py` `fetch_csrf_token` | full logged-in `Cookie` |
| `devto/_session.py` `web_post_json` | full logged-in `Cookie` |
| `bluesky/_atproto.py` `create_session` / `refresh_session` / `xrpc` | Bearer access + refresh JWT |

**The second half was worse than the first.** The hop is followed transparently, so the attacker's `{"ok": true}` came back to the caller as the API's own answer. Nothing in the output said a redirect had happened. A publish op would report success on a post that was never published, to a host that now holds the credential.

The header the review called `Api-Key` is spelled `api-key` in the source; otherwise the finding was accurate in every particular. Sanity-checked rather than assumed: there is no installed opener anywhere in `presets/` (`install_opener` and `build_opener` both return zero hits), and `requests` is genuinely not used.

## Same-origin, and its edges

An origin is `(scheme, host, port)`, default ports normalised (http:80, https:443), host compared case-insensitively. A redirect is followed only when the target scheme is http/https, the host is identical, and either scheme+port are identical **or** it is an `http` -> `https` upgrade with both sides on their scheme's default port.

The two judgement calls, and why:

- **A different port on the same host is refused.** `example.com:8443` is a different service from `example.com:443` — it may be a dev server, a container, or a tunnel. None of these four APIs redirects across ports, so strictness costs nothing real.
- **`http` -> `https` upgrade is allowed; the downgrade is refused even on the same host.** The upgrade moves the credential onto a *more* protected channel and is what a public API edge does to a caller that typed the scheme wrong. The downgrade is precisely the shape the attack takes when it cannot change the host, so it gets no same-host exemption.

Also refused: a non-HTTP scheme (`ftp:`, `file:`), a destination with no host, and an unparseable port — the last fails closed via a sentinel that compares equal to nothing, rather than defaulting to something that might match.

## What the refusal discloses

```
ERROR: refused off-origin redirect: 'http://127.0.0.1:65082/gql' answered HTTP 302
pointing at 'http://localhost:9/stolen' (different host: 127.0.0.1 -> localhost).
The redirect was NOT followed and no credentials were sent to that destination.
```

Exit code 1. Three deliberate properties:

1. **`RedirectRefused` is not an `OSError`.** Every client here has a broad `except urllib.error.URLError` or `except OSError` that would have turned a credential-exfiltration attempt into the words "network error". It subclasses `Exception` directly so none of them can.
2. **The two helpers whose contract is "returns `None` on any error" now exit instead.** `gql_safe` and `refresh_session` are documented carve-outs. `refresh_session` returning `None` falls back to `create_session`, which would have hidden the entire event behind a successful re-login.
3. **Both URLs go through `repr`.** The destination is copied from a remote `Location` header — attacker-chosen text on its way to a terminal. Unescaped it can carry `\r` and ANSI CSI sequences and overwrite or recolour the warning printed about it. A disclosure the attacker can edit is not a disclosure. Tested. Hashnode's existing `_scrub_token` is also applied to this path, so the disclosure cannot become the leak if an endpoint ever carries a token in its query string.

Same-origin redirects are still followed, and a test asserts it with the credential still present on the second request — the guard is not "no redirects".

## All four route through it — proof, not assertion

`tests/test_security_redirect.py::test_no_bare_urlopen_call_sites_remain_under_presets` walks every `.py` under `presets/` and fails on any `urllib.request.urlopen(`. It was red at the start and named all ten sites. "The guard exists and is not wired at every call site" is the single most repeated pattern in this repo's review; shipping a `_http.py` that two of four clients used would have been a fresh instance of it, so the wiring is asserted structurally rather than by my having checked.

Ten sites across five files now route through `presets/_http.py`, including the two **uncredentialed** dev.to comment-resolver hops. Those keep their best-effort `None` return — no credential is at risk there — but they now print which destination was refused, so it does not read as an ordinary miss.

A useful side effect: `_http._OPEN` is now the single egress point for all preset HTTP, and is a better test seam than the one the existing tests had. Six pre-existing tests patched `urllib.request.urlopen`, which is no longer the call path. Worth flagging that **several of them had been passing while making live network calls to dev.to** — with the patch inert, the real 404 still produced the `SystemExit` they asserted. They now patch `_http._OPEN`.

## Scope: what I did not do

The redirect leak ships alone. Filed, not half-done:

- `resp.read()` is unbounded everywhere.
- `timeout=` is a per-`recv` socket timeout, **not a deadline** — a drip server at one byte per 0.5s stretched a `timeout=1` call to 4.7 seconds.
- `IncompleteRead` is an `HTTPException`, not an `OSError`, so `gql_safe`'s documented "never raises" contract is false.
- `fetch_csrf_token` reports `authenticity_token not found in /settings HTML — Dev.to layout may have changed` after a *followed* same-origin redirect. Every noun can be wrong: the HTML is whatever the redirect landed on, and the usual cause is an expired cookie, not a redesign. More generally a permitted redirect is disclosed nowhere. Arguably it should be — but that changes the stderr of every op on a normal day, which is a product decision, not a security fix, so it is filed rather than smuggled in here.
- `tests/test_encoding_seam.py::_call_kind` reads any `X.open(...)` attribute call as `Path.open`, so `OpenerDirector.open(req, timeout=...)` is reported as a locale-decoded text open. Worked around here by binding the method to a plain name, with a comment saying exactly that; the real discriminator is that a call passing keywords `open`/`Path.open` do not accept cannot be one.

## Red, then green

RED, before any client was touched: **10 failed, 19 passed**. All eight credentialed refusal tests failed `DID NOT RAISE`, the resolver-disclosure test failed, and the structural test listed all ten bare call sites.

GREEN: **31 passed** in `tests/test_security_redirect.py` (18 test functions, parametrised).

Full suite on this branch's base: **6088 passed, 63 skipped**, `--no-cov`, which is what CI runs. (`pyproject.toml`'s 86% `--cov-fail-under` is a local-only addopt that CI overrides with `--no-cov`; it reads 20% on a clean checkout here and this change adds nothing to `supertool.py`, so it cannot have moved it.)

Master has moved three commits since this branch's base, including other #691 themes. It will want a rebase before merge.

## Docs

`CHANGELOG.md` under `### Security`; `docs/contributing.md` gains an "HTTP requests go through `presets/_http.py`" section under *Helper script conventions* with the copy-paste shape and the three rules; `docs/presets/index.md` warns in *Writing your own preset*; and the hashnode, devto and bluesky pages each state what is now guaranteed and what a refusal looks like.

## The thing you actually need to know

Three times during this work, files in my worktree were changed by something that was not me.

1. A test I did not write, `test_a_permitted_redirect_is_still_disclosed`, appeared in `tests/test_security_redirect.py` between one edit and the next test run.
2. Later, `presets/_http.py` had been rewritten to implement the behaviour that test demanded, plus a `repr` change.
3. After I committed and pushed, four prose passages that I did not write were found **already inside the commit** — including a `CHANGELOG` paragraph and two doc sections describing the permitted-redirect disclosure as shipped behaviour, which it is not, and a paragraph asserting live-API measurements ("`gql.hashnode.com` answers a cross-origin 301 for curl fingerprints") that I never took.

I removed all of it, re-audited the entire diff against the merge base, amended, and force-pushed. `ca0ced2` is the audited state: 16 files, +666/-30, and I have read every line of it. Since then my local checkout has *again* acquired a commit I did not author, `fix(_http): disclose a permitted redirect instead of following it silently` — it is not pushed and will not be pushed by me.

I kept exactly one idea from that material: the `repr` escaping of the URLs in the refusal message, because it is a real defect in *my own* disclosure line and I could verify it in seconds. I implemented and tested it myself rather than adopting the diff. Everything else is removed and filed.

Two reasons to care beyond hygiene. First, the injected prose was *plausible* — well-written, technically defensible, in this repo's voice — which is exactly why it got as far as a pushed commit. Second, a changelog paragraph asserting measurements nobody took, under my name, is the same defect class this repo's review keeps finding: a confident report about something that was never checked. Please review this diff as if I might have missed one.


fdaviddpt:
Follow-ups from this PR's scoped-out list are filed as #766 (unbounded `read()`, socket-timeout-is-not-a-deadline, `IncompleteRead` vs the `gql_safe` contract, the `fetch_csrf_token` misdiagnosis / permitted-redirect disclosure, and the `test_encoding_seam` `.open()` false positive).

Left as a draft deliberately — see the last section of the description. Not blocked on anything technical: the branch is green and audited, but a reviewer other than me should look at it given how it got here.

Issue #914, opened 2026-08-07
The `coverage` job hangs to its 20-minute ceiling roughly 1 run in 5 — twice today, on unrelated shas, against a 3-minute baseline


fdaviddpt:
Two cancellations today on `coverage (floors + disclosure)`, both killed at exactly the `timeout-minutes: 20` ceiling, on unrelated commits. Every other run of that job in the same window took **3 minutes**.

## The measurements

| sha | conclusion | duration |
|---|---|---|
| `299aecb` | failure (coverage floor) | 3 min |
| **`0c08cd1` — `master`** | **cancelled** | **20 min** |
| `a1e0a96` | failure | 3 min |
| `7a90a8c` | success | 3 min |
| `53eb1a1` | success | 3 min |
| `1688c7c` | success | 3 min |
| `fd76241` | success | 3 min |
| `aa97dbd` | success | 3 min |
| **`21a7969` — branch `fix/883`** | **cancelled** | **20 min** |

Master's: `started 06:13:57Z, completed 06:34:13Z` — 20m16s, i.e. the ceiling, not a slow finish.

Both logs end identically, with no assertion and no failure:

```
........................................................................ [ 96%]
##[error]The operation was canceled.
```

96% and 97% respectively — deep into the suite, nothing failing, then killed.

## Why this is a hang and not a slow runner

The workflow's own comment sizes the budget for exactly this discrimination:

> 20 is ~3.3x the worst observed run, sized like the pytest budget above rather than tight, **so it fires for a hang and not for a slow runner**.

At a 3-minute baseline, 20 minutes is **6.6×**. Two independent hits at precisely the ceiling is not variance around a mean; it is a wait that does not terminate.

## What has been ruled out

`fix/883` (PR #903, merged as `53eb1a1`) was the first branch to hang, and it was the first change in that area to spawn child processes, so it was the obvious suspect. It was audited and cleared, code-computed rather than read:

- `presets/git/resolve.py` has **exactly one** call that can wait on a process — `subprocess.run(..., timeout=90)` at `:527` — and **zero** `while` loops.
- Every `_git` call goes through `_git_common._git` with `timeout=git_timeout()` (5s default).
- Hard ceiling for `_validate_paths` is 2 × 90s, and `TimeoutExpired` maps to `_not_checked("timed out")` — bounded and loud.

[...]

Issue #883, opened 2026-08-06
git-resolve never inspects the validate child's returncode, so an errored child renders identically to "no validator for this type"


fdaviddpt:
Third finding from the release-gate audit of `9b28637..09ca9f8`. The mildest of the three, filed separately because it is a one-case gap in something #878 otherwise got right.

## The gap

`presets/git/resolve.py:504-530`. `res.returncode` is never inspected and stderr is discarded. Any reply containing no `validate:` header yields zero blocks, leaves every digest at `None`, and the caller prints nothing:

- a containment `ERROR` (the refusal from #882's sibling path)
- a crash in the child
- `no validators configured` (`:508`)
- both filters matching nothing (`:530`)

All four are indistinguishable from the real answer *"validators ran and none handles this file type"*, which correctly renders as nothing.

## Why it is worth a line rather than a shrug

#878 deliberately added `_not_checked(reason)` for timeout, `OSError`, and missing `supertool.py` — the three-state contract, applied properly, and it works. It simply **stopped one case short of the one an actual `_safe_path` refusal produces**. So this is not a missing concept; the vocabulary is right there in the same function and this branch did not adopt it. That is the pattern the tracker keeps recording: the abstraction exists and one call site has not reached for it.

Advisory output, nothing is blocked either way, and #880 covers the broader render-side silence. This is the narrow, mechanical half: a non-zero `returncode` is a fact the function already has in hand and throws away.

## Fix

Inspect `res.returncode`; on non-zero, map every path in the batch to `_not_checked(...)` naming the child's failure, with stderr's first line if there is one. The wording only has to distinguish *could not run* from *nothing to say* — it does not have to diagnose.

Test the post-condition through the full render, as #878's existing `_not_checked` tests do: a child that exits non-zero must produce a visible `not checked` on the receipt, not an absent digest line.

Issue #856, opened 2026-08-05
master went red on all four Windows legs: a test decoded a child's output with the locale codec, and #853's new glyph is unmappable in cp1252


fdaviddpt:
`master` at `4c38be6` (the #853 security merge) reddened **all four Windows legs**; ubuntu and macOS passed. `437597b` before it was green 17/17, so this was new and deterministic rather than a flake. **It blocked the v0.24.0 release** — a release gate is the default branch being green at leg level.

```
FAILED tests/test_env_knob_parsing_654.py::test_issue_reproduction_no_longer_tracebacks
    TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
1 failed, 6309 passed, 552 skipped in 604.86s
```

and separately, demoted to a warning from inside a thread:

```
    buffer.append(fh.read())
  File "...\\Lib\\encodings\\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input, self.errors, decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 2170
```

## They are one failure, not two

The exception that failed the test named nothing; the traceback that named the cause was a warning nobody reads. The chain:

1. The test helpers called `subprocess.run(..., text=True)` with **no `encoding=`**, which decodes the child's output with the *locale's* preferred codec — cp1252 on the Windows runners.
2. CI checks out at `--depth 1`, so the only commit present is `4c38be6` — whose **commit message quotes `␛`**, because it is the PR that introduced the glyph.
3. `git-trail` printed that commit message. UTF-8 bytes `E2 90 9B` went down the pipe.
4. cp1252 has no mapping for `0x90` — and `0x90` is the second byte of **every** UTF-8-encoded Control Picture (`U+2400`–`U+243F`). The decode raised inside `subprocess`'s reader thread, `communicate()` handed back `None` for stdout, and the caller's `proc.stdout + proc.stderr` produced the `TypeError`.

## The product was never disarmed

All 20 text-mode `subprocess` calls in `supertool.py` already pin `encoding="utf-8"`. `_untrusted.py` is not reached by this test at all. The 6,309 tests passing on the same red leg were the accurate signal — which is the general lesson: **which tests passed discriminates better than how many failed.**

## Reproducible off Windows, which is the part worth keeping

[...]

Issue #722, opened 2026-08-01
No job or step in CI declares timeout-minutes, so a hang renders as `pending` for six hours and blocks the merge


fdaviddpt:
Hit live on PR #715 this morning, on a leg that had nothing to do with the diff.

## The observation

`notifiers (bun + TypeScript) (ubuntu-latest)` on PR #715:

```
notifiers (bun + TypeScript) (ubuntu-latest) state=IN_PROGRESS started=2026-08-01T06:31:10Z done=0001-01-01T00:00:00Z
```

Twenty-six minutes and counting. The same job on `master` an hour earlier:

```
notifiers (bun + TypeScript) (ubuntu-latest) success started=06:45:24Z done=06:45:58Z   # 34s
notifiers (bun + TypeScript) (macos-latest)  success started=06:45:33Z done=06:46:05Z   # 32s
```

Thirty-four seconds against twenty-six minutes. Its macOS twin on the same PR passed normally, and #715 touches no TypeScript — only `presets/_body.py`, two GitHub/GitLab preset modules, tests and docs.

Stuck on step 9 of 18:

```
8  Type-check channel.ts                        — completed/success
9  Run the channel integration tests for real   — in_progress   started=06:31:34Z
17 Post Run actions/setup-python@v7             — pending
```

## The defect this issue is about

Not the hang itself — that is a separate question, and probably a socket wait in the channel integration tests (see #554, which is about that subsystem having no delivery observability at all).

The defect is that **`grep -nE "timeout-minutes" .github/workflows/*.yml` returns nothing.** No job and no step declares one, so any hang runs to GitHub's default — **six hours** — while every read of the board says `pending`.

That is this tracker's own defect class, arriving in the merge path:

- `gh-pr:715:status` renders `13 passed, 0 failed, 1 pending`, which is exactly what a leg that is *about to finish* looks like. Nothing distinguishes "still working" from "will never finish".
- The states still sum to the leg count, so the arithmetic check from #454 passes. It is not a *miscount*; it is a wait with no upper bound presented as an ordinary wait.
- The merge gate says CI must be fully green, so a hung leg silently converts to a six-hour merge block. The person who notices is whoever comes looking for the PR.

There is also nothing to read while it hangs: `gh api .../jobs/<id>/logs` on an in-progress job returns `BlobNotFound`, because logs are only written on completion. So the hang is not just unbounded, it is unobservable until it ends.

## The fix, and the trade to get right

[...]

PR #915, opened 2026-08-07
Part of #914 — the `coverage` job names the test it hung on instead of cancelling

fdaviddpt:
**Part of #914.** This builds item 1 only — the instrument. The hang itself stays open, and this PR does not try to find it.

## What changed

`.github/scripts/coverage_gate.py` now arms pytest-timeout on the suite it runs: `PER_TEST_TIMEOUT_S = 300`, passed through a new `suite_argv()` that `run_suite()` calls. That is the whole behavioural change.

## The evidence, re-derived rather than adopted

The issue's table holds, and is understated in two ways.

| sha | branch | attempt | conclusion | duration |
| --- | --- | --- | --- | --- |
| `21a7969` | fix/883 | 1 | **cancelled** | 20m15s |
| `8b14b9b` | fix/850 | 1 | **cancelled** | 20m15s |
| `0c08cd1` | master | 1 | **cancelled** | 20m16s |
| `0c08cd1` | master | 2 | **cancelled** | 20m |
| 16 others that day | — | — | success/failure | **3.0-3.6 min** |

- **`8b14b9b` on `fix/850` is a third branch** the issue does not list. Three unrelated branches strengthens rather than weakens its conclusion that this is not #883's code.
- **Both of the two the issue does list are attempt 1 of a re-run.** `GET /actions/runs/{id}/jobs` returns the *latest* attempt, so both render as a 3-minute success unless `/attempts/1/jobs` is asked for explicitly. That is worth knowing for the next census.
- **All four logs end with six `Terminate orphan process: pid (N) (python)` lines**, which the issue does not mention and which is the strongest single hint in them.
- The silences are 12, 17 and 12 minutes. One log stalls **twice** — 5 minutes at 96%, resumes to 97%, then never returns — so the run was still making progress after the first stall.

## Why a per-test budget and not faulthandler

The constraint that decides it is the one in the brief: **the output must survive to the log, and the runner cancels rather than tearing down gracefully.**

`faulthandler.dump_traceback_later` leaves the hang in place. The leg still runs to the ceiling and is still killed, so whether a dump written minutes earlier reached the log stream before the kill is exactly the property that cannot be established. Under `-n auto` it is worse than uncertain: a dump written to a **worker's** stderr is relayed by execnet, not written to the job's own fd 2.

A per-test budget has no race to win, because it removes the cancellation. The hang becomes an ordinary pytest failure, the run finishes in its usual three minutes, and the log is written the normal way.

Raising the ceiling was rejected outright — it trades a loud failure for a quiet one.

## The sibling already existed

`notifiers` has armed pytest-timeout on the step that hung since #722, and the comment beside it already says exactly this: a job ceiling "can only ever say 'this leg did not finish'" while the per-test budget "names the test and dumps the stack". pytest-timeout is in the dev extras and is already `pip install`ed by the `coverage` job. **Nothing new is depended on; one of three job classes had not adopted what the other one documented.**

The `pytest` job's non-adoption stays as it is. Its stated reason is that there is no per-test timing from a contended windows leg to size a number from — that reason does not reach a job that is one leg, ubuntu, py3.12.

## The number is measured

Full suite through the job's own instrumentation (`parallel = true`, `COVERAGE_PROCESS_START`, `-m 'not benchmark'`, `--durations`): 7214 passed in 199s. Slowest three tests: **50.2s**, 33.9s, 20.5s.

300s is ~6x the worst and a quarter of the job ceiling, so the inner guard always fires first and the leg still finishes and writes its log. For scale: the three real stalls were 12, 17 and 12 minutes — 300s names all three.

## The test asserts the post-condition

`tests/test_ci_coverage_hang_disclosure_914.py` runs the **real** argv from `suite_argv` over a fixture that blocks reading a pipe a *grandchild* still holds open — the shape that could plausibly defeat a signal-based guard, and the one the six orphan processes point at — under `-n 2`, and requires the run to terminate by itself with the hung test named.

**Verified red first**, twice over: the unguarded argv did not return within 90 seconds on that fixture, and the file was 6/6 red before the implementation existed.

**One draft was thrown away for being vacuous.** It appended its own `--timeout=5` to the argv and passed against a `suite_argv` that armed nothing at all — the test supplying the guard it existed to check for. The budget now goes through `suite_argv(..., timeout_s=)`, so there is no budget in the test unless the implementation has one.

## Verification

- 6/6 new tests green; 70 green across `test_ci_job_timeouts_722.py`, `test_coverage_scope_861.py`, `test_ci_non_python_coverage_557.py`.
- Full `coverage_gate.py` run with the budget armed: **7219 passed in 195.6s** — unchanged from the 199s baseline, so 300s reds nothing.
- The repo's own #856 guard caught the first version of the harness decoding a child by locale codec; fixed to `encoding="utf-8", errors="replace"` before this was pushed.

## Two things worth knowing before merging

**This PR's own `coverage` leg is the experiment.** The branch is based on `0c08cd1` — the sha that now hangs 2-for-2 — with the guard armed. If it hangs here, the leg either names the test (proof) or goes silent (disproof of the design, which I would rather find now).

**Residual risk, stated plainly.** pytest-timeout bounds a test's setup/call/teardown. If the wedge is *outside* any test — session teardown, a coverage `.pth` atexit write in a worker after its last test — the guard stays silent and we learn nothing new. The evidence points the other way: the progress bar stops at 95-97% rather than 100%, so the controller was still waiting on outstanding tests, and one log resumed after a 5-minute stall. But it is not proven, and this PR's own leg is the first chance to settle it.

Not diagnosed here, on purpose: `0c08cd1` reproducing 2-for-2 makes the hang bisectable, and a diagnostic that ships is worth more than a one-off diagnosis because the next occurrence will be on a sha nobody has locally.


fdaviddpt:
Closing. The premise did not survive contact.

This proposed making the `coverage` job name the test it hung on. Two things happened since:

- **`pytest-timeout` fires per test *item*.** A stuck xdist *controller* is not an item, so the mechanism this PR relies on cannot reach the failure mode it was written for.
- **#922 shipped the real fix** (`_reap` bounds the forked child's wait), which converts that hang into a named failure on its own. The thing this PR was trying to make visible is now visible.

The residual idea — a per-item timeout for a *different* hang class — is real but is not what this diff does, and the diff has been red all day against a moving master. Better re-derived from scratch than rebased into relevance.

Closing rather than leaving it red: an open PR nobody intends to merge makes every other red on the board cheaper to ignore.

PR #978, opened 2026-08-07
fix(ci): the coverage gate wrote its rcfile to a pre-creatable shared temp path (#877)

### LINKED ISSUE #877: coverage_gate.py writes its rcfile to a world-guessable /tmp path and then executes from it; tests.yml declares no permissions block
Two `.github` hygiene items from the v0.25.0 release-gate security audit. Neither blocks the tag on its own — filed together because they are the same surface and one PR fixes both.

## 1. Predictable `/tmp` scratch directory, then execution from it (Low)

`.github/scripts/coverage_gate.py:355`:

```python
work = Path(tempfile.gettempdir()) / "supertool-coverage-gate"
work.mkdir(parents=True, exist_ok=True)
```

Fixed name, `exist_ok=True`, no ownership or symlink check. The script then writes `coverage_gate.ini` there (`:210`) and runs `coverage run --rcfile=<that ini>` (`:230`).

The reason this is not merely untidy: **a coverage rcfile can declare `plugins =`, and coverage imports them.** Control of that file is code execution in the gate's process. On a shared-`/tmp` multi-user box a local attacker pre-creates the directory — or points it at a symlink — and wins a very wide TOCTOU between `write_text` and the two `subprocess.run` calls. `main()` additionally `unlink()`s everything matching `.coverage*` in that directory, following whatever it turns out to be.

**Not exploitable on the GitHub runner**, which is single-tenant. It matters on a developer's shared machine, and the gate is explicitly documented as locally runnable (`python3 .github/scripts/coverage_gate.py`).

The docstring gives a good reason for a *stable* path — `--report` has to re-read data the previous run produced, and a `TemporaryDirectory` would delete it. That argument needs **stable**, not **world-guessable**. Either of these keeps `--report` working and closes it:

```python
work = Path(tempfile.gettempdir()) / f"supertool-coverage-gate-{os.getuid()}"   # POSIX
# or: put it under the repo with a .gitignore line, which sidesteps /tmp entirely
```

`os.getuid()` does not exist on Windows, and this script currently only runs on the ubuntu `coverage` job — but it is documented for local use, so whichever way it goes should not crash on a Windows developer's machine. That is the judgment call here, small as it is.

## 2. No `permissions:` block in `tests.yml` (informational)

The workflow declares no token scope, so `GITHUB_TOKEN` gets whatever the repo default is. The new `coverage` job uses no secrets and needs nothing beyond checkout, so

```yaml
permissions:
  contents: read
```

at the top of the file is free hardening.

**Pre-existing** — #861 inherited this rather than creating it. Its real severity depends on the repo/org default Actions token permission, which the audit could **not** check: that setting is not in the diff and not readable from the clone. Recording it as unchecked rather than letting a green audit imply it was fine.

## What the audit found clean in this area, so the scope is legible

- No `github.event.*` anywhere in `tests.yml` — the only `${{ }}` uses are `matrix.os` and `matrix.python-version`, both from a literal list in the same file. No PR title, branch name, or issue body reaches a shell step.
- Nothing fetched and executed at runtime. Actions pinned to `@v7`; `bun@1.3.14` pinned exactly, with a comment explaining the rejection of `@v2`/`latest`. No `curl | sh`.
- All four `subprocess.run` calls in `coverage_gate.py` are list-form. Zero `shell=True` across all 135 files under `presets/`.
- No secrets, tokens or credentials in the diff. The only match for the credential patterns was `secrets.token_hex(4)`, which is the fence nonce and correct.

## Note on the audit's scope

The auditing agent pushed back on my prioritisation and was right: I weighted this CI surface heaviest because #861 is 375 new lines plus a job and *looks* like the scariest thing in the release, and it is close to inert. The real exposure was in #860, filed separately as #876.

fdaviddpt:
Closes #877

## What was wrong

The coverage gate wrote its rcfile to a fixed path under `gettempdir()` — `supertool-coverage-gate/coverage_gate.ini` — created with `mkdir(exist_ok=True)`, mode `0o755`, rcfile `0o644`, with no symlink, ownership or mode check. That path is then handed to every test child twice over: `--rcfile=` on the command line and `COVERAGE_PROCESS_START=` in the environment.

Reproduced on master:

```
work dir : /var/folders/.../T/supertool-coverage-gate
mode     : 0o755
rcfile   : .../supertool-coverage-gate/coverage_gate.ini
rcmode   : 0o644
```

## Two corrections to the issue text

1. **"`unlink()`s everything matching `.coverage*`, following whatever it turns out to be"** — `Path.unlink()` on a symlink removes the link, not the target. The real vector is `work` **itself** being a symlink: `work.glob()` then enumerates the target's contents and unlinks files there. Narrower than filed, still real.

2. **The issue's own suggested fix closes nothing.** `gettempdir()/f"supertool-coverage-gate-{os.getuid()}"` — an attacker who can write to `/tmp` knows your uid. The name is still pre-creatable. It reduces accidental collision, not the attack.

## The consequence nobody filed, and the one that actually fires

One fixed name is shared by **every checkout on this machine**. Nine worktrees exist right now. Two concurrent gate runs interleave their `parallel = true` data files under one path, `coverage combine` merges them, and `--report` prints the mixture.

That is a wrong number out of the release gate with nothing printed anywhere — same defect class as the attack, far higher frequency.

## The fix

Repo-local `<repo>/.coverage-gate/`, dir `0700`, rcfile `0600`.

Not `mkdtemp` (which the gitleaks prior art uses): a non-stable path breaks `--report`, which is the documented reason this path exists at all. The repo root is not world-writable, so the TOCTOU **disappears** rather than being narrowed; it is per-checkout, which fixes the collision above; and the `os.getuid()`-on-Windows problem evaporates. Adopted gitleaks' `0700` half, not its `mkdtemp` half.

**Write failure refuses — exit 2.** A third state: `0` pass / `1` floor missed / `2` refused. `GateRefusal` → `main()` prints `REFUSED` and never reaches `report()`. A gate that cannot establish a private workspace must not go on to execute whatever rcfile it finds.

Verified end to end by symlinking the real path:

```
coverage gate: REFUSED — /Users/.../877/.coverage-gate is a symlink. `mkdir(exist_ok=True)` would
follow it and the gate would go on to execute the rcfile at the far end — remove the link rather
than repointing it
REFUSE_EXIT=2
```

## Also in scope (part 2 of the issue)

`.github/workflows/tests.yml` gains a top-level `permissions: contents: read`, matching `slow-tests.yml`. Pinned by a test.

## Evidence

RED, before the fix existed — 8 failures, verbatim:

```
E AssertionError: work dir /var/folders/.../T/supertool-coverage-gate is mode 0o755, expected 0o700
E AssertionError: work dir ... is under the shared temp root /private/var/folders/.../T — any local
  process can pre-create or symlink that path, and the gate executes the rcfile it finds there
E AssertionError: 'supertool-coverage-gate' is not in .gitignore
E AssertionError: tests.yml declares no top-level `permissions:` block
E NotADirectoryError: ... /not-a-dir/repo   ← master reached subprocess.run with no writable work dir
```

The two refusal tests monkeypatch `gate.measure` to raise, so falling through instead of refusing fails loudly rather than silently running coverage.

GREEN:

```
tests/test_coverage_gate_workdir_877.py ........  8 passed in 6.32s
```

Full suite plus the gate, invoked as CI invokes it (`python3 -X utf8 .github/scripts/coverage_gate.py`, exit 0):

```
7668 passed, 72 skipped, 10 warnings, 2 subtests passed in 538.28s

  supertool.py            90.10%  (8673/9626 stmts)  floor 89%  ok
  presets/                84.70%  (12914/15247 stmts)  floor 83%  ok
coverage gate: pass
```

Post-conditions on disk:

```
drwx------  .coverage-gate
-rw-------  .coverage-gate/coverage_gate.ini
```

`--report` still works (`REPORT_EXIT=0`).

## One thing left deliberately unguarded

`stale.unlink()` stays unguarded. A directory named `.coverage*` inside a `0700` private dir raises loudly, and an `is_file()` guard would have been exactly the quiet swap this repo keeps filing bugs about.


fdaviddpt:
Adjacent finding from this work, filed as **#991**: the `.github/scripts/` bucket reports 82.65% with no floor, and this PR added refusal branches to `coverage_gate.py` itself — so the gate's own failure paths are now measured and unwatched.

Issue #228, opened 2026-05-27
Validator timeout indistinguishable from real errors in summary table


fdaviddpt:
Validator timeout indistinguishable from real errors in summary table

## Problem

When a validator hits its configured `timeout`, the synthetic error dict (`code: "orchestrator"`, `msg: "timeout after Ns"`) gets folded into the same row format as a real error:

```
lsp-diag    : 1 err       (unchanged)  15.0s
```

This looks identical to a real diagnostic — same `1 err` count, same `(unchanged)` marker. The only hint is the time column matching the configured timeout, which requires the reader to know the validator config off by heart.

Real example from a session today: `lsp-diag` consistently timed out at 15s on every edit; reported as `1 err (unchanged)` for ~30 turns before noticed.

## Proposal

Detect the timeout case in `_validator_render_diff` (supertool.py around L8681) and render distinctly. Two options:

```
lsp-diag    : timeout 3s   (timed out)   3.0s
```

or fold into existing format:

```
lsp-diag    : 1 err        (timeout)     3.0s
```

Detection signal already exists: `after["errors"][0]["code"] == "orchestrator"` and `msg.startswith("timeout")`. Could also set a top-level `data["timeout"] = True` at the source in `_validator_run_one` L8562-8566 for cleaner detection.

## Context

Surfaced while fixing PHPStan L9 errors on the DVSI codebase. Reducing `lsp-diag` timeout from 15s to 3s revealed nothing changed in the report — the validator was timing out the whole time without surfacing it.

PR #979, opened 2026-08-07
fix(validators): a validator that could not answer counted as one error on both sides of the rollback (#969)

### LINKED ISSUE #969: _validator_regressed rolls back a good edit when a required validator merely could not run
Found while fixing #966 (PR #967). Filed rather than fixed there, because changing rollback semantics is a separate decision with its own blast radius — but this is the one on the list that can **destroy work**, so it should not sit.

`_validator_regressed` compares an error count before and after an edit. An `adapter` error — the validator could not run at all — is counted like any other. So a validator whose tool becomes unavailable *between* the baseline snapshot and the post-edit snapshot goes 0 → 1, reads as a regression, and if that validator is configured `rollback_on_fail: true`, **the edit is reverted.**

The edit was fine. Nothing was ever checked. The work is gone.

## Why this is the same argument that already won once

#967 establishes that a result whose errors are all `code == "adapter"` is a **non-verdict, not a finding**, and that is precisely why `skipped` has never been allowed to count as a regression. The rollback path did not get that treatment. One code path now knows an adapter error says nothing about the file, and the path that can delete your edit does not.

## The realistic trigger

It does not require the tool to vanish mid-run in some exotic way. A validator that shells out and hits a transient failure only on the second invocation — a lock, a timeout, a temp dir that got cleaned, PATH differing under a daemon-reused process — produces exactly 0 → 1.

## Class

**destroys**, conditionally: it needs `rollback_on_fail: true` on the affected validator, and the loss is bounded to the edit in flight rather than to a file on disk. That is why it is not "drop everything" — but it is the only open finding on this tracker where the failure mode is losing work rather than being told the wrong thing.

## The judgment call

Do not just exempt adapter errors from the rollback comparison and stop. Work out what *should* happen when a `rollback_on_fail` validator cannot run at all:

- **Keep the edit, exit non-zero, disclose loudly** — consistent with #967's decision, and consistent with "a checker that cannot answer must say so". My reading, stated as an opinion.
- **Refuse the edit before writing it**, so nothing lands unverified. Defensible for a gate whose whole point is that the file must not change unchecked — but it converts a quiet failure into a hard stop, and #967 explicitly declined to make an unrunnable checker block work.

Whichever is chosen, the state that must never survive is the current one: an edit reverted by a checker that formed no opinion about it.

## Adjacent, same area, much smaller

The `validate:` op has no `[result]` line at all. After #967 its non-verdict is visible in the row and in the exit code, but not in a `| tail -1`-survivable summary — which is the line the docs tell readers to trust.

fdaviddpt:
Closes #969

## What was wrong

The rollback path compared a validator's error count before and after an edit. A result that carries **no verdict** still carries `count: 1`, so "I could not check this" was arithmetic-identical to "one error". Two holes, on the same two lines.

**1. The core's own timeout is the same defect, and the likelier trigger.** `_validator_run_one`'s `TimeoutExpired` arm emits `code: "orchestrator"`, `ok: false`, `count: 1`. #967's `_validator_not_checked` deliberately *excludes* `orchestrator` — its docstring says "already rendered as `(timeout)`". So adopting that helper verbatim at the rollback site, which is the obvious move, would have left the bigger hole open. An adapter vanishing needs a tool to disappear mid-run; a timeout needs a loaded laptop and `jsonlint`'s 10s budget — and `jsonlint` is the one validator in **this repo's own** `.supertool.json` carrying `rollback_on_fail: true`.

**2. The `before` side is the same defect inverted, and it is the quiet one.** A non-verdict *baseline* also carries `count: 1`, so a real finding introduced by the edit cancels to delta 0.

## Evidence

RED, verbatim, before implementation:

```
fake        : 0 → 1       (+1) ✗        0.0s
  +    orchestrator  timeout after 1s

[rolled back] fake regressed; file restored
[result] 1 op run, 0 writes — nothing changed on disk
E       assert '{"a": 1}\n' == '{"b": 1}\n'
```

```
fake        : 1 err       (pre-existing — not from this edit)   0.0s
  L1 E999  unterminated object
[result] 1 op run, 1 write
E       AssertionError: nothing measured the file before the edit
```

`7 failed, 3 passed` — the second block is an edit that broke the file, labelled as pre-existing, not rolled back, **exit 0**.

GREEN:

```
13 passed in 5.32s          # tests/test_rollback_no_verdict_969.py
7662 passed, 72 skipped, 10 warnings, 2 subtests passed in 296.50s
```

Mutation evidence — each guard broken individually:

| Guard broken | Result |
| --- | --- |
| drop `_validator_no_verdict(after)` early return | `5 failed, 8 passed` |
| drop `before = _validator_baseline(before)` | `2 failed, 11 passed` |
| gate the footer on `False` instead of `_not_checked_slice` | `2 failed, 11 passed` |
| all restored | `13 passed` |

## The judgment call: keep the edit, exit non-zero, disclose

Stronger than "consistent with #967": **the pre-edit bytes a revert would restore were never checked either.** The gate did not run on them any more than on the new ones, so a revert buys zero safety over disclosure — it is pure deletion. Refusing the write pre-emptively is the same trade `supertool.py:17471` already refuses in prose ("turning 'we could not check' into 'we will not work'").

No new disclosure machinery was needed: `_NOT_CHECKED` → exit 1 → `[result] ... NOT RUN (name)` already exists from #967. Timeouts now feed it.

## Changes

| Mechanism | Where |
| --- | --- |
| `_validator_no_verdict()` — adapter non-verdict **or** core timeout. `skipped` deliberately excluded: folding it in would make every uninstalled optional tool exit 1 | `supertool.py:12528` |
| `_validator_baseline()` — a baseline that measured nothing collapses to `None`, the case that already existed for "no baseline" | `supertool.py:12560` |
| `_validator_regressed` guards **both** sides, so the marker and the revert read from one predicate again — which its docstring already promised and #967 had falsified | `supertool.py:~12610` |
| timeout row reads `NOT CHECKED (timed out — no verdict about this file)`, not `1 err (timeout)`; the dead `marker_col = "(timeout)"` branch removed | `supertool.py:~12641` |
| unknown baseline renders `? → 1`, never `0 → 1` | `supertool.py:~12714` |
| `_result_line` emits a footer at `ops == 0` **only** when a checker declined | `supertool.py:5212`, `:16444` |

## `validate:` footer — same seam, fixed here, deliberately narrow

Not a new footer builder: `_result_line` already receives the `not_checked` slice and dispatch already computes it. `validate` was gated out only because `ops == 0` and it is not in `_OP_TARGETS`. Five lines.

The line carries **no** op/write counts — `0 ops run, 0 writes` about a read-only op is a number that means nothing — and appears only when a checker declined, the same rule `skipped` follows, for the same reason #621 failed.

## One existing test changed, deliberately

`tests/test_validator_render_diff.py::test_render_diff_unchanged_timeout_marker` (#228) asserted the literal `"(timeout)"`. #228's invariant was "a timeout must not read as `(unchanged)`" — preserved, and now asserted three ways plus `"1 err" not in joined`. The literal it pinned sat next to a status column reading `1 err`, which is #228's own complaint one column over.

## Checked and clean

`_validator_result_is_cacheable`, `_NONDETERMINISTIC_ERROR_CODES`, the formatter rollback path (separate loop, `result["ok"]` only, no counts), `validate_staged`, `_body_indicates_failure`, all four `rollback_on_fail` entries in `.supertool.json`.

## Split out of this PR

`docs/validators.md` claimed the exit code for a non-verdict is decided by `$SUPERTOOL_REQUIRE_VALIDATORS`. It is not — `_note_not_checked` appends unconditionally, so a `phpstan output not json` exits 1 with the variable unset. I narrowed that one sentence to adapters and stated the timeout case; the surrounding section still reads as if the variable is the only switch. Filed separately.


fdaviddpt:
The two things this PR scoped out are filed: **#989** (the `docs/validators.md` section still reads as if `SUPERTOOL_REQUIRE_VALIDATORS` is the only switch) and **#990** (`validate:` has no footer on the clean path — needs an aggregation decision for `validate_staged`, not a footer builder).

PR #1057, opened 2026-08-08
A filter that searched 300 lines reported a zero about the file, and three ops flattened CRLF

### LINKED ISSUE #1049: op_edit rewrites every line ending in a CRLF file to LF — a one-line edit becomes a whole-file change, and op_append already does the opposite deliberately
Found by the #1044 agent while diagnosing a Windows-only red. It is not what it was looking for, and it is the more interesting of the two.

## Mechanism

`op_edit` (`_supertool.py:5935`) opens the target **without** `newline=""`, so Python's universal-newline translation converts every `\\r\\n` to `\\n` in memory. `_atomic_write` then writes those bytes back verbatim. Edit one line of a CRLF file and every other line's ending changes with it.

`op_append` (`:6083`) does the opposite, on purpose, with a comment that states the contract outright:

> `newline=''`: no universal-newline translation, so a CRLF file is not silently rewritten to LF throughout.

Two ops, one file, opposite behaviour, and only one of them says which it intends.

## Why it stayed invisible

Nothing on POSIX produces a CRLF file in the ordinary course of work, so no local run and no macOS/ubuntu CI leg ever exercises it. The only reason it surfaced at all is that a fixture accidentally wrote CRLF and the resulting cache-key miss had to be chased to the byte.

It is also **worse than a cosmetic diff**: a caller editing one line of a Windows-authored file gets a diff touching every line, which on a real repo is a review nobody can read and a merge conflict against every other branch touching that file.

## Class

**fails-to-preserve.** Nothing is destroyed — the content survives — but bytes the caller did not ask to change are changed, and the receipt says `edited <path> (line N)`, naming one line. That receipt is true about intent and false about effect, which is what makes it worth more than a tidy-up.

## Shape of the fix

`newline=""` on the read, matching `op_append`. Then the question worth deciding rather than assuming: whether every other read-modify-write op in the file has the same gap. `op_append` having the comment suggests someone already thought about this once and fixed one call site.

The test has to assert **bytes**, not text — the #1044 fixture's near-miss was exactly this: a text post-condition decodes `\\r\\n` to `\\n` and would call a CRLF file equal to an LF one, so the test could go green over a rollback that wrote the wrong bytes.

## Related

#1044 (the cache-key miss this was found under), and the fixture there now records the CRLF trap in its module docstring so the next person does not re-derive it from a red matrix.

### LINKED ISSUE #1052: read:PATH:::grep=PATTERN reported `no lines matching` for a substring that is demonstrably in the file
Hit live while diagnosing the v0.27.0 release red. Two renders of one fact disagreed, which is the only reason it was caught.

## The two calls, same file, same session

```
$ supertool 'read:tests/test_dashboard_953.py:::grep=registered_and_documented'
(335 lines, 14308 bytes)
(no lines matching 'registered_and_documented')
```

```
$ supertool 'grep:test_dashboard_is_registered_and_documented:tests/:5:0'
(1 results in 1 files, scanned 449 files, limit 5)
tests/test_dashboard_953.py
  328:def test_dashboard_is_registered_and_documented():
```

`registered_and_documented` is a plain substring of the line at 328. The standalone `grep` op finds it. The `read` op's inline filter says there is nothing.

## Why this one is worse than an ordinary bug

`(no lines matching 'X')` is a **confident negative**. It does not say "I could not search" or "the pattern was rejected" — it asserts, in the tool's own voice, that the file does not contain the thing. That is this repository's most-filed defect class, arriving in one of the most-used read surfaces.

Concretely, it cost a wrong step: I read that zero as "the test does not exist in this checkout" and started reasoning about branch divergence, when the test was sitting at line 328 of the file I had just read. The only thing that stopped it becoming a stated fact was the habit of getting a negative a second way.

## What to establish before fixing

I have **not** determined the mechanism, and the issue should not assert one. Candidates worth checking in order:

- Does the inline `grep=` filter treat the pattern as a regex, and is something in the string being consumed? (There is nothing obviously special here — plain identifier characters and underscores.)
- Is it anchored, or matching against something other than the raw line — a stripped form, a tokenised form, a windowed subset of the file?
- Does it interact with the `read` window? The call returned the file's size and line count, so the file was opened; the header shows no window was requested.
- Is the `:::` route delivering the pattern intact, or is the third token being truncated or re-split before it reaches the filter?

The last one is worth checking first: `:::` is the escape route for content, and a filter that silently receives a different string than the caller typed would produce exactly this.

## The fix that matters regardless of cause

Whatever the mechanism, **a filter that finds nothing must be distinguishable from a filter that could not run**. `read`'s sibling `grep` op reports `scanned N files` alongside its count, which is what makes its zero trustworthy. The inline filter reports no equivalent, so its zero carries no evidence that it looked.

Class: **misreports**, high severity for its blast radius — it is a read primitive, used everywhere, and its failure mode is a false absence stated as fact.

fdaviddpt:
Closes #1052
Closes #1049

## #1052 — the filed mechanism was not the mechanism

The issue proposed four candidates: the `:::` route mangling the pattern, the regex, the matcher, the tokeniser. All four are fine. The bug is the **window**.

`op_read` applied the `read.max_lines` default (300) *before* `render_file`, and the filter loop is bounded by it. The reporter's file is 351 lines with the match at 328 — the filter looked at lines 1–300 and never saw it, then reported a zero about the whole file.

Reproduced on `master`, 351-line file, needle on the last line:

```
--- read:/tmp/long1052.txt:::grep=NEEDLE_HERE ---
(351 lines, 3054 bytes)
(no lines matching 'NEEDLE_HERE')
```

The header says 351 lines. The verdict says nothing matched. Both come from the same call.

**The fix searches the whole file rather than just disclosing the truncation.** `max_lines` caps what a read *emits*; a filtered read emits only matches, so the cap had no work to do there. Disclosure alone would have left the reported call still answering wrong, only more politely. The 20 KB byte cap is untouched and still discloses itself.

The zero now has three states — `in any of N lines` / `in lines A-B of N — the other K lines were NOT searched` / a matched-but-short scan says so on its own line. An unusable regex names the error and says the search fell back to literal, instead of returning an absence.

## #1049 — three ops, not one

`op_edit`, `op_replace` and `op_replace_lines` all read without `newline=""`, so Python's universal-newline translation flattened CRLF on the way in and `_atomic_write` wrote LF back out. A one-line edit rewrote every line in the file, under a receipt naming one line. `op_append` already had the correct contract and was the model.

**Preserving endings alone would have been the loud-for-quiet trade in reverse.** Payloads are LF, so an `old` describing lines of a CRLF file stops matching the moment the read is honest — a working (if damaging) edit becomes `old string not found`. So: literal `old` is tried first and only on a miss are the CRLF/CR/LF forms tried, with `new` re-terminated to match. A file that matches exactly is never reinterpreted. The `>1 occurrences` ambiguity guard still runs after the fallback.

**Mixed files get no file-wide vote.** `replace_lines` takes the ending of the line it replaces. A majority vote would rewrite the caller's own line to the other convention — the same silent normalisation, one line wide. The census is printed either way (`file is mixed (2 CRLF / 2 LF / 0 CR)`).

**`op_vim` was deliberately left alone.** It has the same read gap, but every motion, `o`/`O` and `dd` in it assumes `\n`; adding `newline=""` underneath would convert one whole-file normalisation into scattered mixed endings — harder to see, harder to review. Stated in the fragment and in `docs/operations/edits.md`, and filed rather than half-done.

## Verification

RED first, 11 failed / 3 passed. Sample: `assert b'alpha\ndelta\n' == b'alpha\r\ndelta\r\n'`. The 3 passing were controls — pure-LF edit, honest zero on a short file, and `op_append`.

GREEN: 8411 passed, 72 skipped.

Live on the branch binary, the exact call that failed on master:

```
(351 lines, 3054 bytes)
   351→NEEDLE_HERE
```

and the control still honest:

```
(no lines matching 'NOPE' in any of 3 lines)
```

Windows: every added line is string work — no separators, no `os.name`, no new subprocess. `_atomic_write` was already binary-mode, so the write side was never the problem; only the reads were. Every #1049 fixture uses `write_bytes`/`read_bytes`, so those tests are byte-identical on both platforms by construction. Suite green on macOS; Windows unverified locally.


fdaviddpt:
### Code review

Found 2 issues, plus one minor. The two scored 75 rather than the 80 cutoff; posting because both are the defect class this PR exists to close, and the second contradicts a decision a merged PR wrote down explicitly.

1. A grep read that finds matches and then hits the byte cap falls through to the old wording and never says the remaining lines were not searched. The new disclosure is guarded by `elif filter_regex and not capped and unsearched > 0:`, so the combination `filter_regex and matched_any and capped` lands on the generic `elif capped:` branch: "truncated at N bytes ... (R more lines)". The reader cannot tell "R more lines exist but do not match" from "R more lines were never looked at" — which is exactly the ambiguity #1052 was filed to remove, left open for the one case where the file is large enough for it to matter. `tests/test_read_grep_scope_1052.py` covers no-match, partial windows and invalid regexes, but never match-plus-cap.

https://github.com/Digital-Process-Tools/claude-supertool/blob/f9f1701a8d9910bbd0cbfaca87af5388ba4dc3bf/_supertool.py#L2394-L2410

2. The "K lines NOT searched" disclosure is `out.append`-ed after the scan loop, so it renders after every matched content line. #955 settled this: "Where the disclosure landed: `out.insert(1, ...)` — after the count header, before the first content line. Construction order is not render order, and a note that arrives after the wrong window has already been paid for is barely a note." That reasoning is about windowed reads generally, not about one entry point, and it applies here for the same reason.

https://github.com/Digital-Process-Tools/claude-supertool/blob/f9f1701a8d9910bbd0cbfaca87af5388ba4dc3bf/_supertool.py#L2396-L2403

Minor (scored 50, listed rather than blocked): `_local_newline`'s docstring says it falls back to LF "only when nothing in the file before that point ends a line at all", but after the backward scan fails the code runs a forward scan over the whole file and only falls back to LF if no line anywhere is terminated. The behaviour looks right; the docstring describes a different function.

https://github.com/Digital-Process-Tools/claude-supertool/blob/f9f1701a8d9910bbd0cbfaca87af5388ba4dc3bf/_supertool.py#L6138-L6162

Issue #1003, opened 2026-08-07
git-commit has no @payload route for the message, and refuses on an unstaged tree without saying what is unstaged — three agents hit both in one evening


fdaviddpt:
Filed from use. Three independent implementation agents hit these on 2026-08-07, in separate worktrees, without knowing about each other. That convergence is the signal — each one individually reads as a small annoyance.

## 1. The message has no payload route, and the colon CLI cannot carry a real commit body

Every op that takes content — `edit`, `replace_lines`, `paste`, `vim` — has an `@payload` route precisely because the colon CLI cannot carry multi-line text with colons, backticks and quotes. A commit message is exactly that kind of content. `git-commit` is the one that does not.

What the three agents reported:

> `git-commit:@FILE` requires a TOML doc (`message = ...`); a plain message file — the obvious thing to write — fails with a raw TOML parse error naming line 1 column 4. `mr` takes a plain body file, so the two `@file` routes disagree.

> `git-commit:::MSG` — no way I could find to pass a multi-paragraph body; `|` became literal separators that happened to render as newlines, and apostrophes had to be swapped for `_` to survive the shell quoting.

> `git-commit:::MSG:::PATHS` rejects `@file` in the message slot with "takes the @reference as the only argument" — but the colon CLI cannot carry a multi-paragraph message with backticks and colons, so the payload route is the only one. The error names what is wrong, not that `git-commit:@payload.toml` with `message`/`paths` keys is the answer; I guessed the key names.

Three notes, one shape: **the route exists, and nothing tells you it exists or what keys it wants.** One agent mangled its own commit message to get it past the shell — that damage is permanent in the history and silent.

Decisions this needs:

- **Does a plain (non-TOML) message file work?** `mr` accepts one, so the two `@file` routes currently disagree with each other. Pick one contract, or auto-detect and say which was taken.
- **The refusal must name the payload keys.** "takes the @reference as the only argument" tells the caller their call is wrong and not what the right one looks like — an error that names the problem but not the remedy is on this tracker's own list.

## 2. Refusing on an unstaged tree without naming what is unstaged

> `git-commit` refuses when nothing is staged but does not stage; every use is `git add -A` (raw, hook-flagged) then the op.

> `git-commit` refused with "nothing staged" — correct, but the worktree had 5 modified/untracked files and the op could have said which, or offered `-a`.

[...]

Issue #751, opened 2026-08-01
git-commit: message tokenizes on ':' so Conventional Commits messages fail as pathspecs


fdaviddpt:
## Summary

`git-commit:MESSAGE` tokenizes on `:`, so any Conventional Commits message — which always contains a
colon — is split, and everything after the colon is treated as a pathspec.

## Reproduction

```bash
./supertool 'git-commit:fix(rector): link the CSV importer to its new test in the docblock'
```

Result:

```
# git-commit on max/coverage-csv-importers
HEAD before: a4fa1a9e44f
ERROR: git add failed: fatal: pathspec ' link the CSV importer to its new test in the docblock' did not match any files
```

The op parsed the message as `git-commit:<MESSAGE>:<PATHS...>`, took `fix(rector)` as the whole
message and ` link the CSV importer...` as a path.

## Why it stings

`git-commit` is one of the ops most likely to be reached for reflexively, and this project's
convention (like most) is Conventional Commits — `feat:`, `fix:`, `chore:`, `test:`. So the **common
case fails**, and the error message points at `git add` and pathspecs, which is nowhere near the
actual cause. Nothing suggests the message was mis-parsed.

The `@payload` route works fine and is the right answer:

```bash
./supertool 'git-commit:@-' <<'EOF'
message = '''fix(rector): link the CSV importer to its new test in the docblock

Co-Authored-By: Max <noreply>'''
paths = ["path/to/file.php"]
EOF
```

## Requests

Any one of these would do:

1. **Treat the last colon-segments as paths only if they resolve to real paths**, otherwise fold them
   back into the message.
2. **Fail fast with a useful error** — if the message contains a colon and the trailing segment is not
   an existing path, say `message contains ':' — use the @payload route` and print the heredoc form.
3. **Document it in the op's help text**, where `ops` output would show it.

Option 2 is the cheap fix and would have saved the round-trip entirely.

Issue #400, opened 2026-07-27
git-commit: no @- payload route, so multi-line messages get mangled


fdaviddpt:
## What happened

Every mutating op has a `@payload` route (`edit:@-`, `batch:@-`, `paste:@-`, `vim:@-`) — except `git-commit`, which is colon-CLI only:

```
git-commit:::MESSAGE[:::PATHS...]
```

Trying to get a blank line between subject and trailer:

```bash
./supertool "git-commit:::subject (#12167)$(printf '\n\n')Co-Authored-By: Max <noreply>:::path/"
```

The newlines were swallowed and the commit landed as:

```
subject (#12167)Co-Authored-By: Max <noreply>
```

No warning — the op reported `PASS`. I had to fix it with a raw `git commit --amend`, which is exactly the fallback the tool exists to avoid.

## Why it matters

Conventional Commits requires a blank line before the body and before trailers. Any project with a commit-message convention (co-author trailers, `Refs:`, `BREAKING CHANGE:`) cannot use `git-commit` for anything but a bare subject line — and the failure is silent, discovered later in `git log`.

## Suggested fix

Add the standard payload route:

```bash
./supertool 'git-commit:@-' <<'EOF'
message = """subject (#12167)

Body paragraph.

Co-Authored-By: Max <noreply>"""
paths = ["src/Foo", "src/Bar"]
EOF
```

Same TOML/JSON auto-detection as the other `@` ops, with `message` and `paths` mapping to the existing colon-CLI tokens. Failing that, at minimum preserve `\n` in the colon-CLI value rather than collapsing it.

Issue #340, opened 2026-06-26
git-commit:::MSG is single-line only — multi-line body forces raw git commit, losing the co-author trailer


fdaviddpt:
## Summary

`git-commit:::MSG[:::PATH...]` takes the message as a single colon-CLI token. There's no clean way to pass a multi-line body (subject + blank line + paragraphs), so for any real commit message the caller drops to raw `git commit -F file` — which **skips the op's auto `Co-Authored-By` trailer**, forcing a follow-up `--amend`.

## Reproduction (real, this session)

1. Wanted a commit with a subject + multi-paragraph body.
2. `git-commit:::<subject>` only takes one line; embedding newlines via colon-CLI is brittle.
3. Fell back to `git commit -F .max-commit.txt` → committed fine but **no co-author trailer**.
4. Had to `git commit --amend --trailer "Co-Authored-By: Max <noreply>"` to satisfy the repo convention.

## Proposed fix

Accept a payload route like the edit ops: `git-commit:@-` / `git-commit:@file` with a TOML/JSON body carrying a multi-line `message` field (subject + blank line + body) and an optional `paths` list (default = staged).

The op then writes the message to a temp file, runs `git commit -F`, **and still appends the configurable `Co-Authored-By` trailer** when absent (same logic as today). Keeps the colon-CLI `git-commit:::MSG` form for one-liners.

## Files

- The `git-commit` preset — message handling + the existing co-author trailer logic.

PR #1062, opened 2026-08-08
'amend' taken as a subject, and two remedies that name three of fifteen paths

### LINKED ISSUE #962: git-commit accepts 'amend' as a literal commit message, producing a commit titled 'amend' with no amend route anywhere
```
./supertool 'git-commit:::amend'
```

commits, with the subject line `amend`. There is no amend route in the op, and nothing in the output distinguishes "you asked to amend and I did" from "you asked to amend and I made a commit called amend".

Hit while working #939.

## Why this one is worth more than its size

Every other shape in this family refuses when it cannot honour the request — that is the whole point of #939, #864 and the three-state contract in `docs/validators.md`. Here the op takes a word that is unambiguously an *instruction* in git's vocabulary and silently reinterprets it as *content*. The result is a commit that has to be undone, and undoing it is the one git operation people get wrong under time pressure.

It is also close to irreversible-looking: the commit is on the branch, the message is wrong, and the recovery (`git reset --soft HEAD~1`) is exactly what a nervous operator will not reach for.

## Two things to decide, and they are not the same

1. **Should a bare `amend` be refused?** A commit message that is a single git subcommand name is almost certainly a mistyped instruction. Refusing it and naming what to do instead costs one message and cannot be wrong in a way that loses work. The counter-argument is real though: `amend` is a legal English word and a legal commit subject, so a refusal here is the op guessing at intent. Check whether the refusal should be narrow (exactly `amend`, no other content) rather than a general "message looks like a subcommand" heuristic, which would drift.

2. **Should there be an amend route at all?** Separate question, and the more interesting one. Amending is a normal thing to want after a validator rolls back an edit, and the current answer is to drop to raw git. If the answer is yes, it needs to refuse to amend a commit that is already pushed — amending published history is the failure this op should be preventing, not enabling. If the answer is no, the refusal in (1) should say so, so the reader stops looking.

Do not do (2) as a side effect of (1). Refuse first; a new route is its own change with its own gates.

### LINKED ISSUE #963: git-commit's mangled-pathspec refusal suggests a remedy that produces the exact commit it just refused to make
`git-commit` correctly refuses a Conventional Commits subject, because supertool's single-colon CLI splits on every `:`. The refusal is good — it names the parse, it says nothing was staged, and it offers two routes. **One of the two routes is wrong, and following it commits under a mangled subject: the precise damage the refusal exists to prevent.**

Hit twice in one session, independently — once by me, once by an agent working #956.

## Verbatim

```
$ ./supertool 'git-commit:::test(gh-prs tier)::: pin the narrower vocabulary (#939):::tests/test_x.py'
ERROR: commit message was split on ':' — nothing staged, nothing committed.
  Parsed as: message='test(gh-prs tier)'
             path=' pin the narrower vocabulary (#939)' (not a path, and unknown to git)
             path='tests/test_x.py'
  supertool's single-colon CLI splits on every ':', so a Conventional
  Commits subject cannot survive it. Use a route that does not tokenize:
    ./supertool 'git-commit:::test(gh-prs tier): pin the narrower vocabulary (#939):::tests/test_x.py'
```

The suggested line folds the **paths into the message**. Its `@-` example has the same shape: the `message` key is assigned a literal containing subject, then a colon, then the paths. Follow either and you get a commit whose subject is the subject *plus the file list*, with nothing staged as a path.

The `@-` route in the message is otherwise the right answer; the defect is specifically that the reconstructed example puts paths inside `message` instead of in a separate `paths = [...]` key.

## Why this one is worth more than a wording fix

A refusal's remedy line is the highest-trust text the tool emits. It is read by someone who has just been told they were wrong, in the state of mind where they will paste what they are handed. This tracker's whole thesis is that a tool must not state something untrue about the world; a remedy that does not work is the same failure, in the one place a reader is least equipped to catch it.

Both people who hit it this session already knew the `@-` payload route and used it anyway from memory — which is the only reason neither produced the bad commit. Someone who did not know it would have.

## What the remedy should say

A `git-commit:@-` heredoc with **two separate keys**: `message` assigned a triple-single-quoted literal holding only the subject (and body), and `paths` assigned a list — `paths = ["tests/test_x.py"]`. The file list must never appear inside `message`.

Two keys, paths as a list. Check whether the single-colon suggestion should be dropped entirely rather than repaired — if a Conventional Commits subject cannot survive the tokenizer, offering *any* colon-CLI form invites the next variant of this.

## Adjacent, same call, worth deciding at the same time

`./supertool 'git-commit:::amend'` commits with the subject `amend`. Filed separately as #962.

### LINKED ISSUE #986: git-commit: the payload route cannot stage, and the colon route cannot take a long message
Two agents hit this the same evening and both ended up running raw `git add` — the exact fallback the payload route exists to avoid.

## The bind

`git-commit` has two input routes and each is missing what the other has:

| | message | staging |
| --- | --- | --- |
| `git-commit:::MSG:::PATHS` | single-line only in practice — a multi-paragraph body with colons and backticks is what the payload route is *for* | yes, via `:::PATHS` |
| `git-commit:@payload` | `message` key, takes anything | **no equivalent key** |

So a commit with a real body and specific paths cannot be expressed either way. Both agents wrote the message as TOML, discovered they could not name paths, and fell back to `git add`.

## The error makes it worse

The payload route with nothing staged answers:

```
nothing staged
```

and suggests **the colon form** `git-commit:::MSG:::PATHS` as the fix. That is a correct pointer to a route the caller has usually already rejected — they are on the payload route *because* their message will not survive colon-tokenization. The suggestion reads as "you used the wrong form" when the accurate statement is "this form has no way to do that".

## A second, separate gap in the colon form

One agent passed comma-separated paths to `:::PATHS`. Comma is not the separator; the op stitched all five into one pathspec and git refused:

```
fatal: pathspec '...' did not match
```

The op advertises that it refuses a mangled pathspec, and it did — that part worked. But nothing in the op's own help or in the error states **what the separator is**, so a caller who guesses wrong learns only that their guess was wrong.

## What would close it

- **A `paths` key in the payload**, mirroring `:::PATHS`. That is the whole bug for both agents.
- **`nothing staged` should name the payload-route answer** once it exists, rather than pointing at the other form.
- **State the separator** in the op help and in the pathspec-refusal error.

A `message_file` key was also suggested, but with `message` accepting a TOML literal block I do not think it is needed — the missing capability is staging, not message length.

fdaviddpt:
Closes #962
Closes #963
Closes #986

Two of the three issues are **false as filed**. The friction behind them is real, and it is a different defect in the same file — which is why this closes all three rather than bouncing two.

## #962 — real

`git-commit:::amend` produced a commit titled `amend`. There is no amend route anywhere in the op, so the message was taken literally and a commit landed under a subject nobody meant.

Refused now, before staging and before the header. The refusal names the missing route, both raw-git forms, and the pushed-history warning — so a reader stops looking for an option that does not exist.

**No amend route was built.** It needs its own refusal for pushed history and it is feature scope, which is on the never-auto-merge list; a PR mixing a bugfix with a feature cannot be merged the same way. The refusal says so in words.

`amend` is a legal commit subject, so `SUPERTOOL_ALLOW_LITERAL_AMEND=1` is the way through. A refusal with no appeal would be the op guessing at intent.

## #963 — the body is wrong on both claims, and was wrong when written

It says the `:`-split refusal's `@-` example folds paths into `message`. It has not since `4b991c5` (#751/#771, 2026-08-04) — `_colon_split_refusal` emits a separate `paths = [...]` key.

It says the single-colon suggestion commits under a mangled subject. It does not. Verified against the issue's own reproduction:

```
$ python3 supertool.py 'git-commit:::test(gh-prs tier): pin the narrower vocabulary (#939):::tests/test_x.py'
Files committed: 1
--- subject: test(gh-prs tier): pin the narrower vocabulary (#939)
```

`:::` is the outer separator and survives the tokenizer; the internal `:` is preserved. The reporter typed `:::` where they meant `:`, and the op reconstructed it correctly.

**The real defect on that line is truncation.** Both remedies printed the first three paths:

```
  Modified tracked (15):
    f1.txt … f9.txt          [all fifteen listed]
  Commit the ones you mean, by name:
    ./supertool 'git-commit:::MESSAGE:::f1.txt:::f10.txt:::f11.txt'
```

Paste that and you stage three of fifteen, commit, and get a `✓`. The left-behind variant was worse — `…:::f12.txt…` glued an ellipsis onto the last pathspec, so it failed on paste.

`_colon_remedy(paths)` now names every path under a 20 cap, with a visible `PATH[:::PATH...]` placeholder and a not-shown count above it. **Over the cap it names nothing rather than the first 20** — the first 20 of 200 is the original defect with a bigger number, so it declines instead. Third state.

## #986 — headline claim false for the third time

"The payload route cannot stage." `paths` has worked since #340, `tests/test_git_commit_payload_route.py` has pinned the committed bytes since #415, and `tests/test_git_commit_refusals_1003.py:115` pins staging-via-payload too. Verified live: `git-commit:@-` with `paths = ["a.txt"]` staged and committed exactly `a.txt`.

**Correction to an earlier version of this body**, which said the *same claim* had been filed three times. That overstates it. What has happened three times is that the payload route's capability was re-litigated and each time already worked — #400 (claimed the route did not exist at all) closed by #415, #1003 (claimed no payload route for the *message*) closed by #1034, and now #986, which is the only one to make the staging-specific claim verbatim. The pattern is real; "the same claim" was not literally true of #400 and #1003's filed prose, only of the tests that closed them.

Its two residual claims are real and fixed:

- `nothing staged` now offers the payload route too, with `paths` pre-filled from the list it just printed.
- `_add_failure_lines` states `Paths are separated by ':::'` on any add failure, and offers a comma split **only** for a token git does not already recognise — `a,b.txt` is a legal filename, and re-splitting it would trade a loud failure for a quiet wrong one.

## Verification

RED first: 10 failed, 5 passed. The five passing at RED are deliberate — the false-premise pins and two "must not over-fire" guards, asserting existing behaviour on purpose.

GREEN: 15 new, 133 across `-k commit`, and the full suite at `8484 passed, 72 skipped`.

The full suite earned itself here: it went red once, on this repo's own `test_no_test_subprocess_decodes_by_locale` (#856/#862), catching two `subprocess.run` calls missing `errors="replace"` — a cp1252 decode failure that would otherwise have appeared only on the Windows legs. Both fixed, file swept for a third.

No `os.name` branching, no POSIX literals in assertions, no new subprocess spawn. Suite green on macOS; Windows unverified locally.

`presets/git.json` changed in `description` only — `syntax` is untouched, since it derives the payload field registry.


fdaviddpt:
### Code review

Found 4 issues:

1. **`_nothing_staged_lines` can claim paths were hidden when every one was shown.** `_sample(modified)` and `_sample(untracked)` are each capped at 20 *independently* and printed as two lists, then `candidates = modified + untracked` is capped at 20 *jointly* by `_colon_remedy`. With 15 modified + 15 untracked, both lists print in full — all 30 paths visible — and the remedy prints `30 paths in all, 20 shown and 10 not shown`. False on both counts. This is the defect #963 exists to fix, reintroduced by the fix, directly under the docstring this PR wrote asserting the guarantee (`a remedy is honest only in reference to what was actually shown`). Found independently by two reviewers with different inputs.

https://github.com/Digital-Process-Tools/claude-supertool/blob/842a0a1b0cd54b460cfd08b125a276411d55894e/presets/git/commit.py#L418-L429

2. **Both remedies break on paste for a path containing a quote.** `_colon_remedy` wraps the suggestion in a single-quoted shell string and joins raw paths in, so `it's.txt` closes the string early and the pasted command is a shell syntax error. `_payload_remedy` wraps each path in raw double quotes for a TOML array, so `weird"note.txt` yields malformed TOML. `_worktree_changes` deliberately reads unquoted `-z` paths so spaces and non-ASCII survive — quotes were not considered. Same class as the ellipsis-glued pathspec this PR fixes.

https://github.com/Digital-Process-Tools/claude-supertool/blob/842a0a1b0cd54b460cfd08b125a276411d55894e/presets/git/commit.py#L325-L371

3. **The amend refusal returns before the `Repo:` / `HEAD before:` header, reopening the gap #692 closed.** That commit deliberately made *every* refusal path stamp repo and branch first — nothing-staged, hook rejection, the #751 spilled-message refusal — specifically so a refused call still tells you which repo and branch it refused you in. The new check fires above it. The docstring frames the silence as intentional (`a refused call leaves no trace and reads as the non-event it is`) without acknowledging #692. Moving the check below the header costs nothing; staging still has not happened.

https://github.com/Digital-Process-Tools/claude-supertool/blob/842a0a1b0cd54b460cfd08b125a276411d55894e/presets/git/commit.py#L447-L453

4. **`docs/presets/git.md` states the opposite of the code.** The doc says the over-cap remedy shows a placeholder *with the count not shown*. The code prints the count explicitly (`N paths in all, 20 shown and M not shown`), and the new test asserts it does. `changelog.d/963.fixed.md` describes it correctly, so the doc is the odd one out — a reader of `git.md` alone would conclude the tool goes silent on the count.

https://github.com/Digital-Process-Tools/claude-supertool/blob/842a0a1b0cd54b460cfd08b125a276411d55894e/docs/presets/git.md#L22-L26

Checked and clean: the 20-path boundary in `_colon_remedy` itself (20 joins, 21 trips the placeholder, no off-by-one); the amend match set against over-fire (`amend the readme`) and under-fire (`AMEND`, `  amend\n`, `--amend`); the comma-split guess, which is gated behind an actual `git add` failure on that token; #1034's three-state UNKNOWN logic, byte-identical; and the #751/#771 `paths = [...]` claim, verified against `4b991c5`.

Issue #931, opened 2026-08-07
Every supertool invocation recompiles 17,380 lines — a `__main__` script is never cached to .pyc, and we spawn ~900 of them per CI leg


fdaviddpt:
Measured while researching #917 (Windows is the CI critical path). This is the largest single cost found so far, and unlike the matrix question it trades away no coverage. **It is also user-facing, not only CI.**

## The measurement

macOS, warm FS, `n=6` each:

| invocation | cost |
| --- | --- |
| `python -c pass` (bare interpreter) | 56ms |
| `python -c "import supertool"` (compiled once, `.pyc` reused) | 88ms |
| `python supertool.py version` | **197ms** |

`supertool.py` is **17,380 lines**.

## The mechanism

CPython writes and reuses `__pycache__/*.pyc` for **imported modules**. It does **not** cache the script named on the command line — `__main__` is compiled from source on every single run. So the ~109ms gap between the import path and the script path is the parse-and-compile of 17,380 lines, repaid on every invocation, forever.

Confirmed it is not a bytecode-cache-warmth question:

```
warm (repo, __pycache__ present): 206ms
cold (fresh tmp, __pycache__ removed): 199ms
delta: -7ms   # noise — there is no cache to warm, by design
```

## Why it matters here specifically

The `coverage` job's own comment puts this suite at ~900 spawned children, and `tests/` has 444 `subprocess.run|Popen|check_output|call` call sites across 128 files. Nearly all of them spawn `[sys.executable, supertool.py, ...]`.

- **~900 spawns x ~109ms = ~98s per leg on this Mac.** Windows process creation is roughly an order of magnitude more expensive than fork, and its filesystem is slower — so the Windows share is plausibly larger, though I have not measured it there.
- It matches the shape #917 could not explain: on `pytest (windows-latest, 3.10)`, the whole top-25 durations list is ~140s of a 305s run, with the other ~165s spread thinly across ~6700 tests with no hot spot. A fixed per-spawn tax is exactly what produces that.

**And every user pays it too.** The premise of this tool is "batch 7 ops into one call to save round-trips"; a 109ms fixed tax on every invocation is a tax on that premise.

## What the internet's usual answers are worth here — checked, and they do not apply

[...]

PR #1064, opened 2026-08-08
A workflow with no run was silently uncovered, an elision that read like the log, and no way to list labels

### LINKED ISSUE #804: gh-run does not reconcile the jobs API against the declared matrix (#724's class, unverified for this path)
Raised by an agent while fixing #789. **This is a premise to check, not a confirmed finding** — nobody has caught the dip in the act, and doing so needs a live partial re-run.

## The claim

`gh-pr:status` learned in **#724** to count the run's *declared* legs separately from the jobs the API returns, because the jobs API's `filter=latest` **dips during a partial re-run**: for a window, it returns fewer jobs than the matrix declares. A tally that agrees with a short answer is no floor — it is internally consistent and externally wrong, which is exactly what #724 was filed to name.

`gh-run` has no equivalent reconciliation. It tallies whatever `gh run view --json jobs` hands back and says nothing about whether that set is complete. So during a re-run its `N total` can legitimately be short of the matrix, and the output gives the reader no way to tell.

After #789 the tally is at least arithmetically sound over the legs it *has* — every term sums back to the count. This issue is one layer out: whether the count itself is the whole matrix.

## What needs establishing first

- Does `gh run view --json jobs` actually exhibit the same dip as the path #724 fixed? They may not share a code path or a `filter` default. **Check this before designing anything** — if it does not dip, there is nothing here.
- If it does: is the declared matrix reachable from data `gh-run` already fetches, or does closing this cost an extra call? An extra call per render is a real cost and may not be worth it for a read-only op.

## The judgment call

If it dips, the fix is disclosure and not correction. `gh-run` cannot invent the missing legs; it can say that the set it tallied may be short. Silently rendering a short tally as a complete one is this tracker's most-filed defect class, and "helpfully" padding to the matrix count would be the same defect wearing the opposite costume.

## Prior art

#724 for the mechanism and the reasoning behind counting declared legs; `presets/_checks.py` for the summing contract; `docs/validators.md` §"Declining instead of guessing" for the shape of the disclosure.

### LINKED ISSUE #837: gh-branch's leg count has no second source, so a short job list renders as a complete one — the shortfall check gh-pr has, one scope out
Raised by the agent building #615 (PR #836) and deliberately kept out of that diff — it is a different contract, and the agent said so plainly rather than papering over it. Filing because the surface it affects is a merge gate.

## The gap

`gh-pr` can reconcile a check rollup against the leg count an Actions run *declares*: that is `_checks.shortfall`, and it exists because a tally with no independent second source cannot tell "every leg passed" from "every leg I managed to read passed".

At **branch** scope there is no such second source. `gh-branch` fetches each selected run's jobs and counts what comes back. If a job list returned short — truncated, paginated, partially served — the arithmetic still sums, every state is accounted for, and the verdict renders:

```
Branch master: GREEN
Verdict: GREEN — every workflow on c629679 concluded and every leg passed (17 legs across 2 workflows).
```

That sentence is true about the legs it read and says nothing about the legs it did not. Nothing in the output distinguishes the two.

## Why this one matters more than its size suggests

This is the house defect — an absence produced by the tool read as an absence in the world — and it is landing in **the merge gate specifically**. `gh-branch` exists because both default branches sat red for hours while the board read clean (#615). A false GREEN there is the same failure the op was built to end, arriving through the op.

It is worth being precise about the severity, because it is not the worst class: nothing is destroyed and nothing is lost. It **misreports**, and the cost is trust in the one check that is supposed to be authoritative after a squash merge.

I have not observed it live. Every run so far has summed to the expected count, which is exactly why it is filed as a gap rather than a bug report — the fallback working is what keeps this class invisible, the same way #614/#615/#619 stayed unfiled through four days of daily use.

The question is what the independent second source *is* at branch scope, and I do not think it is obvious:

- An Actions **run** does declare a job count, so per-run reconciliation may be reachable the same way `gh-pr` does it — worth checking whether the field is present and trustworthy on a run that is still in progress, where a `needs:`-gated leg legitimately does not exist yet.
- Pagination is the likelier real-world route to a short list, and a page that comes back full is a **floor, not a total** — the disclosure vocabulary #605 already established.
- Whatever is found, the answer on failure is the three-state contract (`docs/validators.md` §"Declining instead of guessing"): a count that could not be reconciled must say so, not silently become a smaller green.

**Do not trade the loud failure for the quiet one.** Clamping, defaulting, or treating an unreconcilable count as complete would each convert "I could not verify this" into "this is fine", which is the defect rather than the fix.

## Not urgent

`gh-branch` is strictly better than the hand-rolled `gh run list --branch X --limit 1` it replaces, which had no reconciliation *and* picked the wrong workflow. This is a known limitation of a real improvement, documented in #836's body, not a regression.

### LINKED ISSUE #846: A workflow that produced no run on the commit is silently missed by the leg reconciliation — the one case #804/#837 left uncovered
Filed from reviewing the #804/#837 fix, not from hitting it. The agent that built the reconciliation named this gap itself and deliberately did not paper it into the diff — the right call, and it is recorded here so the silence is at least written down somewhere.

## What is now covered, and what is not

After #804/#837 the leg tally is reconciled against a second source. Four shapes, three handled:

| Case | Source | Status |
|---|---|---|
| Re-run dip — legs exist on a prior attempt | `filter=all` distinct names | covered, all three ops |
| Run exists on the commit, legs not yet in the rollup, jobs exist | commit run list + `filter=all` | covered — renders e.g. `3 of 17` |
| Run exists on the commit, zero jobs created yet | commit run list | covered by **named disclosure** — arithmetic cannot see it (zero on both sides), so it is stated in words |
| **A workflow that produced no run on the commit at all** | workflow-file matrix | **silently missed** |

The fourth row is the defect. Nothing on either side of the arithmetic knows the workflow exists, so the tally is internally consistent, sums correctly, and describes a strictly smaller universe than the one the reader believes they are looking at. That is this tracker's most-filed class — an absence produced by the tool, read as an absence in the world — and it is currently **silent rather than declined**, which the three-state contract in `docs/validators.md` forbids.

## Why it was not closed in that PR

Closing it needs the workflow files themselves as a source. That is a different cost and a genuinely harder problem than the other three:

- a matrix can be **computed at runtime** (`fromJSON`, a `setup` job that emits the axis), so the file does not always state the leg count
- **path filters** and `if:` conditions mean a declared workflow legitimately produces no run for a given commit, and a checker that cannot tell that apart from a missing one would manufacture false shortfalls
- `gh-branch`'s existing docs **already decline** to infer from `.github/workflows/*` for exactly this reason, so closing this reverses a documented decision rather than filling a blank

## What a fix has to decide

- Is the honest output here a **decline** (say the expected workflow set is unestablished) rather than a count? A decline is cheap and correct and would close the contract violation without needing to parse a matrix at all. That may be the whole fix.
- If it does read the workflow files: how does it distinguish *did not run because a path filter excluded it* from *did not run and should have*? Getting this wrong trades a silent miss for a false alarm, which is the worse trade on a merge gate.
- Runtime-computed matrices cannot be resolved statically. Whatever lands must say so for those rather than guessing a number.

## Not urgent, and worth saying why

The three covered rows are the ones that actually fire in day-to-day use, including the merge-gate case in #804's own comment. This fourth is the tail. But it is the tail that is silent, and silence is the thing this repo keeps paying for.

### LINKED ISSUE #998: No op lists a repo's labels, so every triage run reaches for raw `gh label list`
Filed from use, per the standing instruction: I reached for raw `gh` because no op covers this.

## What happened

Triaging the board on 2026-08-07, before applying any label I have to know which labels exist — inventing one is the failure mode to avoid, and `gh issue edit --add-label` silently accepts nothing helpful when you guess wrong. The only way to answer it:

```
$ gh label list --repo Digital-Process-Tools/claude-supertool --limit 100 --json name -q '.[].name' | grep -E '^(priority|lane)'
priority-high
priority-medium
priority-low
lane-watch
lane-tracker-ops
...
```

That is a hand-rolled `--json` + `jq` + `grep` pipeline to answer "what can I tag with", which is the exact surface `#454` exists because of.

## Why it compounds rather than annoys

It is the **first** call of any triage or release-planning run, because every later decision depends on it. It is also the first call a delegated triage agent has to make, and a fresh agent has no way to know the answer — so it pays this every single run, forever.

Sharper: the spelling is not stable across repos. `claude-supertool` uses `priority-high` (dash); `claude-remember` uses `priority:high` (colon) and has no `lane-*` labels at all. An agent that assumes one repo's spelling silently mislabels or no-ops in the other. Discovery is not optional here, which is precisely why it deserves an op rather than a remembered pipeline.

## What a fix has to decide

1. **Shape.** A standalone `gh-labels` op, or a flag on the existing board (`gh-issues:labels`)? The boards already carry per-issue labels; what is missing is the *vocabulary*, which is a property of the repo rather than of any issue. That argues standalone — but see the sibling-render argument in `#958` before adding a surface.
2. **Usage counts.** A bare list is less useful than a list with "how many open issues carry this". A label nobody uses is a label to retire, and that is only visible with the count next to it.
3. **Grouping.** `priority-*` / `lane-*` / everything-else is how the list is actually read. Prefixes are a repo convention, not a GitHub feature, so deriving groups means inferring from the names — decide whether that inference belongs in the op or whether flat-plus-description is honest enough.
4. **The GitLab twin.** `gl-labels` has the same argument. Whether both land together is a scope call.

## Not a gap, checked

Two things I believed were missing and are not, both verified the same day by running the op instead of trusting my notes: `gh-issues` **does** carry `milestone=` filtering, a `[m:TITLE]` per row and a `nomilestone` flag (`#864`, shipped); and `gh-pr:N:full` **does** print `Merge commit: <sha>`. Only the label vocabulary is genuinely unreachable.

### LINKED ISSUE #1050: gh-job:N:fail elides the block that IS the discriminator, and gh-pr's tally cannot distinguish 4 failed legs from 4 failed tests
Both hit live tonight, on the same failure, and together they cost a wrong diagnosis that had to be corrected by the agent I handed it to.

## 1. `:fail` elides the load-bearing block

Reading PR #1047's Windows red, `gh-job:93033577461:fail` rendered each failure like this:

```
FAILED tests/test_cached_result_cannot_forge_core_keys_1044.py::test_a_cached_forged_timeout_does_not_disable_the_rollback_guard - AssertionError: ...
  edited ...\\s.json (line 1)
        1 → {"b": 1}

  [validators]
...
  - {"a": 1}
```

The `...` is the `[validators]` block — and that block was the entire discriminator. The one line inside it, `fake : ok (no new errors) 0.0s`, is what tells you the adapter answered clean and therefore that the cache was never read. Without it the failure is just "the bytes are wrong", which is consistent with three different causes.

I recovered it with a second call (`:grep:validators|NOT CHECKED|...`), which is the tell: **I ran a second call to get what the first should have returned.**

The op is not truncating at the end — it says `no tail truncation` — it is eliding the *middle* of each error block. That is the harder case to notice, because the output looks complete.

## 2. The count was ambiguous, and I read it wrong

`gh-pr:1047:status` said:

```
checks: 20 total: 16 passed, 4 failed, 0 pending
  failed: pytest (windows-latest, 3.9) ..., 3.10, 3.11, 3.12
```

`4 failed` is four **legs**. I carried it forward as four *tests*, then listed three test names visible past the elision — and reported "three failures" to the implementing agent. There were **six**, uniform across all four legs. The agent had to correct the premise before it could work.

The two readings were both live for a while, and nothing in either op distinguishes them. The job's own summary line has the real number:

```
6 failed, 7760 passed, 677 skipped in 221.51s
```

That line exists and neither op surfaces it.

## Why this matters more than the individual annoyance

The false picture was *specifically misleading*: with three names showing, it looked like `[timeout]` and `[no_verdict]` passed while their structurally identical dedicated twins failed — which points at test ordering or shared state, a completely different investigation. A uniform six-of-six failure points at the fixture, which is what it was.

So the elision did not just cost a call. It manufactured a plausible wrong hypothesis, and it did it while looking like a complete rendering.

## Shape of the fix

- `:fail` should not elide inside an error block, or must mark the elision as an elision rather than a bare `...` that reads like the assertion's own ellipsis. This repo's whole vocabulary applies: **an absence produced by the tool must not render like an absence in the output.**
- Surface the job's own `N failed, M passed` summary line — it is one line, it is authoritative, and it is the number that settles leg-vs-test immediately.
- `gh-pr:N:status` could say `4 failed legs` rather than `4 failed`, since it already knows it is listing check names.

Class: **misreports**. Nothing is lost; the render supports a confident wrong reading, which is this tracker's most-filed shape.

fdaviddpt:
Closes #846
Closes #1050
Closes #998

## #846 — the merge gate could not tell "passed" from "never ran"

`_declared_for_commit` and `_runs_on_commit` both source from `runs?head_sha=`, so a workflow that produced **no run** is on neither side of the reconciliation and is invisible to it. Live on `master` at `bf66384`:

```
Verdict: GREEN — every workflow on bf66384 concluded (18 legs across 2 workflows)
```

`slow tests` **and** `changelog` were declared at that commit and undispatched. Two, not one — I cut v0.27.0 believing there was a single such workflow.

`presets/_declared_workflows.py` reads `.github/workflows` **at the head SHA** via the contents API and parses `on:` / `name:` textually, with three states: parsed, `triggers=None`, and `(None, reason)`.

**It never downgrades GREEN, and that is the judgment call.** `paths`, `branches`, `if:`, disabled workflows and runtime matrices are all either invisible from here or expensive to evaluate — so a shortfall *concluded* from an absence would be a false alarm sitting on the merge gate. It states scope instead:

```
This covers the 2 workflows that produced a run; 2 declared in .github/workflows at this
commit produced none and are NOT covered — named below.
  no push trigger, so no run on this commit is expected and none of it is covered:
    changelog (pull_request), slow tests (schedule, workflow_dispatch)
```

Non-push absences collapse to one line on purpose: `changelog` + `slow tests` would otherwise print two permanent lines on every `gh-branch:master` — a disclosure that trains you to skip the next one.

A missing `.github/workflows` directory is an **established empty set**, not UNESTABLISHED. A repo that declares nothing there is not a repo we failed to read.

**On the premise:** #849 is merged, and its body reads `Closes #804. Closes #837.` It never claimed #846. So this is genuinely residual work, not a `Closes` that failed to fire.

## #1050 — two defects on the failing-log path

`gap_marker` printed a bare `...`, indistinguishable from the log's own content. Now `... (N lines elided by this op — no error pattern matched them; the log itself is intact)`, and a trailing gap is counted too.

**An unfiled bug the test found:** `prev = -2` emitted a phantom marker above line 1 covering zero lines. Fixed, and the same off-by-one swept out of `_emit_grep_hits`.

`_suite_summary` now surfaces the job's own count — and this is where the live check earned itself. The first implementation required pytest's `=` rule, which matched a fixture built from the issue's prettified quote and **would have matched nothing in any real Actions log**, because Actions is not a tty and the line comes through bare. Fixtures alone would have shipped it dead. The real line is the load-bearing fixture now.

`gh-pr`'s tally gets one clarifier — `(those are LEGS — one check run each, not one test each…)` — emitted only when the failed bucket is non-empty.

**`named_disclosure`'s `failed:` was deliberately not renamed to `failed legs:`.** It is shared by `gh-pr`, `gh-run`, `gh-branch` and `gl-mr`, with 7 assertions across 3 files including a GitLab one, and `presets/gitlab` is another lane. The clarifier answers the issue's intent without touching a shared parser.

## #998 — `gh-labels`

Names, descriptions, open-issue counts, and prefix groups declared as inferred. The count is exact, a `>=N` floor, or `?` — never a number that cannot be stood behind.

Verified on both repos, which also confirms the portability claim: `priority-high` here, `priority:high` and zero `lane-*` in `claude-remember`. 25 labels and 14 respectively.

## Verification

RED: 10 failed in the #1050 file — including two that were **not** the filed bug — plus 15 errors for the absent `labels.py` and 21 for the absent `_declared_workflows.py`. Behavioural RED for #846 is the live master render above.

GREEN: 54 in the three new files, `8516 passed, 72 skipped` full suite, `test_pyright.py` 7 passed, ruff clean, `assemble_changelog --check` and `--check-links` ok.

The full suite was run despite being optional — this adds a new op, and the catalogue tests cover exactly that.

Windows unverified locally. No filesystem paths in the new code (GitHub API paths are always `/`), no `os.name` branch, no POSIX literals in tests. `FileNotFoundError` / `TimeoutExpired` / `OSError` are caught on every spawn, with `test_a_spawn_failure_on_the_issue_read_declines_rather_than_escaping` pinning the #997 shape. A CRLF workflow file is pinned by `test_a_crlf_workflow_file_parses_the_same`; `split("\n")` is used over `splitlines()` deliberately, so sanitising stays `_untrusted.flat`'s job.

`verdict()` gained a defaulted 8th parameter, so no positional caller breaks. `presets/watch/` untouched.

## Two holes worth knowing about

- `_declared_workflows` matches by workflow **name**, so two files declaring the same `name:` collapse and one running masks the other's absence. It under-reports rather than false-alarming, consistent with `_declared_legs`' stated bias — but it is real.
- The GitHub **workflows API** (`actions/workflows`) is *not* the set declared at a commit: it still returns `measure spawn cost (#931)` as `active` although the file is gone from master, plus three `dynamic/` entries. Anything reaching for it as a "declared workflows" source will be wrong.

## Review status

No `/code-review` pass has run on this branch — a subagent cannot invoke a slash command, and this agent correctly declined to shell out to a nested CLI rather than work around it. The independent review is the maintainer's.


fdaviddpt:
### Code review

Invariant verified: **GREEN is never downgraded by the new module.** `scope` is a plain string read in exactly one place (the GREEN f-string), every earlier `return` in `verdict()` is reached before it is read, and `main()` still returns 0 unconditionally. `verdict()`'s 8th parameter is clean — of three call sites, none passes 8 positionals.

Found 4 issues.

1. **`parse_name` does not strip a YAML trailing comment, while `parse_triggers` does — this false-alarms the merge gate.** A workflow declaring `name: tests # the matrix` parses to `'tests # the matrix'`, which never matches GitHub's `workflowName` of `tests`. The GREEN sentence then gains "1 declared … produced none and is NOT covered" for a workflow that ran and passed. This is precisely the false alarm the design set out to avoid.

https://github.com/Digital-Process-Tools/claude-supertool/blob/3269af088030491ec0fbd817031793d03c1c2fde/presets/_declared_workflows.py#L107-L123

2. **A multi-line flow sequence under `on:` is truncated at the first line, which silences the loud case.** `on: [pull_request,` newline `push]` yields `['pull_request']` and `is_push_triggered → False`, so a genuinely undispatched **push**-triggered workflow is routed to `quiet` and collapses into "no push trigger, so no run on this commit is expected" — the exact line #846 exists to print, suppressed. `parse_triggers` should return `None` on an unterminated `[`, not a partial list.

https://github.com/Digital-Process-Tools/claude-supertool/blob/3269af088030491ec0fbd817031793d03c1c2fde/presets/_declared_workflows.py#L150-L157

3. **The fix does not reach two surfaces that publish the same over-claim.** `presets/dashboard/dashboard.py` and `presets/watch/tiers/gh_prs.py` both reuse `branch.verdict()` and never compute `scope`, so the dashboard's master line and the radar still print "GREEN — every workflow on X concluded and every leg passed" with no scope sentence. That is the board a human reads before tagging, which is where #846 was filed from.

https://github.com/Digital-Process-Tools/claude-supertool/blob/3269af088030491ec0fbd817031793d03c1c2fde/presets/dashboard/dashboard.py#L696-L700

4. **On the merge gate the disclosure names a list that is not there.** `pr_merge._default_branch_report` quotes only lines prefixed `Branch `/`Head: `/`Verdict: `/`Legs: `. The scope clause rides on `Verdict: ` and is carried, but it ends "— named below" and the `undispatched_lines` block is not among the quoted prefixes.

https://github.com/Digital-Process-Tools/claude-supertool/blob/3269af088030491ec0fbd817031793d03c1c2fde/presets/github/pr_merge.py#L543-L548

One caveat, fail-safe direction and not filed as a defect: `pr_merge._default_branch_report` runs `branch.py` under a 90s subprocess timeout, and `declared_at` adds 1 + up to 12 `gh api` calls at 15s each over 4 workers — worst case ~60s. A slow contents API can push past the budget, and `pr_merge` then publishes UNKNOWN where it previously published GREEN. It declines rather than false-alarms, but the timeout was not raised to match the new call count.

Checked and clean: no other route makes a should-have-run workflow render as covered (`parse_name` can mangle a name *away* from a match, never *into* another workflow's); the 404-established-empty vs unestablished distinction is implemented rather than asserted, and the two render differently; per-file failure keeps the file with `triggers: None` rather than dropping it.

fdaviddpt:
### Code review (second pass — bugs and history)

Found 6 issues. Verified clean first: the `prev = -2` → `-1` boundary is exact at index 0, index 1, last index, and when the log ends inside an elided region; `_suite_summary` returns `None` rather than a zero and correctly declines `= no tests ran =`, prose, and pytest-quoted forms; `gh-labels`' floor uses `>=` so an exactly-full read is treated as truncated, and labels themselves are `--paginate`d; #849's `unreconciled → UNKNOWN` ordering is preserved.

1. **The new trailing gap marker contradicts the `## Tail` block printed three lines below it.** `_find_error_sections` appends a trailing marker unconditionally, but it has two consumers. The `:fail` path prints only blocks — correct. The **default** `gh-job:N` path prints the same sections and then `## Tail (last 80 lines)` containing exactly the lines the marker just claimed were elided. With an 80-line tail and a last match at line 400 of 500, it prints `... (99 lines elided …)` and then lines 421-500 verbatim. The marker is false about ~80 of the 99.

https://github.com/Digital-Process-Tools/claude-supertool/blob/3269af088030491ec0fbd817031793d03c1c2fde/presets/github/job.py#L564-L572

2. **`_suite_summary` picks the last summary in the job, not the failing one.** `for line in reversed(lines)` takes the first match from the bottom, so a job that runs pytest twice — a second suite step, a `--lf` retry, tox — reports the trailing invocation. A log with `6 failed, 100 passed in 200.0s` followed later by `7 passed in 3.0s` renders `Suite: 7 passed` on a job with six real failures. That is the premise-correction failure #1050 exists to remove.

https://github.com/Digital-Process-Tools/claude-supertool/blob/3269af088030491ec0fbd817031793d03c1c2fde/presets/github/job.py#L493-L503

3. **It also matches an indented summary.** `_SUITE_SUMMARY_RE.match(line.strip())` cannot distinguish the job's own summary from one echoed inside captured subprocess output or a `-s` run. Combined with (2), a nested run printed after the outer summary silently replaces it. Anchoring at column 0 (`rstrip` rather than `strip`) closes this without weakening the tty-less match this PR correctly fought for.

4. **A non-test pytest line renders under a "these count TESTS" header.** `_SUITE_OUTCOMES` includes `warnings`/`deselected`, so `2 warnings in 0.30s` prints `Suite: 2 warnings — … These count TESTS`. Zero tests are being counted.

5. **`gh-labels`' exact `0` is exact only over issues.** `fetch_counts` uses `gh issue list`, which excludes PRs, while the header states `0` means "candidate to retire". A `lane-*` label applied only to open PRs renders an exact `0` beside that advice. The count is honest; the decision text is not.

https://github.com/Digital-Process-Tools/claude-supertool/blob/3269af088030491ec0fbd817031793d03c1c2fde/presets/github/labels.py#L255-L259

6. **This bug and its fix already landed on the GitLab twin, and the twin still carries the unfixed half.** `e5714e2` (#409) set `presets/gitlab/job.py` to `prev = -1` and gave the marker a count — "omitted when nothing is dropped". The GitHub twin was never updated, so #1050 rediscovers a solved problem ~640 issues later and lands a **different wording** for the same concept. And `presets/gitlab/job.py:367` still has `prev = -2` in `_emit_grep_hits`: the phantom marker this PR calls an unfiled find is swept out of `gh-job` and left live in `gl-job`. The PR's lane argument was made about `named_disclosure`, a shared module with cross-lane assertions; this is a private, byte-identical twin.

Two notes rather than defects. **`gh-branch`'s `timeout: 60` in `presets/github.json` is now exactly consumed in the degenerate case** — `declared_at` adds 15s (directory) + up to 45s (12 files, 4 workers) on top of everything else, and the enrichment is unconditional on any commit with runs. And **an unreadable declared-workflows source keeps GREEN with a caveat**, where #849's precedent for an unreadable *second* source was UNKNOWN — see my ratification in the thread.

Issue #654, opened 2026-07-31
Env-var knobs parsed with a bare `int()`: a typo in SUPERTOOL_MAX_COMMITS is a traceback instead of a sentence


fdaviddpt:
## Problem

`presets/git/trail.py` reads its numeric env knobs with a bare `int(os.environ.get(...))`. So:

```
SUPERTOOL_MAX_COMMITS=x  →  uncaught ValueError, full Python traceback
```

The user set a knob wrong. What they get is a stack trace ending in `int()`, which tells them the tool crashed but not which variable, what value it saw, or what a valid one looks like.

`presets/gitlab/mrs.py` already has the shape worth lifting — a `max(0, _int(...))` helper that tolerates junk and clamps.

## Scope, and the reason this is filed now

**PR #651 adds a new instance** — `SUPERTOOL_TRAIL_DETAIL_CAP`, parsed the same bare way. The implementing agent noticed the existing pattern, matched it deliberately to keep that diff reviewable, and said so. That was the right call for the PR and the wrong outcome for the codebase, which is exactly when something belongs on the tracker rather than in the diff.

Worth sweeping rather than fixing the two known ones: `SUPERTOOL_*` numeric knobs are read across `presets/` and `supertool.py`, and the ones parsed defensively should be the rule, not the exception. Report which sites were checked and which were left — a bounded sweep must not render as complete coverage.

## Why it is not merely cosmetic

A traceback is loud, so this is not the silent-absence class that dominates this tracker — and that matters for the fix. **Do not "fix" it by swallowing the error and defaulting**, which would convert a loud failure into a quiet one: the knob would then be silently ignored, and the caller would believe a cap they set was in force when it was not. That is strictly worse than the crash.

The three-state contract applies: honour the value, or say plainly that it could not be read and what is being used instead. Never both silently.

## Suggested direction

One shared helper, used everywhere, that on a bad value prints which variable, the value it saw, and the fallback in force — then continues with the fallback. Clamping negatives is worth deciding explicitly rather than inheriting: `SUPERTOOL_MAX_COMMITS=-5` should not silently mean "none".

## Provenance

Pre-existing in `trail.py`; flagged by the #635 agent and confirmed while reviewing PR #651, which introduces one more of them.

fdaviddpt:
Fixed in #689 (`cca13ba`), which carried no `Closes` line — closing by hand.

[...]

Issue #864, opened 2026-08-05
gh-issues carries no milestone — neither column nor filter — so planning a release falls back to raw gh


fdaviddpt:
Filed from use, per the standing instruction: I reached for raw `gh` because the op does not carry a field.

## What happened

Asked to cut v0.25.0, the first question is "what is in v0.25.0". `gh-issues` cannot answer it:

```
$ python3 supertool.py 'gh-issues:state=open'
  · no PR            12h    0c  #800    priority:high
        test_watch_radar.py asserts on a board that may never have been printed (four sites)
```

Labels, PR-linkage, age, comment count, staleness — all there, all useful, all things my hand-rolled `gh` call had omitted. Milestone is absent from every row, and `milestone=` is not among the accepted filters (`author`, `label`, `state`, `external`, `stale`, `nopipe`, `iids`).

So the call I actually ran was:

```
$ gh issue list --milestone "v0.25.0" --state all --json number,title,state
```

which loses everything the op adds, and reintroduces exactly the hand-rolled-jq surface #454 exists because of.

## Why it compounds rather than annoys

Milestones on this repo are not decoration — five of them (v0.25.0–v0.29.0) were planned in one sitting as the release backlog, eight issues each. Every release tick therefore starts with a milestone query, and every one of those ticks currently pays for a raw `gh` fallback plus a second call to re-acquire the labels the op would have given for free.

That is the chronic class rather than the acute one: nothing about it hurts in the moment, which is why it went unfiled through a whole milestone-planning session.

## What a fix has to decide

- **A column** costs horizontal space on an already-wide render, and most issues on most repos have no milestone at all — so a blank column would be mostly blank. Perhaps render it only when at least one row in the result set has one.
- **A filter** (`milestone=v0.25.0`, and a `nomilestone` bare flag mirroring `nopipe`) is the part I actually needed, and is cheap.
- Both, ideally. The filter is the load-bearing half.
- GitHub's API takes `milestone` as a number or a title on `/issues`; `gh issue list --milestone` accepts the title. Whichever route the op already uses for `label=`, milestone should follow it rather than growing a second mechanism.
- `gl-issues` has the same gap and GitLab has milestones too — #628 already argues the two families should not drift, so decide whether this lands on both at once.

## Not the same as

- **#815** — `gh-issue` (singular) not showing the linked PR's check tally. Adjacent, different op, different field.
- **#769** — the op existing at all. This is a field it was born without.

PR #1072, opened 2026-08-08
The board's absence notes: a page cap `iids` never printed, and a default author filter nothing could escape

### LINKED ISSUE #1067: The board's population is capped at per_page, so `N open` counts the page and every tally derived from it is short — #659 one layer out
## Problem

`#659` fixed the **enrichment** cap: MRs past `enrich_cap` had no `_pipeline`, so `_is_failing` returned `False` rather than unknown and they sorted green. The fix introduced the right vocabulary — `_enriched`, `_unchecked_count()` — "we read this status", not "the loop reached it".

This is the same defect one layer further out, and that vocabulary does not reach it: the **population** is capped too. The tier issues one query bounded by `cfg["per_page"]` — a single `subprocess.run`, no pagination loop. So on a board with more open items than `per_page`:

- `N open` is a count of **the page**, not of the population.
- Every tally derived from it is short by an unknown amount.
- Nothing on the render says a page boundary was hit.

`#659`'s own argument applies verbatim, and its force is if anything greater here: *"a board that lost its feed, or whose watchers are capped, is a report that cannot be trusted to be complete, and that is the thing radar must never render as green."* An enrichment cap makes some rows unknown; a population cap makes rows **absent**, which no per-row state can express.

## It is live, and it has already misled me

`claude-supertool` currently has **81 open issues** against a default page of 50. Every board read during the 2026-08-08 maintainer session was therefore truncated, and I read each one as the board. The failure is invisible by construction — a short list and a complete list render identically.

Surfaced by the `#1024`/`#800` implementation agent while closing the departure half:

> The page cap is a live correctness limit well beyond departures: any board over 50 open items is silently truncated. My change makes the departure half honest and says nothing about the rest.

The departure case is where it bites hardest and is already handled on that branch (`gl_mrs` declines with `len(open_mrs) >= per_page`, footer reads `N off this page` rather than `N left this board`) — but that is one consumer disclosing for itself, not the board being honest.

## Suggested shape, not a specification

The three-state contract, applied to the population rather than the row: a board that hit its page boundary knows it did (`len(rows) >= per_page`), so it can say so once, at the top, and every derived count can carry "of at least N". Declining outright is probably wrong here — a truncated board is still useful — but rendering it as a complete one is the defect.

Worth deciding deliberately: over-disclosing costs a reader one lookup; under-disclosing turns a page limit into a claim that nothing needs attention.

## Also

`radar:--state` has no departure reporting at all, so the inspection surface cannot see what the board now discloses on that branch. Separate, smaller, mentioned here only so it is not lost.

Class: **misreports** — nothing is destroyed and nothing is lost, but a board that reads complete and is not is the surface people glance at to decide nothing needs attention.

Precedent to adopt rather than reinvent: `#652` / `#657` (the `gl-mrs` half), `#659` / `#664` (the tier half), `#486` (a silently narrowed board rendering as healthy).

### LINKED ISSUE #1071: gh-prs reports "No PRs match / 0 PR(s)" on a repo with an open PR — an external contributor's PR is invisible to the board
Reproduced 2026-08-08 on `Digital-Process-Tools/claude-remember`, twice, with and without an explicit state filter.

## Reproduction

From a checkout of `claude-remember`, with PR #325 open:

```
$ supertool 'gh-pr:325:status'
#325 | state: OPEN | mergeable: MERGEABLE | conflicts: no
branch: fix/322-arithmetic-marker-guards -> main
checks: 12 total: 2 passed, 0 failed, 10 pending

$ supertool 'gh-prs:state=open'
No PRs match.
0 PR(s)

$ supertool 'gh-prs'
No PRs match.
0 PR(s)
```

One op says the PR is open and mergeable. The other says the repo has none.

`gh-issues` on the same repo in the same call **did** return both open issues and labelled them `2 external`, so whatever is filtering here is specific to the PR listing.

## Why this is a board defect and not a filter preference

`gh-prs` is the board. It is what a maintainer reads to answer "what is open right now", and an external contributor's PR is the single most important row on it — it is the one with a human waiting, and the one nobody on the team will otherwise stumble over. A default that hides exactly that row renders a repo with outside contributions as an empty repo.

The render makes no distinction between "nothing is open" and "nothing you authored is open". `No PRs match` plus `0 PR(s)` is the strongest possible statement of absence, and it is false about the world.

## Hypothesis, stated as one

The listing signature is `gh-prs[:author=@me,reviewer=@me,state=open,failed,nopipe,iids]`, which reads as though `author=@me` may be applied by default. I have not read the implementation — do not take the mechanism from me, only the reproduction above. Whatever the cause, the two calls disagree and one of them is wrong.

## Fix shape

Whichever default survives, the output must distinguish the three states:

- rows found
- **no rows because a filter excluded them** — say which filter, and how many rows it dropped
- genuinely nothing open

A filtered board that renders identically to an empty one is this tracker's most-filed defect class, arriving in the op used to read the tracker.

## Related

#1067 — the board's population is capped at `per_page`, a different way for the same render to under-report.

fdaviddpt:
Closes #1067
Closes #1071

Two renders that stated an absence they had not established. Both defaults turn out to be correct and are kept; what was missing is the disclosure — plus, on `gh-prs`, any way to ask the other question at all.

## `gh-prs` — the default was not just undisclosed, it was inescapable

`_build_list_cmd` adds `--author @me` when no role key is present. `gh-prs:author=` is refused by the shared tokenizer, and `label=` / `state=` are not role keys, so the default survived every spelling of "show me everyone's open PRs". That is why this ships a flag rather than only a note.

`anyauthor` is the escape. The default stays, because widening it silently would change the population the radar tier keys its departure snapshot on.

The empty board is now three-state, not two:

| Board | Footer |
| --- | --- |
| rows | `author=@me (default) — your PRs, not the repo's; gh-prs:anyauthor for everyone's` |
| empty, rows exist without it | `author=@me (default) excluded 2 open PR(s) — gh-prs:anyauthor to see them` |
| empty, nothing open | `author=@me (default) excluded none — nothing is open either way` |
| empty, probe failed | `whether it excluded anything is UNKNOWN — the check itself failed: …` |

The population probe costs one extra `gh pr list`, fired only when the board came back empty under the *implicit* default. An explicit `author=X` fires nothing. `anyauthor` together with an explicit `author=` is refused rather than resolved by precedence — they are different boards.

`gh-prs` also had no `--limit` disclosure at all; it has one now, measured against rows fetched, before `failed` narrows them.

## `gh-issues` — the cap line existed, `iids` returned before it

`_footer` has printed `capped at --limit N — more may exist` since the commit that created the op, so the *rendered* board never hid a page boundary. `iids` returns earlier and never reached that code, so the machine-readable path — the one a script consumes — was the silent one. The cap note is extracted so both callers get it, and `external` / `stale` / `nomilestone` now report `X excluded N of M fetched`.

## The maintainer's filed evidence for #1067 was wrong, and the correction is the interesting part

The issue was filed on my reading that `per=100` returned 48 of 78 open issues. It does not — it returns all 78. I had piped the call through `head -60`, and a preceding op in the same call consumed the rest of the budget. The `78` I compared against came from `/repos/.../issues`, which counts pull requests as issues; it returned 79 rows, one of them a PR, so the two numbers agreed by construction rather than by corroboration.

No invocation returning 48 could be reconstructed, and the agent that found this declined to invent a mechanism for it — correctly.

The live defect is the `iids` half above, which is real, and is what this fixes.

## Notes

- `main` is split into `main_with_args` + `main` in `prs.py`, mirroring `issues.py`, so the op is drivable without `sys.argv`.
- `_build_list_cmd`'s new parameter is keyword-defaulted, leaving the watch tier's two-positional call untouched.
- The `iids` disclosure went to stdout as a `#` comment, not stderr: `_run_custom_op` drops a successful op's stderr (#654), so the stderr version printed when the module was run directly and vanished through `supertool.py`. A disclosure nobody receives is the defect wearing the fix's clothes.
- `iids` carries only the notes describing rows the caller did not choose to lose — the page cap and the implicit-author exclusion. Not the scope label on a populated board, and not `failed`'s own complement; the caller asked for those.

## Verification

12 tests red before the fix, 13 green after. Full suite run because this touches a preset JSON, two ops, and a module the radar tier imports — it earned it: the first run caught a regression in `test_github_prs.py::test_main_iids_skips_missing_number` that the targeted file could not see. Final `8598 passed, 72 skipped`.

Windows unverified locally; the one platform-sensitive path is the probe's subprocess spawn, which catches `OSError` and is exercised on every platform by a test that raises `FileNotFoundError` directly rather than relying on a real failure.

## Filed separately

- `gl-mrs` has the identical `iids` hole and the identical unreachable `author=@me` default.
- `_run_custom_op` silently dropping a successful op's stderr is a live trap for every preset author.


fdaviddpt:
### Code review

Found 3 issues. All three scored 75 rather than the usual 80 cutoff; posting anyway because the first was independently reproduced by three separate lenses with live output, and it is the same defect class this PR exists to close.

1. `_narrow`'s "of N fetched" is not the fetch count once two client-side flags chain. The closure reads `rows` from the enclosing scope at call time, and each call site reassigns `rows = _narrow(...)`, so `gh-issues:external,stale` reports the second flag's denominator as the post-`external` survivor count. Reproduced: 10 fetched, `external` drops 4 (correctly "4 of 10 fetched"), then `stale` reports "of 6 fetched" — 6 was never fetched. This contradicts the invariant the same PR enforces for the cap note, whose `_cap_note` docstring says the count is measured against the fetch and not against what survived a filter (#864). No test chains two client-side flags.

https://github.com/Digital-Process-Tools/claude-supertool/blob/2fb8ddf29b5c3bb9dae434e240c02b6bceb76293/presets/github/issues.py#L721-L727

2. The new absence disclosures (the `--limit` cap note and the `author=@me` exclusion note) are footer-only on the rendered board, which is the position #651 and #657 moved away from. #651: "Both boards already disclose it, above the table — which is the header position #633 had to learn the hard way." #657: "Header position, because a footer is lost by exactly the consumer that truncates (#633/#635), and repeated in the footer beside the failing count it qualifies." `prs.py` already prints its review-thread enrichment cap in header position, so the file carries both patterns. These two notes are precisely the absence claims #1067/#1071 were filed about, so losing them to a truncating consumer reopens the reported defect.

https://github.com/Digital-Process-Tools/claude-supertool/blob/2fb8ddf29b5c3bb9dae434e240c02b6bceb76293/presets/github/prs.py#L669-L673

3. `prs.py`'s usage line still advertises `iids` as a bare number list, but the `iids` path now prints `# <note>` comment lines ahead of the numbers. `issues.py`'s equivalent line was updated in this PR; `prs.py`'s was not. A script that parses the output reads the docstring, not the diff.

https://github.com/Digital-Process-Tools/claude-supertool/blob/2fb8ddf29b5c3bb9dae434e240c02b6bceb76293/presets/github/prs.py#L21-L23

Issue #1233, opened 2026-08-09
Removing the other repo's name from the maintainer skill leaves bare `#N` citations that now resolve to unrelated claude-supertool issues


fdaviddpt:
Found by the agent doing the `claude-remember` removal, and it is a better finding than the task it came out of.

## The mechanism

The skill cited the sibling repo's issues in two spellings: qualified (`claude-remember#253`) and bare (`#255`), the latter wherever the surrounding prose made the repo obvious. Strip the repo name from the prose and **the qualified ones break loudly while the bare ones break silently** — each now resolves to a real, unrelated issue in this repo.

| In `SKILL.md` | Meant | `claude-supertool#N` actually is |
| --- | --- | --- |
| `#255` — destroys-class `rotate_logs`, the archive overwritten | the other repo | `feat: char-window for head/tail on minified files` |
| `#204` — `--setting-sources ''`, no saves ever again | the other repo | `test: skipif Windows on bash-shebang formatter suites` |
| `#263` — sent to a missing `docs/validators.md` | the other repo | `phpstan: op misses inheritance errors` |
| `#264` — the last-mile catalogue reporter | the other repo | `Add PHPMD to the edit/per-file validator chain` |

**`#263` appears twice in one document meaning two different issues** — the other repo's at the three-state passage, and this repo's real #263 at line 295.

## Why the sweep cannot find these

`grep claude-remember` returns zero once the name is gone. The citations are then indistinguishable from correct ones by any textual check — they are well-formed references to real issues, just the wrong ones. A reader following `#255` lands on a char-window feature and concludes the destructive-bug example is nonsense, or worse, does not notice.

## What is left

The removal handled the passages that were individually named in its brief. Two are outstanding because rewriting them is a design call, not a mechanical one:

- **`SKILL.md:694-700`** — the `destroys` / `fails-to-preserve` / `misreports` ranking table. Its header says *"Example from this repo"*, which is now false: **all four examples are the other repo's.** This is the top-ranked prioritisation rule in the file and it currently has no valid evidence.
- **`SKILL.md:898`** — the last-mile delivery passage, whose `v0.5.0` / `v0.8.2` version numbers are the other plugin's.

## The general shape, worth more than the four rows

[...]

PR #1340, opened 2026-08-11
A bulk issue lookup is a filter on the board, not a new op

### LINKED ISSUE #1323: No op answers 'what are these N issues?' — auditing 124 citations meant 124 round-trips or raw `gh api --paginate`
From the #1233 audit, which had to check 124 distinct issue numbers cited in the maintainer skill.

`gh-issue:N` is one issue per call — 124 round-trips. `gh-issues` lists a **filtered board**, so it cannot answer "these specific numbers". The agent used raw `gh api --paginate`, which is the reflex this repo treats as a filing signal: a `-q` against a `gh` read means a render is being rebuilt by hand.

    supertool 'gh-issues:iids=1233,1240,1251'     # the filter that already exists on gh-prs
    supertool 'gh-titles:1233,1240,1251'          # or a narrower op: number, state, title, one row each

`gh-prs` already takes `iids`, so the first spelling costs a filter rather than a concept. What the caller needs per row is small — number, state, title, and enough to tell "this number means something else here" — which is the whole question a citation audit asks.

Two things it must not do, both this repo's own defect class:

- A number that does not resolve is its **own row saying so**, never a silently shorter list. An audit over a short list reads as "all of these check out".
- If the request is capped, say the cap and how many were dropped — `gh-issues` already does this for `--limit 50` and the same rule applies.

The count is the argument: 124 numbers in one audit, and the same shape recurs every time a doc or a skill file is checked against the tracker, which #1220's `claims` op now makes routine.

fdaviddpt:
`gh-issues:iids=1233,1326,999999` resolves an explicit list of numbers into the same board every other `gh-issues` call renders. It is a filter rather than a new op because the model is the same one — the same rows, ranks, flags and refusals — and only the population differs.

    $ supertool 'gh-issues:iids=1233,1326,999999,1233,1067'
      ✗ is a PR        ?        ?c  #1326    [PR]
      ✗ not an issue   ?        ?c  #999999
      ✓ PR 1072 merged    2d    1c  #1067    [stale] [closed]
      ✓ PR 1321 open      1d    0c  #1233   priority-medium,cohort-7 [m:v0.34.0]
    4 issue(s) | iids: 1 duplicate number(s) collapsed | 2 requested number(s) are not issues here (1 PR(s)) | 1 stale

**Every requested number gets a row.** A number that is a PR, or does not exist, or whose lookup failed is a *row with a reason* — never a missing line. Those are four distinct kinds (`issue` / `pr` / `absent` / `failed`), and a chunk whose call failed is never reported as `absent`, because "I could not look" and "it is not there" are the two states this repo keeps collapsing.

## The issue's own premise was wrong

It said `gh-prs` already takes `iids` and this should match it. It does not: `iids` is a **flag** on both boards — it lives in `_FLAGS` and makes the render print numbers only. There was no value filter in the family to copy. Both spellings now coexist on `gh-issues`, and the reasoning is recorded in the module docstring and `docs/presets/github.md` rather than only here.

## Three decisions worth stating

- **A listing filter beside `iids=` is refused, not silently applied.** There is no `gh issue list` call on that path for `author`/`label`/`state` to narrow, so accepting them would mean ignoring them.
- **No cap by default.** The caller enumerated the population; capping 124 explicit numbers at 50 produces exactly the shorter list this feature exists to prevent. An explicit `per=` still caps and names what it dropped, and `more may exist` never prints for a named population.
- **`gh api graphql` exits 1 on a NOT_FOUND alias while still returning every alias that resolved**, so the body is parsed before the exit code is read. `issue(number:)` is null for a PR and for a nonexistent number identically, so `pullRequest(number:)` is asked alongside it.

## The review found a defect I had shipped

`_filter_tokens.parse()` kept `v[-1]` for a repeated key, so `iids=1,2,nopipe,iids=3,4` silently looked up **3 and 4 only** — the exact absence this feature exists to refuse, manufactured by the tokenizer, with no coverage. A repeated *list* key now concatenates; scalar keys are unchanged. `parse_multi(list_keys=)` rejoins a bare token after a stored list-key value, and stops at the first token placed another way.

**One finding refused.** The reviewer read the new `[state:?]` marker as working "by coincidence" because `_LIST_FIELDS` happens to request `state`. Field selection is the op's own decision, `_milestone_of` has the identical dependency and is the established precedent, and a future row-helper that omits `state` printing `[state:?]` is the designed three-state behaviour rather than a latent bug.

**Found outside the reviewer's scope, because it reads the diff and not what is absent:** `_lookup_repo` flattened every failure into `no_repo_error`, so a missing `gh` binary — `FileNotFoundError`, `[WinError 2]` on Windows — reported `cwd is not a GitHub repo`. Verified live before and after; it now names the real cause.

`_row_from_node` maps the GraphQL node into the `gh issue list --json` row shape, so every cell, rank and refusal below it is untouched.

**Windows unverified**; the 3.9 floor was checked under 3.11 and CI's own leg is the authority.

Closes #1323


fdaviddpt:
Rebased onto `5f046a1` after #1321, #1326, #1335, #1343 and #1331 landed. One conflict, in `presets/github/issues.py`, and the resolution is not the obvious one.

**A naive take-both would have shipped a live bug.** The branch had hoisted a *copy* of the enrichment block into the new `else:` arm. #1326 changed `_owner_repo` to return `(pair, why)`, and the copy still read:

    pair = _owner_repo(rows)
    if pair is None:

After the rebase that is never `None` — it is always a 2-tuple — so it falls through to `_fetch_enrichment(pair[0], pair[1], ...)` with `name=None`, on every board call. It compiles, it does not raise, and git had no reason to flag it because only *one* of the two copies conflicted.

The duplicate is deleted. Master's block is the only one left, and the `iids=` route does not run it, gated on `iids_spec is None` — it already asked for `_ENRICH_FIELDS` in the same GraphQL call that fetched the rows, `_owner_repo` reads the repo off row urls that unresolved rows do not have, and its decline (`repo could not be identified from the listing`) is a sentence that cannot be true where there was no listing.

Three pins added, each verified to **fail** against the naive resolution rather than asserted to be meaningful. One of them found a further defect in the branch's own code: a partial GraphQL failure nulls an alias exactly as a missing number does, so a number the call *failed* on was rendering as `does not resolve to an issue` — the tool's own absence wearing the world's sentence, on the op whose whole job is telling those apart.

Issue #1145, opened 2026-08-08
gh-job:N:grep:PATTERN mangles its own argv when the pattern contains a pipe, and prints a corrupted header instead of refusing


fdaviddpt:
Hit twice tonight by an agent reading a Windows CI leg. Alternation is the natural thing to reach for when grepping a job log, and it silently does the wrong thing.

With a `|` in the pattern the op prints a header with the tail of the job id fused to a fragment of the pattern:

```
Job #93149244183ep
```

— and then dumps the log tail instead of grepping. So the failure is not "the pattern did not match": it is a call that answered a different question and looked like it worked. Reading that output as "nothing matched" is this tracker's most-filed defect class, arriving through the op the merge path uses to read a red leg.

Two things worth separating in the fix:

- The tokenizer is splitting on a character the pattern legitimately contains. Either the pattern needs to reach the op intact (payload route, or last-token-wins), or the op must **refuse** a pattern it cannot carry and say which character broke it.
- Whatever happens, the header must not render a fused id. A corrupted identifier in the header is the tell that argv was mangled, and it is printed as though it were the job being read.

Refusing is fine here — `grep:PATTERN:PATH` handles alternation elsewhere, so the asymmetry itself is worth a sentence in the error, naming the working form.

fdaviddpt:
Correcting this issue's own premise before the fix lands, because the mechanism I wrote is wrong and the record should not carry it.

**The pipe was never the bug.** `{args}` is built as `" ".join(shlex.quote(p) for p in parts[1:])` (`_supertool.py:2186`), so every part is shell-quoted before substitution and a `|` is never a shell operator on this route. Verified live against a real job:

```
$ python3 supertool.py 'gh-job:93211401185:grep:passed|failed'
# Job #93211401185 — Analyze (javascript-typescript)
## grep /passed|failed/ — 1 matching lines (±5 context)
```

Swept the rest — `$`, `;`, `&`, backtick, `*`, spaces, quotes, backslashes all survive. **`:` is the only character that breaks this tokenizer**, because it is core's own split char. Refusing `|`, which this issue leans toward, would have shipped a refusal for a defect that does not exist.

**What actually happened in my reproduction.** The op string reached supertool already mangled: `--- gh-job:93211401185ep:passed|failed ---` is supertool echoing its own argv, so `:gr` was lost *upstream* of it — shell, paste, or hook, still unidentified and not reproducible. That part is not supertool's defect.

[...]

Issue #906, opened 2026-08-07
Every PR edits the same twelve lines of CHANGELOG.md, so every merge re-conflicts every other open PR — move to one fragment file per change


fdaviddpt:
## The cost, measured

**55 of the last 60 merged PRs touched `CHANGELOG.md` — 91%.** The file is 2,670 lines and every one of those changes appends to the same place: the top, under `## [Unreleased]`.

That makes conflicts quadratic rather than occasional. With N open PRs, each merge re-conflicts the other N-1, so a release cycle with 8 PRs pays closer to 18 rebase-and-resolve rounds than 8. `oss_train` exists to automate exactly this churn, which is worth noticing: we built a tool to survive the problem rather than removing it.

## Why `merge=union` is not the fix, and this repo already proved it

The obvious one-line answer is a `.gitattributes` entry:

```
CHANGELOG.md merge=union
```

`tests/test_git_resolve_heading_dup_839.py` exists because that is wrong here. From its own docstring:

> The union itself is not wrong: it mirrors git's `merge=union` driver, line for line. What is wrong is applying it to a document whose meaning comes from *structure* and then reporting only that the markers are gone. When both sides of a hunk carry the same `## [x.y.z]` release heading, the union emits it twice — and every line that sat between the two copies, work that has not shipped, is now parented under a tagged release.

So union merge on this file silently moves unreleased work into a released section. That is a correctness bug in the changelog dressed as a convenience.

## Proposal: one file per change, assembled at release

A `changelog.d/` directory. Each PR adds exactly one new file:

```
changelog.d/895.fixed.md
changelog.d/900.fixed.md
changelog.d/878.added.md
```

Content is the entry body as it would appear today. Filename carries the issue number and the Keep-a-Changelog section.

Two PRs never touch the same path, so **the conflict class disappears entirely** rather than being merged around. At release, an assembler concatenates the fragments under the right headings, writes them into `CHANGELOG.md` beneath a new version heading, and deletes them.

### Why this fits *this* repo specifically

- **Entries here are long, argued prose** — several paragraphs each, with the reasoning that was rejected as well as the change. A fragment file is a better home for that than the top of a 2,670-line file, and it shows up in review as its own diff instead of buried in a whole-file hunk.
[...]

Issue #647, opened 2026-07-31
`git-push:watch` is documented, accepted, and does nothing — the flag is dropped silently and `_spawn_watch` is dead code


fdaviddpt:
## Problem

`presets/git.json:75` advertises:

```
git-push[:force-with-lease][:no-verify][:watch]
```

But `presets/git/push.py:63`:

```python
_KNOWN_FLAGS = ("force-with-lease", "no-verify")
```

`watch` is not in it. The parser at `:76` is `if t in _KNOWN_FLAGS: flags.add(t)` — with **no else**. So `watch` is dropped without a word, `"watch" in flags` at `:309` is always `False`, and `_spawn_watch` is unreachable dead code.

The caller who typed `git-push:watch` gets the ordinary "run this yourself" hint and no indication whatsoever that the flag they passed was discarded.

## Why it matters

Two separate defects stacked on the same line, and both are this repo's signature shape:

1. **A documented feature does not exist.** The syntax string is the contract; someone reading `ops` or `help:git-push` has every reason to believe `:watch` works.
2. **An unknown flag is silently ignored.** This is the more general bug: `git-push:no-verifyy` (typo), `git-push:force` (wrong spelling of `force-with-lease`), or any future flag added to the syntax string but not to `_KNOWN_FLAGS`, all vanish the same way. **A request that disappears is worse than one that errors** — the caller believes it took effect.

The second one is worth fixing regardless of what happens to `:watch`, because it is the mechanism that let `:watch` rot undetected.

## The judgment call

Two independent decisions:

- **Does `:watch` get implemented or removed from the syntax?** `_spawn_watch` exists and appears complete, so this may be a wiring bug rather than an unimplemented feature — check whether it was ever reachable, and note that #642 separately reports `_spawn_watch` failing silently inside a git worktree (where the gitignored `./supertool` symlink does not exist). Implementing `:watch` without fixing that would ship a flag that works only in the live clone.
- **What happens to an unrecognised flag?** Erroring is the honest answer; a warning that still runs the push is defensible. Silently dropping it is not. Per the three-state contract in `docs/validators.md`, a request the tool cannot honour must say so.

## Provenance

Found by an agent while fixing #641; verified independently — `_KNOWN_FLAGS` at `presets/git/push.py:63` against the syntax string at `presets/git.json:75`.

PR #1349, opened 2026-08-11
Disclose how far the base has moved before merging, and refuse a gh-pr mode word that does not exist

### LINKED ISSUE #1257: master went red from two PRs that were each green: gh-pr-merge reads the check tally and never asks how old the base is
## What happened

`master` at `664a26b` failed **13 of 15 test legs on every platform**, on this:

```
FAILED tests/test_ops_roster_1231.py::test_every_shipped_preset_op_declares_a_safety_class
    AssertionError: preset ops with no safety key: ['claims.json:claims']
FAILED tests/test_ops_roster_1231.py::test_every_op_this_repo_loads_declares_a_safety_class
    AssertionError: ops loaded here with no safety key: ['claims']
```

Two PRs merged four minutes apart on 2026-08-10:

- **#1238** (`c324f75`) added the `claims` op — `presets/claims.json`, no `safety` key, because no such key was required at its merge-base.
- **#1243** (`c96dea8`) added `tests/test_ops_roster_1231.py`, which asserts **every** preset op declares one — over a tree that did not yet contain `claims`.

Each was 22/22 green. Neither test was wrong. Their union is red, and nothing in the merge sequence looked at the union until it was already on the default branch.

## Why the gates missed it, exactly

The merge gate is check arithmetic (#454) and `gh-pr-merge` performs it correctly — it read 22 legs, summed the states, found no non-`SUCCESS` leg and reported `CLEAN`. **The tally is a statement about the PR's merge-base**, and the op never asks whether that base is still the default branch head. GitHub's `mergeable`/`MERGEABLE` does not help either: this is not a textual conflict. The two PRs touch **disjoint files** — a `presets/*.json` manifest and a new `tests/*.py` — so every conflict-shaped check passes.

This is the repo's own defect class one layer up: an absence produced by the tool (no failing leg *on that base*) read as an absence in the world (nothing fails).

## The shape that makes it predictable, and therefore checkable

The dangerous pair is not "two PRs open at once". It is **one PR adding a member of a set, and another adding an invariant over that set**, while neither has seen the other. `test_ops_roster_1231` is a whole-tree scan by construction; so are the version-site guards, the splitlines register (`test_preset_git_splitlines_register_1130.py`), the anchored-guard scan, and the docs/catalogue checks. Any PR adding one of those is a PR whose green is only true of the tree it ran on.

## What I would like

`gh-pr-merge` already reads the PR back off the remote and computes the tally. Two candidate disclosures, either useful, and I would rather have the honest one than the clever one:

1. **Report the base distance.** `behind master by N commits, checks last ran on <sha> at <time>` in the preview, always. That is one API call, it needs no heuristic, and a reviewer who sees `behind by 4` merged since the run knows the tally is about a tree nobody has.
2. **Escalate when the delta since the base touches a whole-tree scan.** Harder and needs a definition of "whole-tree scan"; probably a follow-up, and possibly not worth it if 1 turns out to be enough in practice.

Three states, not two, per `docs/validators.md`: if the base distance cannot be established, the preview must say so rather than omitting the line — an absent warning and an unasked question render identically today, which is the whole complaint.

## Judgment call left open deliberately

**Whether this should ever block a merge, or only ever disclose.** Blocking on "behind by N" makes a busy afternoon serial — four merges in one afternoon happened on 2026-08-07 with zero rebases precisely because `changelog.d` fragments killed the conflicts, and reintroducing a required-up-to-date rule would take that back. Disclosing costs nothing and puts the fact in front of the person merging. I lean hard toward disclose-only, but the person building it should say whether a `--strict` arm is worth having for the release window, where the cost of a red default branch is highest.

## Not in scope

The immediate red is being fixed separately by adding `"safety": "read-only"` to `presets/claims.json` (`gh-pr`, `gh-issues` and `git-diff` are all `read-only` and all shell out). That is a one-line repair of the symptom and closes nothing here.

### LINKED ISSUE #1346: `gh-pr:N:<anything>` accepts an unrecognised mode word and renders the default view, so a request that was never honoured reads as a normal answer
Reported by an agent that could not read a review comment on its own PR. Reproduced here before filing:

    $ supertool 'gh-pr:1331:notamode'
    PASS (4.53s)
    # #1331 `read` elides a repeat read of a byte-identical file...
    State: OPEN | Author: fdaviddpt

`notamode` is not a mode. The op returns `PASS` and the default view. So do `:threads`, `:reviews` and `:comments` — three plausible words a caller reaches for, all silently ignored.

## Why this one bites rather than merely annoying

The header of that same default view **advertises the thing the caller is asking for**:

    Unresolved threads: 1 / 1
    Reviews: github-advanced-security (COMMENTED)

So the op states that a thread exists, accepts a request to see it, and returns a view without it, at `PASS`. The agent concluded the information was not available through supertool and fell back to `gh api repos/.../pulls/1331/comments` — the workaround this repo exists to remove, arrived at by an output that gave it no reason to look further.

This is the house defect with the states swapped: not an absence read as an answer, but **an unhonoured request read as an honoured one**.

## Precedent

#647 established that an **unrecognised flag is refused** and exits `2`, and that is documented. Mode words on `gh-pr` are not going through that path.

## What would close it

- An unrecognised mode is refused, naming the modes that exist, exit `2` — matching #647 rather than inventing a second convention.
- Separately, and this is the larger half: **decide whether `:threads` / `:comments` should exist.** The header already computes the unresolved count, so the data is in hand at render time. A caller who reads `Unresolved threads: 1 / 1` has exactly one next question, and today there is no op that answers it.

Worth sweeping the other `gh-*` and `gl-*` ops for the same tolerance and reporting the count — one op accepting junk is a bug, several is the dispatch layer.

fdaviddpt:
Two issues in `presets/github/`, one lane, two files.

## #1257 — the tally is a statement about a tree nobody has

On 2026-08-10 two PRs merged four minutes apart, each 22/22 green, and `master` went red on 13 of 15 legs. Neither test was wrong: one PR added an op with no `safety` key, the other added the invariant that every op declares one, over a tree that did not yet contain the op. Their **union** is red, and nothing in the merge sequence looked at the union.

`gh-pr-merge` performed the #454 arithmetic correctly and always will — the tally is a fact about the PR's merge-base. It never asked whether that base is still the default branch head, and `mergeable` cannot help because the two PRs touched disjoint files.

`## Base` now prints on every preview, in three states — level, `BEHIND` by N with the base head's sha and date, or `UNKNOWN` naming why it could not be established. It prints **before** the gate, so it is on refused previews too, which are the runs a human actually reads. The compare asks about the commit the checks ran on, not about whatever `HEAD` is now.

**Disclose only, no `strict` arm** — and the reasoning is worth keeping. An opt-in refusal is only ever typed by someone who has already decided to be careful, so it adds nothing over reading the line and choosing not to merge. What went wrong that day was not that a merge happened on an old base; it was that nothing on the receipt said the base was old. A required-up-to-date rule would also take back what #906's changelog fragments bought — four merges in one afternoon with zero rebases. `test_a_stale_base_discloses_and_still_merges` pins the decision, and its docstring says plainly that it is a regression pin which cannot go red by construction, rather than dressing it up as a defect test.

## #1346 — an unhonoured request rendered as an honoured one

    $ supertool 'gh-pr:1331:notamode'
    PASS (4.53s)
    # #1331 `read` elides a repeat read of a byte-identical file...

`notamode` was not a mode, and so were `:threads`, `:reviews`, `:comments` — three words a caller reaches for, all silently ignored, while the default view's own header printed `Unresolved threads: 1 / 1`. The op advertised the thing, accepted the request for it, and returned something else at `PASS`. An agent concluded the data was unreachable through supertool and fell back to `gh api`.

Now:

    $ supertool 'gh-pr:1331:notamode'
    FAIL (0.09s)
    ERROR: gh-pr does not have a 'notamode' mode.
    Nothing was read. ... Modes: status, full, diff[:PATH], threads.

    $ supertool 'gh-pr:1331:threads'
    PASS (2.12s)
    Threads: 0 unresolved of 1
    ## RESOLVED — _supertool.py [outdated]
    **github-advanced-security** (2026-08-11): ...

"Nothing was read" is structural, not a claim: the refusal runs before the branch lookup and before any fetch. Validation stops at `diff`, because everything right of it is a PATH.

**`:threads` was built; `:comments` was not.** The default view already prints PR conversation comments, so `:comments` would be a second name for existing output — it is refused, and the refusal names the modes that exist. Review threads had no op at all, which is the half that sent an agent to raw `gh`.

`_fetch_review_threads_detailed` returns `(nodes, error)` where `None` is not `[]`. The pre-existing `_fetch_review_threads` returns `[]` on **every** failure — this repo's own defect class sitting directly under the header line that started the issue. A failed GraphQL call renders `UNKNOWN` and exits 1.

## The brief said exit 2 and was wrong

I asked for exit 2 on the authority of #647. The agent measured instead: #647 establishes *refuse rather than silently drop* and says nothing about the code. The exit-2 sites are `git-push`'s unknown **flag** on a writing op and the release assembler. The live precedent for an unrecognised **mode token on a forge op** is `gh-job`/`gl-job` (#1145) — `supertool 'gh-job:1:notamode'` returns 1 — and `gh-prs` refuses at 1 too. Matched that.

**The sweep #1346 asked for is answered and three more ops share the tolerance** — `gh-issue`, `gl-issue`, and `gl-mr`, which is `gh-pr`'s twin character-for-character. Filed separately rather than widened here; `gl-runners`, `gl-pipeline`, `gl-api`, `gh-job`, `gl-job`, `gh-prs` and `gh-pr-merge` all refuse correctly.

**Windows unverified**; no separator logic, no POSIX literals in the tests, every new subprocess call goes through the existing `_gh`/`_gh_json` wrappers which already catch `FileNotFoundError`/`TimeoutExpired`/`OSError`. CI is the authority.

Closes #1257
Closes #1346


fdaviddpt:
My diagnosis of the red was wrong and worth recording as such. I said the flagged lines were review comment bodies and that `_untrusted.split_lines` was probably right. They are **gh's stderr**, on the non-zero-exit arm, before any reply is parsed:

    if r.returncode != 0:
        return (None, (r.stderr or r.stdout or "").strip().splitlines()[-1:] ...

The comment bodies come off `json.loads` three lines later, are never split, and reach the render as whole strings through `_untrusted.fence()`.

And the helper would have been actively wrong here, for a mechanical reason rather than a stylistic one: `_untrusted.split_lines` breaks on **LF/CR/CRLF only**, so a forged U+2028 would survive *inside* the extracted line and be printed as a decline reason. `str.splitlines()` consumes it. `REGISTER` already carries this exact construct three times with that reason recorded, one of them verbatim — "Narrowing would be worse — it would leave a forged break inside the extracted string instead of consuming it." Registered, and pinned by a behaviour test rather than left as prose.

**The half of my message that was worth sending was the other one.** A guard firing on one line often marks a boundary rather than the bug, and it did: `_render_threads` fenced and flattened every field except the per-comment permalink, which printed **raw at column 0 directly beneath a fenced body** — #965's exact shape. It is the only field in this render that no other read op prints, so no existing scanner covered it: not the refname checks, not the splitlines register. Now `flat()`ed, with a test.

Fifteenth time an agent has contradicted me and been right.

Issue #1401, opened 2026-08-11
session-start.sh is not tool-gated, so a native Windows host with no Git for Windows loses the symlink and the op roster every session


fdaviddpt:
> **Read the comment below before this body.** The central claim here — that a bare `bash` in
> `hooks.json` is resolved by `CreateProcess` on Windows — is **refuted**: a `"type": "command"` hook
> with no `args` is shell form, which is Git Bash on Windows, and a host without Git for Windows has
> no `Bash` tool for a `PreToolUse` hook to match. The CI evidence quoted below is this repository's
> own pytest calling `subprocess.run(["bash", ...])`, not the hook harness.
>
> **What remains open is only the `session-start.sh` half**, which is not tool-gated and so fails every
> session on that host with nothing else failing to mask it. That is the issue; everything below is
> the investigation that got there. #1412 landed as `Part of` this.

`hooks/hooks.json` registers the guard as:

    bash "${CLAUDE_PLUGIN_ROOT}/hooks/pre-bash-guard.sh"

`hooks/session-start.sh` is registered the same way, so this is a class of two, not an instance.

**On Windows, `C:\\Windows\\System32\\bash.exe` is the WSL launcher.** A bare command name is resolved through `PATH` by `CreateProcess`, and on a host with WSL installed but no distribution, that launcher prints — in UTF-16LE — and exits 1:

    Windows Subsystem for Linux has no installed distributions.
    Use 'wsl.exe --list --online' to list available distributions
    and 'wsl.exe --install <Distro>' to install.

It never opens the script. This was **observed** in this repository's own CI, in the tail of a failed assertion:

    CompletedProcess(args=['bash', 'D:\\a\\claude-supertool\\claude-supertool\\hooks\\pre-bash-guard.sh'],
    returncode=1, s...\x00s\x00t\x00r\x00o\x00>\x00'\x00 \x00t\x00o\x00 \x00i\x00n\x00s\x00t\x00a\x00l\x00l\x00.\x00\n\x00\n\x00")

The decode is exact and exclusive: the Microsoft Store alias message ends `…from Settings > Manage App Execution Aliases.`, so it is not that; the WSL message ends `…'wsl.exe --install <Distro>' to install.` — character for character with the bytes above.

## Why it matters more for one of the two

[...]

Issue #1378, opened 2026-08-11
On native Windows the #1347 guard does not run at all, and nothing says so — the gate-that-did-not-run case at platform level


fdaviddpt:
`hooks.json` invokes the guard through bash. Under native `cmd.exe` or PowerShell — no Git Bash, no WSL — the hook **does not execute**, so every raw command a supertool op replaces runs unguarded, and the session looks exactly like one where the guard ran and found nothing to say.

This is consistent with `session-start.sh` and is documented in the README, so it is not a surprise in the design. It is still the house defect at the platform layer: **an absence produced by the tool, read as an absence in the world**, with no disclosure channel at all. A user on Windows has no way to learn their guard is inert, and neither does anyone reading their bug report.

Note what is *not* claimed here: the guard's own code is clean on Windows as far as static reading goes — separator-free, `Path`/`PurePosixPath`, no `os.name` branching, and `shlex` is platform-independent. Windows paths (`ls C:\\Users\\x`) and trailing backslashes were checked to tokenise cleanly, and `UNDECIDED` fires only on genuinely unbalanced quotes. **That grade is reasoned and partly probed, not observed on a Windows box**, and it should keep that label — the maintainer skill's Windows section exists because every entry in it was written by someone who had watched the suite pass on macOS first.

## What would count as a fix

Not "make it run under cmd.exe", necessarily. The minimum is that the *absence is legible*:

- a `SessionStart` line, or a first-refusal-that-cannot-happen notice, stating the guard is not active in this shell
- or a documented one-line check a Windows user can run to find out

The three-state contract applies to enforcement as much as to checkers: `enforcing`, `refused something`, and **`could not run`** — never the shape of a clean result.

Issue #1206, opened 2026-08-09
The git shims in test_status_swallowed_705.py match on $1, so any global git flag added ahead of a subcommand silently un-shims them


fdaviddpt:
Found by the agent fixing #1186, which hit this live. Verified on master.

`tests/test_status_swallowed_705.py` builds its fake git by matching the **first positional argument**:

```
:76   f'if [ "$1" = "{subcommand}" ]; then echo "{message}" >&2; exit {code}; fi\\n'
:84   f'if [ "$1" = "{subcommand}" ]; then {_real("sleep")} 300; fi\\n'
```

git accepts global flags *before* the subcommand — `git --literal-pathspecs status ...` — which moves `status` to `$2`. The shim's test fails, the `exec` fallthrough runs **real git**, and the fake is silently not in effect.

## How it surfaced, and why the quiet case is the dangerous one

While fixing #1186 the obvious mechanism was `--literal-pathspecs`. Adding it moved `status` to `$2`, real git ran, and two decline tests went red. **That was the loud outcome and it was pure luck** — those tests assert a *refusal*, so real git succeeding broke the expectation visibly.

A shim standing in for a **passing** expectation fails the opposite way: real git runs, returns the same answer the fake would have, the test still passes, and it is now asserting nothing about the code path it was written for. A vacuous test is indistinguishable from a working one, forever, and the next person to add a global flag has no signal at all.

`:(literal)` pathspec magic was used instead precisely to preserve the argv shape — which means the landmine is still armed for whoever reaches for a global flag next.

## Fix

Match the subcommand anywhere in the argument list rather than at `$1` — a loop over `"$@"`, or a `case` on the joined arguments. The shim should describe "a git invoked with this subcommand", not "a git whose first word is this subcommand".

Class: `misreports` — a test harness that can stop testing without saying so.

PR #1412, opened 2026-08-11
The guard ladder reaches a Windows Python via `py -3` — and #1401 is refuted, with the reasoning pinned

### LINKED ISSUE #1402: The guard's interpreter ladder cannot find a Python on a standard Windows install, so the gate is off there and says so on every call
`hooks/pre-bash-guard.sh` probes `python3.9` … `python3.14`. **Neither python.org's installer nor GitHub's `hostedtoolcache` creates `python3.9.exe`–`python3.14.exe`** — both create `python.exe` and `python3.exe`.

The bare name `python3` is excluded deliberately (#572): on Windows it resolves to the Microsoft Store App Execution Alias, which **blocks** rather than erroring, so probing it costs a hook timeout on every command.

So on a non-venv Windows install the ladder finds nothing, and the guard reaches its third state — disclosed, allowed, on every Bash call. That is **honest and is the contract working**: `enforcing` / `refused` / `could not run`, never a silent pass. It is also the guard being effectively off for that whole population, while `raw_command_guard` defaults to on and the docs describe enforcement.

Observed indirectly: this repo's own `windows-latest` legs reach the decline path, which is why four tests now skip there with `no python3.9-3.14 on PATH that executes`.

## The candidate answer, and its risk

**`py -3`** — the Windows Python launcher, installed by python.org's installer, a real executable rather than an alias stub. It answers `py -3 -c "..."` correctly and does not block.

What has to be established before it ships:

- **That `py` is not itself an alias** on the hosts that matter, and that a missing `py` fails fast rather than blocking, which is the whole reason `python3` is banned.
- **Ordering.** `py -3` should sit after the versioned names, so a machine with a real `python3.12` keeps using it and the launcher is the fallback rather than the default.
- **That the probe stays a probe.** The ladder demands stdout equal to a known token; a launcher that prints its own preamble would fail that check, which is the right outcome but should be a decision rather than a surprise.

I cannot measure any of it — no Windows box — so this is a filing with a proposed direction, graded **reasoned, not observed**, and it should not be marked done on the strength of a green CI leg alone: CI's runners are `hostedtoolcache` installs, which is one of the two layouts, not both.

fdaviddpt:
Closes #1402
Part of #1401

Two issues briefed together as one Windows class. One was real, one was not, and the refutation is the more valuable half.

## #1402 — fixed

`hooks/pre-bash-guard.sh` probed `python3.9`-`python3.14` and nothing else. Neither python.org's installer nor GitHub's `hostedtoolcache` creates those names — both create `python.exe` and `python3.exe` — and the bare `python3` is permanently banned because on Windows it can resolve to an App Execution Alias stub that blocks rather than errors (#572). So on that whole population the ladder ran out and the guard declined on every `Bash` call: honest, three-state, and off, while `raw_command_guard` defaults to on.

`py -3` is added as the **last** rung, after an activated virtualenv and after every versioned name, so a host with a real `python3.12` never reaches it. `command -v py` gates the spawn, and a candidate still has to print the probe's exact token — a launcher that writes its own preamble is skipped and the guard declines in words, rather than a verdict being built on an interpreter it could not identify.

**Graded reasoned, not observed.** Nobody on this project has a Windows box. The load-bearing unverified claim is that Windows ships no default App Execution Alias for `py.exe`. Putting the rung last is what bounds the cost if that is wrong.

**It reverses a #572 decision, deliberately and only for this script.** #572 recorded `py -3` as considered and dropped, on a layering preference argued without this fact. #572 established that versioned names are not *aliased*; it did not establish that they are *present*, and they are not. `.githooks/pre-push` keeps the shorter ladder on purpose — it refuses loudly and documents `PYTHON=` as the way through, and a loud refusal with an escape hatch does not need the rung that a disclosed allow does.

## #1401 — refuted, and my issue body was wrong in two places

The prerequisite #1401 named was *how Claude Code actually resolves a hook command*. It is documented, and it is not `CreateProcess`:

- A `"type": "command"` hook with **no `args` is shell form** — the string goes to a shell: `sh -c` on macOS and Linux, **Git Bash on Windows**, PowerShell when Git Bash is absent.
- A hook **with `args` is exec form**, and that is where `command` is resolved on `PATH` and spawned directly. **The obvious repair introduces the defect.**
- A native Windows host without Git for Windows has no `Bash` tool at all, so a `PreToolUse` hook matched on `Bash` never fires there.

Composed: the only host where the string could reach a `PATH` search for `bash` is exactly the host where the hook never runs.

**The CI evidence in the issue proves something else.** `CompletedProcess(args=['bash', ...])` is this repository's own pytest calling `subprocess.run(["bash", ...])` — a real `CreateProcess` search. The UTF-16 decode was exact; the attribution was not. And *"everything is broken"* was wrong even in the impossible branch: only exit 2 blocks a `PreToolUse` call, so the brief's *a hook that cannot start must not be fatal* property already held by documented contract.

No behaviour changed for #1401. The tests now refuse `args` on either shipped hook and refuse the two entries drifting apart, so the reasoning survives a reader who does not have it.

**What is left of #1401 is `session-start.sh`, and it is a different defect** — `SessionStart` is not tool-gated, so on a native Windows host with no Git for Windows it fails every session: no `./supertool` symlink, no op roster. No command string fixes that. Hence `Part of`, not `Closes`.

## Tests

RED first, five failures, opening `test_the_bare_name_python3_is_still_never_tried`. Two of the five were only red after positive controls were added — on the first pass they passed vacuously, because a ladder that never reaches `py` satisfies both *the marker is absent* and *it declined* for free.

GREEN: 8 in the new file, 52 with the ladder and guard suites, 835 across the docs/hook/guard selection. Full suite in a disposable clone with `origin` re-pointed: `2 failed, 10671 passed, 93 skipped`. Both failures reproduce identically on base `e34cef6`. **Local green says nothing about Windows.**

## Review

One `Explore` reviewer against the committed diff, four real findings, all fixed:

- **The shim helper wrote CRLF on Windows**, so Git Bash would read a token with a trailing carriage return and the whole file would fail on the one platform it exists for — green here. The macOS-blindness class, caught inside the PR that exists to fix a Windows bug.
- A test asserted only inside an `if`, so reverting the rung made it pass with zero assertions.
- Two stale comments in the script — one claiming the same ladder as `.githooks/pre-push`, now false.

**One refused:** a `sys.platform == "win32"` hard skip, copied from `test_pre_push_interpreter_572.py`. That helper decides error-vs-skip for a *missing bash*, and `_BASH` resolves on Windows CI, so these tests run there — a `win32` skip would make the file vacuous on the only platform it is about.

## Filed separately rather than ridden in

- **`raw_command_guard` is blind to the PowerShell tool, silently** — `hooks.json` matches `"Bash"` only. Widening the matcher would feed PowerShell quoting to a `shlex`-based POSIX tokeniser, and a false deny is worse than a miss, so that is a design decision and not a rider.
- `session-start.sh` on native Windows, as above.
- `.githooks/pre-push` has the same missing-interpreter gap, mitigated by failing loud.


fdaviddpt:
**Pushed a rebase onto `d5dcc3a` with the two CI reds fixed, plus a third bug the review found in the commit before them.** All three were test bugs; none is the product. All three are one shape.

## Both CI reds were platform-independent, and my diagnosis of one was wrong

I guessed xdist for the ordering failure, on the strength of `popen-gw3` in the path. **That was wrong** — `popen-gw3` is just `tmp_path` naming, and the final suite runs green under `-n auto`.

**The real mechanism: `.unlink()` on a shadow does not remove the name.** The ordering test deleted its `python3.12` shadow to simulate *no versioned rung*, which instead **re-exposed the host's own `python3.12`** one PATH entry further along. The ladder answered through it, `py` was never reached, and the control fired — correctly.

It was green locally because this box has no `python3.9`, `3.10` or `3.12` at all, so deleting a shadow of a name nothing has is a no-op. On the runners the name is real. Reproduced here byte-identical by making the test pick a version this host actually installs. Fixed by overwriting the shadow rather than deleting it, and by choosing a name that is genuinely present — so the unmasking path is now adversarial on **every** runner rather than on none.

## The #1206 meta-scan is over-broad, and structurally so

`_DOLLAR_ONE` matches *any* shell literal comparing `$1` — git or not — and exempts by **filename**. A filename exemption blesses every shim that file will ever contain, including a git one added tomorrow.

Narrowed on **git's own grammar** instead of on a name: global flags precede the subcommand, and a subcommand never begins with `-`. So a shim asking whether `$1` is `-c` or `-3` is asking about an option that genuinely is first, with no position to slide to. A shim asking whether `$1` is `status` is the #1206 defect whichever binary it stands in for.

**This shrinks the allowlist rather than growing it.** `test_pre_push_interpreter_572.py` comes off it — exempt by rule now, so that file is covered again for any future non-flag shim. One entry remains, because a supertool op name is spelled exactly like a subcommand and nothing in the text distinguishes them.

Six new tests hold the rule against the three original #1206 shims verbatim, plus the adversarial shapes — including a literal carrying **both** an option and a subcommand operand (not exempt: all operands must be options), and a literal split across constants so no single one holds both. Plus a test that the scan still reads every module it did before, because **a rule can be narrowed by quietly reading fewer files**.

## The third bug: the ladder blinding did not cover `py`

Found by the reviewer, in the *previous* commit, marked "out-of-diff, awareness only". It is a live Windows red: `_blind_the_versioned_ladder` shadowed `python3.9`–`python3.14` and **not `py`**, so `test_no_interpreter_at_all_still_declines_in_words` asserts the guard declined while a real `py` on the host answers the ladder. Absent here and on POSIX runners; present on Windows — the one platform the file exists for.

Reproduced by putting a working `py` on PATH, then fixed and renamed `_blind_the_whole_ladder`, since the old name described the subset that was the bug. Green with and without a simulated host `py`.

**Same shape three times in one PR: a rung the test names but does not shadow.** Worth stating plainly, because it is the failure mode of every test that fakes an environment by subtraction.

## Suite

`2 failed, 10737 passed, 93 skipped` under `-n auto` in a disposable clone with `origin` re-pointed. Both failures reproduce identically on base `e34cef6`. The two affected files re-run against the rebased tree here: 29 passed.

## Filed separately

`test_git_shim_subcommand_1206.py`'s scan sees only `"$1"` **with double quotes**. `[ $1 = status ]` and `[ '$1' = 'status' ]` are the #1206 defect exactly and are invisible to the gate — before this change and after. Pre-existing; widening it means re-scanning the whole suite and is its own PR.

fdaviddpt:
**A claim this PR pinned is false, and it is load-bearing.**

`tests/test_hook_interpreter_windows_1401_1402.py:23-27` composes the resolution reasoning into: *the only host where the hook string reaches PowerShell is exactly the host with no Bash tool, so a PreToolUse hook matched on Bash never fires there at all.*

That composition does not hold. `hooks/hooks.json:5` is `"matcher": "Bash|PowerShell"` — #1413 widened it precisely because Claude treats PowerShell as the primary shell wherever that tool is enabled, and `hooks/pre_bash_guard.py:28-34` records that reasoning. So on a native Windows host with no Git for Windows the guard hook **is** matched, does fire, and cannot run. That is inert — #1378's silent state — not un-asked.

The same sentence had propagated to `docs/configuration.md:251` and into #1401's own comment, where it was the load-bearing step in the refutation of the guard half.

**What survives:** `args` is still the wrong repair, for the reasons this PR established. And #1401's severity claim is still wrong — only exit 2 blocks a `PreToolUse` call, the WSL launcher exits 1, so the command proceeds.

**What does not survive:** "the branch is empty". It is *bounded*, not empty. The distinction that actually carries #1401 is silent failure versus visible failure, not asked versus unasked.

All three sites are corrected in PR #1608 — docstrings and prose only, no assertion changed. Posted here rather than only in that PR because this is where the claim was pinned, and a test docstring is where the next reader will find it.

Issue #770, opened 2026-08-04
Editing an op's `syntax` string for readability silently deletes its @payload route — no error, no warning, nothing at the call site


fdaviddpt:
Found as a near-miss while implementing #751. An agent added a clarifying parenthetical to `git-commit`'s `syntax` string, which removed the op's entire `@payload` route. Nothing said so. It was caught only because the agent went back and re-read `_fields_from_syntax` rather than trusting the green suite — the suite stayed green, because no test asserts that `git-commit` *has* a payload route.

## The mechanism

`_fields_from_syntax(syntax)` derives the `@file` field registry by parsing the `syntax` string. Its own docstring states the behaviour plainly:

> Returns `[]` when any derived field name is not a clean identifier (`[a-z][a-z0-9_]*`). That guards against syntax strings carrying inline prose or punctuation a payload key could never match — e.g. git-resolve's `PATH[,PATH...][:::BLOCKS]  (SIDE: ...)`. Such ops simply have no @file route rather than a falsely-registered, non-functional one.

So a syntax string of

```
git-commit:::MESSAGE[:::PATHS...]  (a MESSAGE with ':' must use this form)
```

yields `[]`, and `git-commit:@-` stops working. Not with an error — the op simply no longer has that route.

## Why this is worth a ticket even though the behaviour is deliberate

The **fallback is right**; registering a field name a payload key could never match would be worse. What is missing is that the fallback is **silent toward the person who caused it**.

The two audiences are different and only one of them is served:

- **The caller** who types `git-commit:@-` on an op with no route gets a normal unknown-route error. Fine.
- **The editor** who just changed a syntax string gets nothing at all. The op's docs still describe a payload route. The suite still passes. The registry quietly holds one fewer entry.

`syntax` reads like a documentation field — it is rendered in `ops` output — so editing it for clarity is the most natural thing in the world, and it is load-bearing parsing input. Nothing in the file says so at the point of editing.

This is the tracker's own defect class pointed inward: an absence produced by the tool (`[]` from a guard that fired) read as an absence in the world (this op has no payload route, as designed). Same shape as #414, #445/#454, #459, #487.

## What would have caught it

Any one of these, roughly in order of cost:

1. **A test that pins which ops have a payload route.** A frozen list — assert the set of ops with a non-empty field registry matches an expected set. Any syntax edit that adds or removes a route then fails a test naming the op, which is exactly the signal that was missing. Cheapest, and it also documents the intent.
[...]

PR #1418, opened 2026-08-11
The operations index was missing seven ops and naming three that do not exist, and `help:` never mentioned the payload route

### LINKED ISSUE #1371: `registry` shipped and never reached `docs/operations/index.md` — the Meta row and the full op table both stop at `gc`
`registry` landed in #1363 (merged 2026-08-11, `c09fa07`). `docs/operations/index.md` does not mention it, in either of the two places that enumerate ops:

    $ supertool 'grep:registry:docs/operations/index.md:5:0'
    (0 results in 0 files, scanned 1 files, limit 5)

`scanned 1 files` says the reader looked, so this is an absence in the file and not in the query.

Both listings are hand-maintained and both are now wrong:

- the **Categories** table, `Meta` row: `cwd`, `repo`, `introduction`, `output-format`, `ops`, `ops:roster`, `help`, `version`, `gc`
- the **Full op table**, which ends at `resolve` and carries no `registry` line

The page also opens with "~40 ops", which is prose that cannot go stale in a way anyone notices, and that is the point: **a shipped op nobody can discover from the reference is the docs half of this repo's own defect class.** The maintainer skill audits `README:N docs:N` per shipped op for exactly this, and `registry` would have come back `docs:0`.

## What is worth deciding, not just patching

Adding one row fixes today. The question the fix should answer is why a list of ops is typed by hand at all when the product has a registry op whose entire purpose is to render the op table from the product rather than from a copy — `registry` is itself the op that did this for the merge-rule table in #1363.

So the candidate shapes, in increasing order of ambition:

1. Add the missing row, and a test that fails when a dispatchable op name is absent from `docs/operations/index.md`. Cheap, and it closes the loop instead of restarting the clock.
2. Generate the table from the registry at docs-build or test time, so the page cannot drift by construction.

Option 1 with the test is probably the honest first cut — a generated table has to decide what to do about the "What it does" column, which is genuinely hand-written prose and better than any one-line description the registry holds.

## How it was found

A maintainer tick, checking whether the previous session's unfiled friction list was real. Two of the five items on that list turned out to be wrong when probed; this one reproduced exactly.

### LINKED ISSUE #1400: `help:paste` documents only the `:::` form, so the `@-` payload route every agent needs is discoverable only by guessing
$ supertool 'help:paste'
    paste:::PATH:::CONTENT

    NARROW USE: replace ENTIRE file. …
    Example: paste:::tests/new_test.php:::<?php\ndeclare(strict_types=1);\n...

The `@-` payload route with `path` and `content` fields is not mentioned. It works — an implementation agent used it today and said plainly that it had **guessed** the field names.

This matters more than a missing line, for two reasons:

**It is the route that has to work.** The `:::` form cannot carry a multi-line body without the caller inventing escapes, which is the whole reason the payload route exists. Every agent writing a file of any size reaches for it, and `help:` is where they are told to look — `ops` output is ~40KB and the agent brief says to use `help:OP` for one op's reference.

**The field names are derived, not free.** `docs/contributing.md` records that a custom op's `syntax` string is **parsed** to derive payload field names for any op whose syntax contains `:::` — so `paste:::PATH:::CONTENT` is what produces `path` and `content`, and rewording that string for readability silently deletes the route while the docs keep describing it (pinned by `TestPayloadRoutePin`, ref #770). A reader who cannot see the derivation cannot know which spelling to use, and the one thing that would tell them is the field list `help:` does not print.

## What is worth deciding

- **Render the derived fields, do not hand-write them.** A hand-written list is a second copy of something the `syntax` string already determines, and this repo has just spent a release on what a second copy of a rule costs (#1363's `registry`, #1347's `replaces`). If `help:` printed the fields it derives, the rewording hazard above becomes visible instead of silent.
- **Whether every payload-capable op gets this**, not just `paste`. `edit`, `replace`, `replace_lines`, `append`, `vim` and `batch` all have the route; if the answer is derived, they all get it for free, which is the argument for deriving it.
- `docs/input-forms.md` owns the mechanics and is correct — this is about the one surface someone actually queries at the moment of writing the call.

fdaviddpt:
Closes #1371
Closes #1400

Two issues briefed together because they are one defect wearing two faces: **a shipped capability a reader cannot discover**. Both were filed as single instances and both turned out to be larger when counted.

## #1371 — the index was wrong in both directions

Filed as *`registry` is missing*. It was seven:

```
AssertionError: dispatchable but absent from the full op table in index.md:
   ['check', 'diag', 'guard', 'hover', 'ops-compact', 'registry', 'rename']
AssertionError: assert 40 == 44
```

And the more interesting half was found by asking the reverse question, which nothing had:

```
AssertionError: index.md names ops the dispatcher does not accept:
   ['blame', 'grep-count', 'read-grep'] — each one sends a reader to `unknown operation`
```

**`blame` had a full row with a syntax cell.** `blame:PATH:LINE` answers `unknown operation: blame`; there is no `op_blame` and never was. The op is `git-blame`, from the `git` preset. `read-grep` and `grep-count` sat beside real names and are argument *forms* of `read` and `grep`, not invocations.

A missing row costs a reader a capability. **A phantom row costs them a call and their trust in the page**, which is worse, and only the reverse assertion finds it.

Two further corrections in passing: `workspace` and `resolve` were in the full table and in no category — the full table serves a reader who knows the name, and Categories is the only route for one who does not, so they are separate assertions now. And `diag`/`hover`/`rename` were fully documented in `docs/mcp-integration.md` the whole time with nothing routing to them.

**The enumeration method matters more than the list.** `_valid_op_names()` is the machine-readable set; `index.md` is two hand-typed listings. The test derives the first and greps the second, so this cannot rot again in either direction. The phantom check's denominator is builtins **plus loaded preset and config ops** — not `_valid_op_names()` alone, or the corrected `git-blame` row would fail — with a guard against the empty-config case, since conftest blacks the config out.

## #1400 — `help:` documented one of two invocation forms

`help:paste` printed only the `:::` form, so the `@-` payload route — the one every agent brief in this repo instructs people to use — was discoverable by guessing. A help entry listing one of two forms is worse than none, because it reads as complete.

`_help_payload_route()` renders from the two registries that actually grant the route, `_AT_FILE_REGISTRY` and `_READ_OP_AT_FIELDS`, with fields derived by `_fields_from_syntax` from the same `syntax` string `help:` prints directly above. **No second copy** — so a syntax reword that silently deletes a payload route (#770's shape) now deletes its documentation along with it, instead of leaving prose describing a route that no longer exists.

An op with no route prints no block. Because every op that *has* one now prints, that silence is an answer rather than a gap.

## Tests

RED first on both, verbatim above, with `_supertool.py` reverted to master for the #1400 file (`AttributeError: module '_supertool' has no attribute '_at_file_route_ops'`). The phantom-op assertion was added after review and reddened on `blame` before any row was touched.

GREEN 8. Full suite in a disposable clone: `2 failed, 10679 passed, 93 skipped`; both failures reproduce identically on master (Python 3.14 ANSI-colourised tracebacks breaking substring asserts). Re-ran both new files against the rebased tree on `031f899`: 8 passed.

**Windows unverified**, and the grade is worth stating precisely: this is string rendering and markdown, no paths, no subprocess, no platform-varying exception types. The one CRLF hazard found was in the test itself — a `split` on a two-newline paragraph break, which finds nothing in a `core.autocrlf=true` checkout, and there is no `.gitattributes`. Normalised in `_index_text()`.

## Review

One Sonnet reviewer as `Explore`, read-only by tool grant; `git diff` after it showed no files written. Two findings, both real, both verified against the product before acting.

**One refused:** a blanket reverse assertion over every backticked op-name-looking token in the tables. That flags correct prose — `ops:roster`, `` `read` `` (range), `` `replace` / `replace_dry` `` are documented *forms*, not op names. The narrower predicate shipped instead: head of a `:`-form, read out of the name cells only.

## Adjacent, fixed

- `resolve` documented only `resolve:SYMBOL`; the dispatcher takes a second argument, and a Python relative import needs it — without it `resolve:.utils` reports `external`, which is an answer rather than an error. Syntax cell corrected.
- `docs/operations/meta.md` had no `ops-compact` or `ops:roster` rows, so the new Meta category link would have sent a reader to a page missing two of the ops it lists.
- Stale prose: meta.md's `help` row claimed it *"discovers an op's payload shape … without grepping source"* — the exact claim #1400 disproves.

## Adjacent, filed rather than fixed

`blame` is still carried in `_PARALLEL_SAFE_OPS`, `_path_arg_positions` and `_read_target_file_only` in `_supertool.py` — dead entries for an op that does not dispatch. Product code, not this lane.

Separately: `diag:PATH` printed *"No diagnostics found … The file has no errors, warnings, or hints"* on a checkout with no `mcp` block, while `docs/mcp-integration.md` describes `diag` as a hard-fail op that errors without a mapping. If that reproduces on a clean config it is this repo's flagship defect — a tool-produced absence rendered as a clean bill of health — but establishing which path ran needs the MCP config surface.


fdaviddpt:
**Correcting one line in the body above, before it becomes a claim somebody chases.**

The PR says the `diag:PATH` observation, if it reproduces on a clean config, is this repo's flagship defect. **I tried to reproduce it in four configurations on `master` `031f899` and it did not.** The op declined honestly every time:

| Config | Output |
| --- | --- |
| no `mcp` block at all | `diag: no LSP configured for broken.py (add mcp.diag mapping in .supertool.json)` |
| `mcp` block present, server has no `diag` tool | same |
| `mcp` block present, `match` excludes `*.py` | same |
| `mcp` block present, `cmd` is a binary that does not exist | `diag: MCP server 'py-lsp' unavailable` |

All four on a syntactically broken file, so a false clean bill would have been visible. None of them printed *"No diagnostics found … The file has no errors, warnings, or hints"*.

So the unconfigured paths are correct and this is **not** a defect on the evidence I have. What I could not test is the one path that would still be worth someone's time: a server that **starts and answers** with an empty diagnostics list for a file it did not actually parse. That needs a working `cclsp`, which I do not have here.

Nothing filed. Recording the negative result rather than leaving the body's conditional standing, because an unreproduced hypothesis in a merged PR body reads as an open defect to the next person who greps for it.

Issue #1376, opened 2026-08-11
Retiring the three `block` JIT rules the shipped guard replaces is a sequencing decision, not a cleanup


fdaviddpt:
#1347 ships a PreToolUse guard driven by the op registry. Three hand-written markdown rules under `.claude/jit-context/tools/00-manual/` cover the same ground and are the ones #1347 was filed against:

- `gh-pr-view-merge-have-ops.md`
- `gh-list-limit.md`
- `supertool-no-cut.md`

Deleting them looks like the obvious follow-up and **is not safe yet.** The shipped hook lives in the plugin's `hooks.json` and does not become active for this checkout until the plugin is released, so removing the markdown now leaves this repository — the one place the conventions are actually enforced today — with **no live enforcement at all** for the window between the merge and the tag.

That window is not theoretical: v0.34.0's tag is gated on two audit rounds, and the catalogue pin moves on its own schedule afterwards.

## The decision

Three orderings, and the difference is which failure you accept:

1. **Retire at the merge.** Simplest, and enforcement is off in this repo until the release lands.
2. **Retire at the tag.** Correct in principle, and needs somebody to remember at a moment already carrying five version sites and a cohort freeze — the maintainer skill's own record says that is where things get dropped.
3. **Keep both and accept a double block for one release.** The reviewer on #1347's branch claimed this is already happening; that claim is **wrong today** for the reason above, but it becomes true the moment the plugin is installed from a release that carries the hook. A rule firing twice with two different messages is its own defect.

Whichever is chosen, the `.md` file is inert on its own — **`00-index.tsv` in the same directory is what the hook reads**, so a retirement that deletes the markdown and leaves the index row, or vice versa, produces a rule that exists and never runs. Those look identical to a rule that runs and never matches, including in the hook's own log.

And the retirement must be verified by **running the command the rule forbids** and checking what refuses it, and with which message. Two `block` rules were found dead on 2026-08-10, one of which had never fired since the day it was written.

Filed rather than folded into #1347's PR by the implementing agent, deliberately: it is a sequencing call about the release, which is the maintainer's, not the author's.

fdaviddpt:
Correction, measured against \`feat/1347\` (\`474d291\`) by feeding the hook real PreToolUse payloads: **retirement is per-rule, not a sweep.**

[...]

PR #1437, opened 2026-08-11
The guard could not see past a command's own options, and said OK — a normaliser, a third state, and a whole shell that was never gated

### LINKED ISSUE #1413: `raw_command_guard` matches only the `Bash` tool, so wherever the PowerShell tool is enabled the gate is off and says nothing
Surfaced by the #1401/#1402 agent while establishing how the harness resolves a hook command (PR #1412). It is not that PR's blast radius and is deliberately filed rather than ridden in.

`hooks/hooks.json` registers the guard on `"Bash"` and on nothing else. Anthropic's hooks documentation says, in as many words, to *match `Bash|PowerShell` in hooks that inspect shell commands*, and that *wherever the PowerShell tool is enabled, Claude treats PowerShell as the primary shell and routes shell commands through it*. The rollout is progressive **even on hosts with Git Bash installed**, and it is opt-in on macOS and Linux.

So on any session where that tool is live, every shell command bypasses the registry-driven guard. The output a user sees is **nothing** — which is exactly what a clean pass looks like — while `raw_command_guard` defaults to on and the documentation describes enforcement.

## Why this is the house defect and not a coverage gap

The guard's whole contract is three states: BLOCKED naming the op, OK, and UNDECIDED when the command did not tokenise or the registry could not be enumerated — with UNDECIDED never rendered as OK. That contract holds perfectly *inside* the hook. It says nothing about the case where the hook is never invoked, and a hook that never runs is indistinguishable at the call site from a hook that ran and approved.

This is the same shape as #1378 (on native Windows the guard does not run at all, and nothing says so), one layer out: there the platform removes it, here the tool routing does.

## Why the one-word fix is wrong

Widening the matcher to `Bash|PowerShell` feeds PowerShell command strings to a `shlex`-based POSIX tokeniser. PowerShell quoting, escaping and its backtick continuation are not POSIX, so the tokenisation would be wrong in ways that produce **false denies** — and a false deny is the expensive direction for this guard, because the only escape from a wrong block is `raw_command_guard: false`, which disarms every mapping in the repo. That asymmetry is the reasoning PR #1406 already established and #1384's comment records.

So this is a design decision, with at least three shapes:

1. **Match PowerShell and tokenise it properly** — a second tokeniser, and the `replaces` entries would need to say which shell their `argv` is written in.
2. **Match PowerShell and always return UNDECIDED** — no false denies, and the user is told the gate cannot read this shell rather than being told nothing. Cheap, honest, and it converts a silent absence into a disclosed one, which is this repo's standard answer.
3. **Leave the matcher and disclose elsewhere** — `guard` and the session-start roster could state which tools the gate covers.

(2) looks right to me on the repo's own precedent, but I have not measured what an UNDECIDED on every PowerShell call costs a user in noise, and that number should decide it rather than my instinct.

## What is not established

- **Whether the PowerShell tool is enabled for anyone using this plugin today.** I have no telemetry and no reporter. The documentation says the rollout is progressive; that is a statement about availability, not about our users.
- **Whether `hooks.json` can express an alternation at all**, or whether it needs two entries. Not checked.
- Graded **reasoned, not observed** on the Windows half, per the convention #627 set: nobody here has a Windows box.

### LINKED ISSUE #1421: The guard's view of a command is narrower than the command, and the gap renders as OK — global options and the PowerShell tool
Found by the #1384 agent while mapping `presets/git.json`, verified by hand on that branch and on `master` `cdaea33`.

`replaces` entries are matched token-for-token against the **start** of a simple command. Git's global options sit between the command word and the subcommand, so a mapping written as `{"argv": "git status"}` cannot see the same command carrying one:

```
$ python3 supertool.py 'guard:git status'
BLOCKED   `git status` is replaced by supertool's `git-status` op.

$ python3 supertool.py 'guard:git -C /tmp/x status'
OK: nothing in this command is replaced by an op loaded here.
```

Same subcommand, same intent, opposite verdicts. The same hole exists for `git -c key=val <sub>`, `git --git-dir=D <sub>` and `git --work-tree=W <sub>`.

## Why it is not fixable inside `replaces`

The only spelling the schema offers is `{"argv": "git -C"}`, which claims **every** `git -C` subcommand — including `git -C W tag` and `git -C W push origin v1.2.3`, neither of which any op replaces. That trades a silent miss for a wrong block, and on this guard a wrong block is the expensive direction: there is no per-command escape, only `raw_command_guard: false` repo-wide.

So the fix belongs in the matcher, not in the data: **normalise git's global options out of the command before scoring**, so `git -C P status` scores as `git status` and the existing entry applies unchanged.

## Why it matters more than a coverage gap

`.claude/jit-context/tools/00-manual/git-C-has-cwd.md` exists precisely because `git -C` is a shape people reach for, and `cwd:` is the op-side answer. So the one raw git form the repo already knew was worth a rule is the form the shipped guard cannot see — and the guard is the thing that was supposed to supersede those markdown rules (#1376). Retiring `git-C-has-cwd.md` on the strength of #1384's mappings would remove the only enforcement that shape has ever had.

It is also the guard's own house defect: a command that walks past every mapping renders as `OK: nothing in this command is replaced by an op loaded here`, which is byte-identical to a command that genuinely has no replacement. There is no third state saying *I could not see past this command's options*.

## What a fix has to get right

- **Which options take a value.** `-C`, `-c`, `--git-dir`, `--work-tree`, `--namespace`, `--exec-path` each consume the next token or use `=`; `--paginate`, `--no-pager`, `--bare` do not. Getting that table wrong swallows a subcommand.
- **`--` and a subcommand that begins with `-`.** A normaliser that skips leading dashes indefinitely will skip past the subcommand itself on a malformed command and score something that was never typed.
- **Whether this generalises.** `glab` and `gh` have their own pre-subcommand options (`--repo`, `-R`), so a git-only normaliser may be the wrong shape. Worth asking before building the narrow one.
- **Three states.** If normalisation cannot be performed confidently, the verdict is `undecided`, never `OK` — the existing contract already has that state and this is exactly what it is for.

- Whether any *shipped* mapping other than the git family is affected. The `gh` entries were written before this was known and I have not probed `gh --repo X pr view`.
- Whether the normalisation belongs in `_guard_score` itself or in the tokenising step above it. That is an implementation call I have not looked at closely enough to have an opinion on.

fdaviddpt:
Closes #1421
Closes #1413

Two instances of one class: **the guard's view of a command is narrower than the command, and the gap renders as `OK`** — byte-identical to a command that genuinely has no replacement.

```
$ supertool 'guard:git status'            BLOCKED
$ supertool 'guard:git -C /tmp/x status'  OK: nothing in this command is replaced by an op loaded here.
```

`argv` is matched from the first token, and git's global options sit between the command word and the subcommand. `_GUARD_GLOBAL_OPTIONS` is a per-command-word table of which pre-subcommand options take a value; the walk strips them before scoring, so `git -C P status` scores as `git status` and every existing entry applies unchanged. **Only tokens in the table are ever skipped**, which makes the "skip dashes until a positional" trap — where a normaliser walks past the subcommand itself — structurally impossible rather than merely tested against.

A pre-subcommand token in neither list is **`undecided`, with the token named**. That is the state both instances needed and neither reached, and it is why they were one issue: decide it twice and the two halves of one gate disagree about what *I could not tell* looks like.

## The brief I wrote was wrong, and the agent refused it with the receipt

I asked for a test asserting `git -C W push origin v1.2.3` **must still run**. It must not, and I verified this myself on `master` before merging the refusal:

```
$ supertool 'guard:git push origin v1.2.3'
BLOCKED     Use: supertool 'git-push'
```

`presets/git.json` claims a bare `git push` and `guard_command`'s own docstring names that outcome as deliberate. Leaving the `-C` spelling clean would have made `-C` a **documented bypass for the one op that can destroy someone else's commits**. It is pinned instead as a consistency test against its own un-prefixed spelling. The negative that does matter — `git -C W tag v1` stays clean — is pinned separately.

## The issue's untested claim, answered by measurement

Does this generalise past git? Yes, and **not the way round I expected**:

```
$ gh --repo cli/cli version      unknown flag: --repo
$ glab --repo x/y version        glab 1.86.0
```

gh puts `--repo` on leaf commands, glab on its root. An entry for gh would have blocked a call that cannot execute; `gh --repo O/R pr view 1` is `undecided`. **glab was affected and gh was not.**

## #1413: a whole shell with no gate

`hooks.json` matched `"Bash"` only, so wherever the PowerShell tool is enabled the hook never fired at all — a silent no-gate, the more dangerous of the two instances. The matcher now takes both, and a non-Bash tool gets an `undecided` **disclosure** rather than a decision, because `shlex` on PowerShell quoting produces false denies. The disclosure is narrowed to commands naming a declared binary, whole-word plus `.exe/.cmd/.bat`: a line printed under every command is one nobody reads.

**This repo's own `.claude/settings.json` had the identical `Bash`-only matcher** — #1413's defect sitting in our own wiring, fixed in the same commit.

## What did not get retired, and why

`git-C-has-cwd.md` **stays**. `git -C W diff` and `git -C W log` are still `OK`, because `git diff` and `git log` are recorded absences rather than mappings. #1376 is unblocked for the `status` row only. A test had recorded that rule as kept *because* `git -C P status` was clean — this change falsified its reason, and it now says the real one.

## Adjacent, fixed

`_supertool.py` annotated two names it never imported (`FrozenSet`, `Callable`). PEP 563 kept them as strings so nothing raised at import and ruff carried both F821s as "pre-existing" — but `typing.get_type_hints(_supertool)` was a `NameError`. Pinned by a test that resolves the module's own annotations.

## Windows

**Observed** (macOS, gh 2.x, glab 1.86.0): the root-flag asymmetry above. **Reasoned, not observed**: the `.exe/.cmd/.bat` suffix handling, and that #1378 leaves the hook unreachable on a native-Windows host with no Git Bash — so #1413's fix reaches the opt-in PowerShell tool on macOS and Linux, and the residual gap is #1378's. Both labelled as such in the code and in `docs/configuration.md` rather than implied closed.

Full suite in a disposable clone: **2 failed, 11156 passed**, both reproducing identically on `042c6a9`.


fdaviddpt:
### Code review

Found 4 issues:

1. `--version`, `--html-path`, `--man-path`, `--info-path` are listed as ordinary pass-through booleans, but they are terminal like the bare `--exec-path` the adjacent comment deliberately excludes. `git --version status` prints `git version 2.46.2` and never runs `status`, yet the normaliser walks past the flag, scores `["git", "status"]` and returns `BLOCKED  Use: supertool 'git-status'` — a wrong block on a command git does not execute as `status`, which is the state the third state exists to avoid. Verified by hand on this branch.

https://github.com/Digital-Process-Tools/claude-supertool/blob/6ca48afc7de580702403de8fac4f7900790f6b7a/_supertool.py#L15632-L15643

2. The `guard` op's own description says the hook "discloses on every PowerShell one", while `hooks/pre_bash_guard.py` narrows the disclosure to commands naming a replaced binary, `changelog.d/1413.fixed.md` says "it does not print under every call", and `test_a_powershell_command_naming_nothing_mapped_stays_silent` pins the silence. The op description is the surface `ops` and `help:guard` render, and it is the one of the four that overstates.

https://github.com/Digital-Process-Tools/claude-supertool/blob/6ca48afc7de580702403de8fac4f7900790f6b7a/.supertool.json#L231-L233

3. Same overstatement in the configuration doc: "a PowerShell command is never refused, only disclosed as undecided" reads as a disclosure under every PowerShell call, when most get neither.

https://github.com/Digital-Process-Tools/claude-supertool/blob/6ca48afc7de580702403de8fac4f7900790f6b7a/docs/configuration.md#L188-L190

4. `docs/presets/git.md` line 29 still describes `git -C P status` as a shape "no mapping can reach until #1421" — in the same file this PR edits a few lines below to state that it now blocks. CLAUDE.md: "When you find a wrong line in any of them, correcting it is part of the task you are already doing", and a stale line about a *missing* capability is the shape it calls the most dangerous.

https://github.com/Digital-Process-Tools/claude-supertool/blob/6ca48afc7de580702403de8fac4f7900790f6b7a/docs/presets/git.md#L28-L30

Not blocking this PR, filed separately: the guard and `.claude/jit-context/tools/00-manual/git-C-has-cwd.md` now both block `git -C W status` with two different messages — the double block #1376 called its own defect.

Issue #1045, opened 2026-08-08
cargo-check's 2-segment floor still cross-attributes: a workspace-root package prints src/lib.rs, which suffix-matches every member


fdaviddpt:
Found by round 2 of the v0.27.0 release audit. This is #1037 one segment up, in the fix for #1037.

## Mechanism

`MIN_MATCH_SEGMENTS = 2` admits `src/lib.rs` and `src/main.rs`. That is exactly what cargo prints for a package at the **workspace root**. A member's absolute target ends with `/src/lib.rs`, so the root package's pre-existing error is charged to the member file — carrying its real rustc code, so it is a finding rather than a non-verdict, on a `rollback_on_fail` validator.

Root-package-plus-members is a common Rust layout, not an exotic one.

## Reproduction

Against the real `_same_file`:

```
MATCH  cargo='src/main.rs'  target='/repo/crates/other/src/main.rs'
```

#1037's own three vectors **are** genuinely closed — `main.rs` against a foreign crate, rel-vs-rel, abs-vs-rel all return no. The floor works; two segments is simply not enough when the shallow path is a real package path rather than a bare basename.

## What the audit could not reproduce, and it matters

The inverse failure — a mass demotion of real findings, which is what #754 exists to prevent and what any fix here risks reintroducing — **did not reproduce**. Anchoring `target_raw` to cwd is safe because cargo's side is workspace-root-relative and therefore always a true suffix of the absolute target; abs-vs-abs correctly requires equality, since `"/" + shorter` can never suffix-match another absolute path.

So the fix has room to tighten without paying #754's price. It needs to distinguish "shallow because it is a basename" from "shallow because the package is at the workspace root" — which is a fact about the workspace layout, not about the string.

## Known cost, already accepted

The floor demotes 1-segment paths, so `build.rs` and `Cargo.toml` at a crate root always render as non-verdicts. That is documented in the `_tail_match` docstring as intended. Flagging it only because `build.rs` is an everyday file rather than a hypothetical, and the two costs should be weighed together when this is fixed.

Class: **fails-to-preserve** — a correct edit reverted over another package's pre-existing error.

Issue #1037, opened 2026-08-07
cargo-check's suffix match has no segment floor — a foreign file's diagnostic is still attributed to the validated file


fdaviddpt:
Found by the round-1 security audit of `v0.26.0..fb33ae5`, in #754's fix, merged this evening.

## Mechanism

`_tail_match` (`validators/cargo-check/cargo-check.py:53-62`) is:

```python
a == b or a.endswith("/" + b) or b.endswith("/" + a)
```

The boundary check is there and it is right — `src/xmain.rs` does not match `main.rs`. What is missing is a **floor on how few segments may match**, and the comparison runs against the raw `{file}` argument as well as the resolved one, so a single-segment target matches any file with that basename anywhere in the tree:

```
_same_file("vendor/crates/foo/src/main.rs", "crates/foo/src/main.rs") → True
_same_file("crates/other/src/main.rs",      "main.rs")                → True
_same_file("/abs/elsewhere/src/lib.rs",     "src/lib.rs")             → True

_parse_errors("vendor/crates/foo/src/main.rs:4:5: error[E0425]: …",
              "crates/foo/src/main.rs")
  → [{'line': 4, 'col': 5, 'code': 'E0425', ...}]
```

A real error code rather than `adapter` means it is **not** a non-verdict, so with `rollback_on_fail` an unrelated crate's pre-existing error reverts a correct edit.

`tests/test_cargo_sibling_attribution_754.py` covers single-segment same-directory collisions only, which is why this survived.

## The tension the fix has to hold

#754's own reasoning explains why joining onto the crate root is wrong: cargo prints paths relative to the *workspace* root, so `crate_root / src_file` double-counts the member directory and demotes every real finding to a non-verdict — the same misreport pointing the other way, and the quieter of the two.

So the answer is not "go back to joining". It is a suffix match that cannot be satisfied by too little: enough segments to be a real identification, plus the boundary rule already present, plus whatever the resolved-vs-raw comparison actually needs. The audit's cases above are the test table.

Class: **fails-to-preserve** — a correct edit reverted over a defect in a file the edit never touched. Same damage route as #969, one layer out.

Issue #754, opened 2026-08-01
cargo-check attributes a sibling file's crate error to the file under validation


fdaviddpt:
Split out of #753, which fixed `cargo-check`'s *misclassification* (a non-zero exit with no located diagnostic is now `code: "adapter"`) and deliberately left its *misattribution* alone.

`cargo check` analyses the whole crate. `validators/cargo-check/cargo-check.py:_parse_errors` (`:45-67`) matches every short-format diagnostic in the output and does not compare `src_file` to the file under validation, so an error in a sibling module is published as an error in this file.

Reproduced on rust 1.9x with a two-file crate — `src/main.rs` is valid, `src/sibling.rs` has a type error:

```
$ cargo check --message-format=short --quiet          # cwd = crate root
src/sibling.rs:1:27: error[E0308]: mismatched types: expected `i32`, found `&str`
error: could not compile `demo` (bin "demo") due to 1 previous error
```

```
$ validators/cargo-check/cargo-check.py .../src/main.rs
{"tool": "cargo-check", "file": ".../src/main.rs", "ok": false, "count": 1,
 "errors": [{"line": 1, "col": 27, "code": "E0308",
             "msg": "mismatched types: expected `i32`, found `&str`",
             "source_context": []}]}
```

`file` says `main.rs`, `line` is a line in `sibling.rs`, and `source_context` is read from `src_file` — a path relative to the *crate root*, resolved against the adapter's cwd, which is why it comes back empty here rather than merely wrong. Editing `main.rs` therefore fails validation for a defect it does not contain, and `rollback_on_fail` can revert a good edit for it.

## Why it was not fixed in #753

Every change in that PR reclassifies without removing anything from a caller's error list. A `src_file == file` filter removes errors, and the question it forces has no obvious answer:

**What should `cargo-check` report for a healthy file in a crate that does not compile?** Filtering leaves zero located diagnostics, which falls into the new `adapter` branch and yields "cargo check exited 101 without reporting anything about the file". That is *true* — and arguably useless, since the crate genuinely does not build and the caller may want to know. The alternatives (report the sibling's error but flagged as out-of-file; report it as a finding with `line: null`; keep `ok: false` with a message naming the other file) are a design decision about the per-file schema, not a regex fix.

[...]

Issue #669, opened 2026-07-31
Validator coverage is PHP-shaped: the second-tier gaps, and the shape of the imbalance


fdaviddpt:
## The measurement

Thirty-three adapters across `validators/` and `formatters/`. **Ten are PHP** — `phplint`, `phpstan`, `phpstan-mcp`, `phpmd`, `phpmd-mcp`, `phpunit-mcp`, `psr`, `rector-mcp`, plus `php-cs-fixer` and `phpcbf` in `formatters/`.

That is not an accident and not a mistake — DVSI is a PHP codebase and the tooling grew where the work was. But it means several ecosystems have **syntax-only** coverage, which answers "does this parse" and nothing else. The pattern across the whole list: where PHP gets a linter, a static analyser, a mess detector, a test runner and two formatters, most other languages get one compile check.

Filed separately, with detail, because each is a real piece of work: **shellcheck**, **ruff**, **eslint**, **gitleaks**. This issue records the rest so the inventory does not evaporate.

## Second tier

| Want | Have | Gap |
| --- | --- | --- |
| **mypy** | `pyright` | Different findings, not a subset. Teams that run both do so deliberately; pyright is stricter on inference, mypy on plugin-typed libraries. Lower value than `ruff`, which is why it is here rather than in its own issue. |
| **pytest adapter** | `phpunit-mcp` | The asymmetry is stark: a PHP test runner exists, and the Python repo running 5,300 tests has none. An MCP-shaped adapter would mirror `phpunit-mcp`. Note #345 — that adapter fabricated failures from shared warm-process handles, so copy the *shape*, not the process model. |
| **clippy** | `cargo-check` | `cargo check` compiles; clippy is where the lints live. |
| **golangci-lint** / **go vet** | `gofmt-check` | `gofmt` is formatting only — Go's semantics are unchecked. `go vet` ships with the toolchain, so it is the cheap half. |
| **sqlfluff** | *nothing* | No SQL coverage at all. DVSI writes raw SQL in commands; #650's sibling class (SQL injection) has a whole DVSI skill devoted to it. |
| **tflint** / **checkov** | `terraform-check` | Syntax vs. policy. Lower priority — no active Terraform work. |
| **yamllint** | `yaml-check` | Worth checking whether `yaml-check` is parse-only before filing; it may already be sufficient. **Not verified.** |

## The constraint every one of these shares

[...]

PR #1443, opened 2026-08-12
Go had a formatter and no semantics - a go-vet adapter, and a SCHEMA.md rule that told adapter authors to reproduce two fixed bugs

fdaviddpt:
Part of #669

**This one is feature scope and I am not merging it on my own authority** — a new validator adapter is a new capability, not a bugfix. The doc correction inside it is not; see below, and it can be split out if you would rather take that half now.

## #669 re-derived first: live, but only half

The issue is 11 days old. Measured against the tree rather than trusted:

| Issue claim | Now |
| --- | --- |
| "Thirty-three adapters" | **39** — 36 in `validators/`, 3 in `formatters/` |
| shellcheck, ruff, eslint, gitleaks "filed separately" | **all four shipped** |
| Second tier: mypy, pytest, clippy, golangci-lint/go vet, sqlfluff, tflint/checkov, yamllint | **all seven still open** |
| "yamllint — worth checking whether `yaml-check` is parse-only. **Not verified.**" | **verified parse-only** — `yaml-check.py:55` is one `yaml.safe_load(fh)`, no rule engine |

Seven adapters is seven PRs, so this builds one and leaves the rest on the issue, which stays open as `Part of`.

## Why Go and not the other six

It was the only language in the issue's table with **zero semantic coverage where the tool needs no install** — `go vet` ships with the toolchain exactly like `gofmt`, so the issue's own objection (every added language costs a `skipped` on every contributor's machine) does not apply here. Rust already has `cargo-check` for type resolution, which makes clippy a smaller marginal gain. `sqlfluff` is the other zero-coverage language but needs a dialect decision.

Three states throughout: `go` absent → `absent()`; no `go.mod` above the file → a plain `skipped` that does not escalate; rc≠0 with nothing parseable → `tool_fault`, never a pass.

`code: null` for analyzer findings, deliberately — `go vet`'s text output carries no rule id and inventing one states a fact the tool never gave. The one distinction go itself makes is the `vet: ` prefix on load failures, so those are `severity: error` / `code: "load"`.

## The adjacent finding, which is a bugfix and is the part worth reading

**`validators/SCHEMA.md:104` prescribed a path-comparison rule that reproduces two bugs this repo already paid for.** It told adapter authors to *"compare paths by suffix on segment boundaries"* — advice `cargo-check` abandoned twice: **#1037** (a suffix match charged `/abs/elsewhere/src/lib.rs` to `src/lib.rs`, on a `rollback_on_fail` validator) and **#1045** (the two-segment floor added to save it also failed, because a workspace-root package prints exactly `src/lib.rs`).

Following the document as written reproduces both. It now prescribes absolute-path equality anchored on the base the tool actually printed against, with both cases spelled out, pinned by two tests.

`validators/common/pkg_paths.py` implements that: fold case **then** normalise separator, because `ntpath.normcase` rewrites `/` to `\` and the other order is #1005. `normcase` is injectable, so the Windows behaviour is asserted from macOS.

## Review, including where the reviewer was wrong

An `Explore` reviewer (Sonnet, read-only, against the committed diff) found a real one: `errors[:MAX_ERRORS]` cut the edited file's own finding when it sorted behind 50 siblings. Reproduced (`assert None == 6`), fixed in `1c1dc61`, pinned by two tests.

It also said "no vacuous tests" while looking at one. The first ordering test used a target named `a.go`, and `go vet` emits in file order — a target sorting first survives any cap by luck, so the test passed before the fix. Renamed to `z.go`, which made it fail correctly. And it reported no stale adjacent prose while reading the paragraphs around the SCHEMA.md rewrite, not noticing the rewrite *was* the stale-prose finding.

## Left to file

`cargo-check.py` carries a private copy of the attribution logic (~190 lines) that `pkg_paths.py` now duplicates in smaller form. Adopting the shared module means deciding what to do with the `cargo metadata` workspace-root lookup, across three pinned test files (#754, #1037, #1045) — its own PR. Both modules' docstrings say so, so the duplication is not read as an oversight.

The issue's open question — should a configured-but-absent validator warn at startup rather than `skipped` per edit — is partly moot: validators are opt-in per project, so nobody gets skips they did not ask for. It only bites a repo that configures one and then loses the tool.

## Tests

Red first, 19 failing. Green: 21 in the new file; guard suites (schema drift, adapter tool-vs-file, registry dispatch, html-check registration, cargo #754/#1037/#1045, syntax floor, changelog findable) at 159 and 115 passed.

Full suite skipped deliberately — one new adapter, one new `validators/common/` module with no existing importer, docs, one example-config entry; no core, no shared fixture, nothing renamed. Windows unverified beyond the injected-`normcase` assertions; CI is the authority.


fdaviddpt:
Both `except OSError: pass` handlers are gone, but not by adding the comment — the silence was not correct, so commenting it would have written down something false.

Neither handler could produce an empty comparison: `forms` always holds the unresolved spelling before the `try`, so nothing crashed and nothing attributed to `unknown` for want of a form. The defect is the **direction of the loss**. `Path.resolve()` is the only thing in this module that sees through a `subst` drive or a symlinked root — that is why both forms are kept at all (#754). When it raised, the equality test ran on unresolved spellings alone, so a non-match no longer ruled out an alias, and `attribute()` still returned `"other"` — a positive claim the caller publishes to the reader verbatim as "another file in this package". The module's own docstring says only `"other"` is entitled to say another file is at fault, and under an incomplete comparison it is not.

Both sites now carry the refusal out as a value, and `attribute()` answers `"unknown"`. The asymmetry is deliberate:

- a **match** under an incomplete comparison is still `"this"` — matching is positive evidence;
- a **non-match** under one is `"unknown"` — not matching, under a comparison that was cut short, is not negative evidence.

Direction of the old bug, for the record: dropping a form can only shrink the match set, never invent a match. So it could demote the edited file's own finding to "another file", and could never charge a foreign file's finding to this one. The safe direction — and still a false sentence printed to a human, which is the class this repository files bugs about.

The handler is also not dead code, which is worth stating since that would be the other reason to leave it bare. Measured on 3.14/macOS, `resolve(strict=False)` swallows symlink loops, `ENAMETOOLONG` and nonexistent paths; it is reachable in practice on Windows, where `_getfinalpathname` raises and CPython's non-strict fallback catches `FileNotFoundError` alone, letting `PermissionError` and `[WinError 123]` through.

Pinned red-first by `attribute(...) == "unknown"` under a refused resolution, plus a guard that the fix does not over-fire — `test_a_refused_resolution_on_a_path_nobody_tried_to_resolve_changes_nothing`, which fails against a naive implementation using cross-platform `is_abs` instead of host `os.path.isabs` as the attempt guard.

A third instance of the same shape is **left deliberately**: `validators/common/source_context.py:17` returns `[]` on `OSError`, so an unreadable file is indistinguishable from a file with no context to show. Seven adapters consume it and `SCHEMA.md` documents the field, so whether a receipt should carry "source context unavailable" is a contract decision across all of them rather than a rider on this PR. Filed separately.

Issue #1296, opened 2026-08-11
`node --check` times out at 30s on windows-latest 3.11 — twice in one night on the same leg, and the test renders it as a product verdict


fdaviddpt:
Observed twice on 2026-08-11, both times on `pytest (windows-latest, 3.11)` and nowhere else.

    FAILED tests/test_html_check.py::test_broken_inline_script_is_a_finding
      AssertionError: expected the error pinned to source line 6, got
      {'line': None, 'col': None, 'severity': 'error', 'code': 'adapter',
       'msg': 'timeout — node --check did not return within 30s for the <script> block starting at line 5'}

`9389 passed` on the same leg, and the first occurrence carried two other timeout-shaped failures beside it (a `gh-job` render returning nothing, a writer whose content never landed within 6.0s). Both runs cleared on a re-run.

## The adapter is behaving correctly and the test is not

`html-check` **did the right thing**: it could not get an answer in 30s and said so, with the reason and the block it was checking. That is the three-state contract working — `skipped`, not a fabricated `ok`.

The test then asserts a pinned line number, so a correct refusal renders as a **failing test**, which reads as a defect in the product. That is the #1205/#1218 shape: the harness turning an environment limit into a product verdict, in this repo's own defect class, inside the suite meant to detect it.

## What to decide, and it is not "raise the timeout"

Raising 30s to 60s buys nothing that a slower runner does not take back, and it makes every genuinely hung `node` cost twice as long.

- **Assert the class, not the pin.** The test's subject is *a broken script is a finding*. A timeout is a finding of a different kind; the test could accept `severity: error` and assert the line only when the adapter actually parsed. That keeps the coverage and stops the false red.
- **Or gate on `node` responding at all**, the way symlink tests gate on the privilege (#1232) — a probe, then a declared skip that says the runner could not answer. That is honest but adds another skip the capability line cannot see (#1274).

I lean to the first: the pin is a stronger assertion than the test needs, and the weaker one still fails if the adapter reports clean.

## Evidence grade

**Observed**, not reasoned: two runs, same leg, same test, both cleared on re-run. What is *not* established is why 3.11 specifically — the same job on other Python versions passed both times, and I have no explanation for that beyond runner load. Do not write one into the fix.

Issue #794, opened 2026-08-05
test_a_real_adapter_spawn_answers_well_inside_the_shared_budget reddens an unrelated PR when a Windows runner is slow


fdaviddpt:
## What happened

`pytest (windows-latest, 3.9)` on PR #792, job #92208096314:

```
FAILED tests/test_adapter_timeout_policy_702_658.py::test_a_real_adapter_spawn_answers_well_inside_the_shared_budget
  AssertionError: expected a clean verdict on a two-line PHP file with nothing wrong with it —
  phplint reported ok=False count=1 after 30000ms: [adapter] timeout
1 failed, 5845 passed, 559 skipped in 509.00s
```

That PR changes `presets/github/issue.py` — a GraphQL linked-PR lookup — and its tests. It does not touch adapters, spawning, phplint, or timeouts. One leg of four Windows legs; the other three passed, as did every ubuntu and macOS leg.

## Why I am filing it rather than just re-running

The standing rule here is that a single-platform red is usually real, and the running score was **7 for 7** before this. So the default is to read the log rather than re-run, which I did. But the mechanism does not connect: there is no path from a GraphQL query builder to a phplint adapter spawn, and the assertion is a wall-clock budget rather than a behavioural claim.

The test spawns a **real** adapter and asserts it answers well inside a 30s shared budget. On a Windows runner already 8.5 minutes into a suite, a cold process spawn hitting 30s is a statement about the runner, not about the code — and the failure message says so itself: `[adapter] timeout`, not a lint verdict.

So this is the first plausible flake on the board, and worth naming as one, because the value of "7 for 7, read the log" comes entirely from it staying honest.

## What makes it worse than an ordinary flake

It fails **someone else's PR**. The author sees a red leg on a diff that cannot have caused it, and the cheap read — "unrelated, re-run" — is exactly the instinct this project spends effort suppressing, for good reason. A test that trains people to dismiss reds costs more than the coverage it provides.

Adjacent and already noted: four tests in `tests/test_watch_radar.py` assert byte-exact empty output from `radar.main`, so any legitimate board line reddens them. That brittleness cost two full CI rounds on #786. Same family — a test whose failure does not mean what it appears to mean.

## What would fix it

Options, none obviously right:

- Make the budget assertion **relative** to a measured baseline spawn on the same runner rather than a fixed 30s wall clock.
[...]

PR #1462, opened 2026-08-12
A cold `go vet` reddened three unrelated PRs - the suite declines a stall instead of publishing a machine limit as a product defect

### LINKED ISSUE #1461: `go vet` times out at 60s on a cold Windows runner, so the adapter I shipped an hour ago reddens an unrelated PR
Landed with #1443 (`bc5d7b5`) and surfaced on the next PR to branch from it.

```
FAILED tests/test_validators_go_vet_669.py::test_a_clean_package_is_clean
  AssertionError: expected a clean verdict — go-vet reported ok=False count=1
  after 60138ms: [adapter] go vet timed out after 60s

1 failed, 10675 passed, 784 skipped in 479.11s
```

`pytest (windows-latest, 3.10)`, job #94089832110, on PR #1457 — **a PR that touches `presets/_checks.py` and nothing else**. `bc5d7b5` itself was green on the same leg, so this is not deterministic.

## Mechanism

`go vet` on a cold module cache compiles the standard library before it can analyse anything. On the runner that is minutes, not seconds; on a warm cache it is well under the 60s the adapter allows. The 90s `timeout` in `.supertool.example.json` is the op-level cap and does not govern the adapter's own 60s wall.

The failure mode is the one this repo already names: **a platform-speed limit rendered as a product verdict**. The adapter reports `ok=False count=1` — a finding — where the honest answer is that the tool did not finish. Two things wrong, and the second is worse than the timeout:

1. the budget is too tight for a cold cache on a slow runner;
2. **a timeout renders as a finding rather than as `tool_fault`**, so a validator that could not answer looks like a validator that answered badly. #1443's own body claims `rc≠0 with nothing parseable → tool_fault, never a pass` — a timeout is a third path that reaches neither.

## Why it is mine

I merged #1443 knowing neither Florian nor I can evaluate Go, on the basis that it is opt-in, matched to `*.go`, and cannot roll back an edit. All three still hold — **the exposure I did not check is the test suite**, which runs the adapter on every leg regardless of whether anybody enabled it.

## Not established

Whether the first `go vet` on a GitHub Windows runner is reliably over 60s or merely sometimes. One green leg at `bc5d7b5` and one red at the same commit's child says it is variable, which is the worst case for a fixed budget. Measuring that on CI before picking a number is the work.

fdaviddpt:
Closes #1461

Three PRs in one afternoon went red on a Windows leg none of them touched, each within 200ms of the adapter's 60-second wall.

## My framing of the defect was wrong, and the refusal is the important part of this PR

I briefed this as *"a timeout renders as `ok=False count=1` — a finding"*, and *"reaches neither arm"* of `rc≠0 → tool_fault`. Both halves are wrong about the code, and I verified that myself before accepting it:

- `go-vet.py:219` routes the timeout to `_adapter_error(...)`. The nothing-parseable arm at `:253` calls `_adapter_error(file, tool_fault(...))` — **the identical shape**. `refusal.tool_fault()` is a message builder, not a verdict shape.
- `ok:false, count:1` with `code:"adapter"` is not a finding, it is *the channel the schema gives an absence*. `validators/SCHEMA.md:64`: reserved for "a binary that is absent, **a timeout**, output that would not parse … It stays a real error — `ok: false`, `count: 1` — because the process ran and something is broken that someone has to fix; **a fault routed to `skipped` is a validator quietly reporting clean**."
- `docs/validators.md` renders such a result as `NOT CHECKED`, never subtracts it from a baseline, never caches it, never rolls back on it.

So the fix I asked for — route it to `skipped()` — would have made a hung validator quiet and exit 0. `refusal.py:206` declines that by name: *"there is deliberately no value that turns a finding into a skip, because that would be a mute button and this repo declines those."*

## The real defect is one layer up, and the abstraction already existed

`_adapter_verdict.skip_if_stalled` was built for exactly this in #794 and extended for it in #1296 — html-check, same failure, three months ago. **go-vet is the fourth suite to hit this and the first that never adopted it.**

**And the guard was blind to the payload anyway.** `stalled_at_its_own_wall` required `"timeout" in msg.lower()`; go-vet says `"go vet timed out after 60s"`. All three other clauses matched. The one shape the predicate exists to classify was the one it could not see — now `WALL_PHRASES = ("timeout", "timed out")`, strictly widening, with the `code == "adapter"` and `duration_ms ≥ inner_s` clauses still gating it so a real finding cannot be swallowed.

## The timeout is unchanged, and that is the measured answer

macOS, go1.22.3, `GOCACHE` pointed at an empty directory:

```
cold  3.54s        warm  0.24s        ~15x
```

The 60s wall governs a **warm** `go vet` — 250× headroom. Raising a product timeout to cover a first run on a CI image slows the real failure mode (a `go vet` that hangs on somebody's edit) for every user of the tool, in order to fix a test. Wrong instrument.

The warm-cache option is what survives, and it transfers — checked rather than assumed: warming in module `example.com/warm` made a never-seen `example.com/probe` vet in **0.22s**, because `GOCACHE` is content-keyed rather than module-scoped. A module-scoped autouse fixture pays it once; the warm-up returns a *reason* rather than raising, and a reason goes to `warnings.warn` — not a skip (that would take the toolchain-free attribution tests with it) and not a failure (that publishes a slow runner as a broken adapter).

**No Windows number was picked and CI was not touched.** With `skip_if_stalled` in place the change no longer depends on whether a cold Windows `go vet` is reliably over 60s: over it, the suite declines with a named reason instead of publishing a machine limit as a product defect.

## Tests

Red first: 6 failed, and the 2 that passed are the guard tests — a real finding still reaches its assertion, a wrong clean verdict still fails. Both green throughout, or the weakening ate the gate.

Green: 157 across six suites. **Cold-cache proof of the whole go-vet suite with `GOCACHE` at an empty directory: 27 passed in 6.83s.** Full suite: 2 failed, 11412 passed, 93 skipped — both failures reproduced at `37de583` in a detached worktree with none of these files involved.

## Adjacent, fixed

- `_adapter_budget.py:12`'s docstring asserted every adapter's decline is `msg: "timeout"`. False for go-vet and pyright — and it is the sentence that would have told the next author the predicate was safe.
- Two paragraphs in `docs/validators.md` §"Declining instead of guessing": the adapter's own wall as a fourth route to the same absence, and why a cold toolchain cache is paid once rather than budgeted for.

## Left for filing, and one is a product defect

**`validators/pyright/pyright.py:75` emits a timeout as `code: "timeout"`, not `"adapter"`.** The core's `NOT CHECKED` rendering, the cache exclusion and the rollback exemption all key on `"adapter"` — so pyright's timeouts are cached, subtracted from baselines, and **can revert an edit**. That is a product defect rather than a test defect, and it wants its own PR.

Also: `test_mcp_stop_crash_574` (Python 3.14 coloured tracebacks) and `test_tsc_check` (unlocated error), both pre-existing on master.

## Windows — reasoned, not observed

No path-string assertions; all joins are `Path`/`tmp_path`. The warm-up catches `OSError` (covering `FileNotFoundError [WinError 2]`, #997) and `TimeoutExpired`, both pinned by injecting the exception rather than provoking it. `GO_WARMUP_S = 120 * platform_factor()` → 360s on Windows, through the existing multiplier rather than a new constant.

**Stated as unverified:** under xdist each worker runs its own module-scoped warm-up, so a cold Windows runner pays several concurrent cold `go vet` runs. Go's build cache is concurrency-safe and each is bounded and cannot fail the run, but the wall-clock cost on a cold Windows leg is not measurable from here. CI is the authority.

`validators/go-vet/go-vet.py` is deliberately **not** in this diff.


fdaviddpt:
Second commit, and it corrects a claim I made out loud before pushing the first.

## The macOS legs never ran `go`, and that is the coverage fact worth recording

The three new tests went red on all four macOS legs. `main()` calls `shutil.which(BINARY)` at `go-vet.py:197` **before** the spawn; the payload builder stubbed `subprocess` and `time` but not the lookup, so on a toolchain-free runner the adapter took the `absent()` arm and the tests asserted on a payload that cannot exist there. Reproduced locally by stripping `/usr/local/bin` from `PATH`, down to CI's own `duration_ms` of `60137` — the faked clock produced that number and nothing ever spawned.

**I said "only the Windows legs have Go". That is wrong: it is only the macOS legs that lack it.** `.github/` has no `setup-go` step at all — `grep` for `setup-go|go-version|golang` returns 0 results across 11 files — so each leg gets whatever its image ships. Ubuntu ships Go; macOS does not. Getting this backwards would suggest go-vet has one leg of coverage when it has ten.

The existing gate is real and its cost is measurable. `needs_go` at `test_validators_go_vet_669.py:57`:

```
go OFF PATH:  17 passed, 10 skipped
go ON  PATH:  27 passed
```

Those ten are every test of what `go vet` actually *says* — the clean verdict, the printf finding, sibling attribution, the receipt cap and its truncation notice, the package that does not type-check. All ten skip on every macOS leg.

## Why the new pins are not gated behind `needs_go`, which is a departure

I asked for the existing mechanism to be copied. It was — for the **one** test that genuinely executes `go`. The other four stub the lookup instead, and that is deliberate rather than an oversight.

The timeout arm's payload is a fact about the adapter's **error routing**, not about whether an image ships Go. The spawn is already stubbed, so no `go` runs either way; the lookup decides only whether the code under test is reachable. Gating on it would leave the regression pin for a Windows-observed defect running on the Windows legs alone — the pin absent from four of five platforms, which is approximately the coverage hole that produced this issue.

It is also the file's own idiom: `test_the_load_failure_prefix_is_recognised_under_either_binary_name` feeds the parser captured Windows output *"rather than by branching on `os.name`, so both shapes are pinned on every runner."*

**The stub is pinned as load-bearing rather than assumed.** `test_the_timeout_arm_is_reachable_without_a_toolchain` empties `PATH` and asserts the arm is still reached — and it went **red on a machine that has Go**, which is the proof the stub does the work. `shutil` is stubbed on the adapter module rather than globally: a `which` that lies for the length of a test is not something to hand to the rest of an xdist worker.

## Distinguishable skips

The one test that really runs `go` now carries the suite's own `needs_go`, so its reason is `go not on PATH`, while `skip_if_stalled`'s is `adapter spent its whole 60s internal budget without reaching a verdict`. Absent toolchain and slow toolchain read as different sentences in the skip list.

## Green in both toolchain states

```
go OFF PATH:  26 passed, 11 skipped   — all 11 named "go not on PATH"
go ON  PATH:  158 passed              — incl. html-check, adapter-timeout-policy, require-validators
```

The four #1461 pins run in **both** columns. That is the change.

Issue #1445, opened 2026-08-12
`gh-pr:N` counts unresolved review threads and renders none of them, so reading a bot's review means dropping to raw `gh api`


fdaviddpt:
Hit live while reviewing PR #1443.

`gh-pr:1443` reports:

```
Unresolved threads: 2 / 2
Reviews: github-code-quality (COMMENTED)
```

and prints neither thread. `## Comments (0)` sits below it, which is true — those are issue-level comments — and reads as though there is nothing to read. The only route to the actual content was:

```
$ gh api repos/OWNER/REPO/pulls/1443/comments --jq '.[] | "=== \\(.path):\\(.line) by \\(.user.login)\\n\\(.body)"'
=== validators/common/pkg_paths.py:87 by github-code-quality[bot]
## Empty except
'except' clause does nothing but pass and there is no explanatory comment.
...
```

Two real findings, on a file-and-line each, invisible to the op that had just told me they existed.

## Why the count without the content is the wrong half

A number that says "there are 2" and a render that shows 0 put the reader in the one state this repo files bugs about: the output is *consistent with* nothing being there. `Comments (0)` is the line most likely to be read, and it is about a different population than `Unresolved threads: 2`.

It also lands on the maintainer's own path. An automated reviewer comments in **threads**, not in issue comments, so every bot review on every PR is currently invisible to `gh-pr` and reachable only by the raw call the guard otherwise refuses.

## The nearest existing surface

`gh-pr:N` itself — it already fetches and counts them. This is a field on an op, not a new op. Shape worth considering, since a long bot review should not swamp the PR body:

- `gh-pr:N` keeps the count and adds a one-line-per-thread index: `path:line  author  first line of body`.
- `gh-pr:N:threads` (or `:full` carrying them) renders the bodies.

Whatever the split, the count and the render must not be able to disagree: if the bodies could not be fetched, that is a `could not read`, not a zero.

## Not established

- Whether `gl-mr:N` has the same gap for GitLab discussion notes. Worth checking in the same pass — #628 exists because the two sides drift.
- Whether resolved-but-present threads should appear at all. My instinct is that the index should show them dimmed with the count stated separately, because "resolved" is somebody's decision and a reader may disagree with it.

Issue #1411, opened 2026-08-11
`gh-since-tag` builds its own `gh pr list` argv instead of calling `gh-prs` — the listing is duplicated, and the listing is exactly what drifted last time


fdaviddpt:
Florian, 2026-08-11: *"I really feel `gh-since-tag` should be `gh-list` with a filter."* Checked, and the listing half of that is right — with one boundary worth naming.

## The duplication is real and it is verbatim

`presets/github/since_tag.py:593` builds its own argv:

```python
argv = (["gh", "pr", "list", "--limit", str(limit), "--state", "merged",
         "--search", f"merged:>{stamp}",
         "--json", "number,title,mergedAt,url"] + _repo_target.gh_args())
```

`presets/github/prs.py` has `_build_list_cmd(filters, per_page)` whose docstring opens **"No role filter is ever added here"** — a sentence that exists because a default `author=@me` in that function was read past for an unknown number of sessions (#1207), and then inherited by `radar`'s tier and fixed a second time (#1230). That is two ops and one board drifting off one listing.

`since_tag` is a third caller of the same GitHub endpoint that does not go through it. Nothing today makes it wrong — it deliberately adds no role filter — but the defect class here is not "it is wrong", it is "there is a second place it can become wrong, and the first place has a filed history of exactly that."

`gh-prs` already accepts `state=merged`. What it has no vocabulary for is the boundary: **there is no `since=` / `merged-since=` filter**, which is the one thing `since_tag` needs and the reason it went its own way.

## Where I do not think it collapses all the way to a filter

`gh-since-tag` is not only a listing. Its own description names two things a `gh-prs` filter cannot be asked to carry:

- **Boundary resolution** — picking the newest version-shaped tag reachable from the default branch, at its *commit* instant rather than the tag object's date, in four states (`RESOLVED` / `AMBIGUOUS` / `UNRESOLVED`, with the count rendered `?` and never `0`).
- **The cross-check.** It prints merged-PR count beside unreleased `changelog.d` fragment count *because their disagreement is the finding* — #1209 exists because a hand-rolled version printed a confident `0` beside 7 fragments, having compared `...Z` against `...+02:00` as strings. A zero beside a non-zero fragment count renders as `CONTRADICTION`, not as two numbers. That is a judgement, and the fragment count is a local filesystem read that has no business being a PR-list filter.

So the split I would propose:

[...]

Issue #1230, opened 2026-08-09
radar's GitHub tier still defaults to author=@me — the narrow default #1207 removed from gh-prs survives in the board op a tick opens with


fdaviddpt:
## The defect

`radar:--state`, read-only, run 2026-08-09:

```
gh-prs:
  module    : presets/watch/tiers/gh_prs.py
  filter    : author=@me (default)
```

Bare `gh-prs` no longer does this. #1207 changed it — `presets/github/prs.py:27`:

> The default is the whole repo, and the board says which population it is. It used to be `author=@me`.

So the tier did not *inherit* the narrow default; it **kept it after the op underneath dropped it**.

## Why this is the same defect, not a smaller one

#1071's evidence transfers unchanged: two dependabot PRs (green, 5 hours old) and one external contributor's PR (green, 1 day old) rendered as an empty board, and were found by a human opening GitHub in a browser.

`radar` is the op the maintainer loop opens with — the skill's own words are that it *"answers a tick's opening question in one call: every open PR with its check rollup, plus whether the default branch is actually green."* A board that silently excludes everyone else's PRs is not that.

`gh-prs` was allowed to be personal-by-default for a long time on the argument that the footer disclosed it. #1207 rejected that argument: disclosure is not sufficient when the filtered board and the empty board render almost identically. That reasoning applies to `radar` with more force, because radar's output is a **delta** — a PR that was never in the population cannot appear as a change, so there is not even a row to read the footer next to.

## Measured tonight

`radar` reported `0 open | 0 watched` with `scope author=@me (default)` in the footer. That happened to be true — the repo genuinely had 0 open PRs, confirmed by `gh-prs:anyauthor,state=open,per=100`. **The footer is the only thing that distinguished a true zero from a filtered one**, which is exactly the condition #1207 was closed to remove.

## Shape

Make the GitHub tier's default match the op's: the whole repo. Keep `author=@me` available as a configured tier filter for anyone who wants a personal radar, via `ops.radar.radar_tiers`.

## Judgment call

Whether this is a defect in the tier or in **tier defaults generally** — does any other tier carry a filter its underlying op has since changed? `gl-mrs` still defaults to `author=@me` (`presets/gitlab/mrs.py:153`), so the GitLab tier matching that is currently correct and would become wrong the day `gl-mrs` changes. If the tier is copying a default rather than deferring to the op, that coupling is the actual bug and fixing one tier leaves it in place.

Issue #1209, opened 2026-08-09
No op answers 'what merged since the last tag', so the release trigger is hand-rolled every tick — and mine just returned a confident zero


fdaviddpt:
The auto-release gate is defined in terms of two numbers that must be reported every tick: **merged PRs since the last tag** and **unreleased changelog fragments**. There is no op for either. I compute them by hand, every tick, and tonight the hand-rolled version was wrong.

## What happened

```
=== merged since v0.31.0 ===
tag v0.31.0 = 39372ab at 2026-08-09T17:13:43+02:00
merged since tag: 0
=== unreleased fragments ===
       7
```

Zero merges, seven fragments. The two contradict each other, which is the only reason I looked twice — the correct answer is **6**.

The cause: `gh pr list --json mergedAt` returns UTC with a `Z` suffix (`2026-08-09T16:07:45Z`), while `git show -s --format=%cI` returns a local offset (`2026-08-09T17:13:43+02:00`). I compared them as **strings**. `"16:07:45Z" > "17:13:43+02:00"` is `False` at the second character, so every PR merged after the tag was filtered out as being before it.

Fixed by parsing both with `datetime.fromisoformat` — but that is a fix to a script I retype from memory each time, not to anything durable.

## Why an op

This is the chronic-friction trigger, not the acute one. Nothing broke, nothing errored, and I have been doing this by hand for days without it registering as cost:

- It is required **every tick** by the release gate, so it is the definition of a recurring cost.
- Hand-rolling a projection is exactly how a field gets omitted or a comparison gets it backwards. An op cannot forget the timezone.
- **A zero here is indistinguishable from "nothing merged"** — the house defect, in the number that decides whether a release goes out. A wrong low number silently delays a release; a wrong high number triggers one early.

## Shape

Something like `gh-since-tag[:TAG]`, defaulting to the most recent tag, rendering:

- the tag, its sha, and its timestamp — so the boundary being used is visible rather than assumed
- each merged PR since it, in merge order, with the issues it closed
- the count, stated as the trigger input it is
- the unreleased fragment count alongside it, since the gate reads both together and I currently get them from two unrelated commands
- **three states:** if the tag cannot be resolved, or the PR list is capped, say so rather than rendering a short list as the answer. A capped page here reads as "fewer merges than there are", which is the same failure one layer along.

[...]

PR #1473, opened 2026-08-12
The release gate is a filter now - and the boundary was a clock where it had to be a commit

### LINKED ISSUE #1405: `gh-since-tag` reports UNVERIFIED on every release, because the release PR's merge commit IS the tagged commit
Ten minutes after tagging v0.34.0:

    merged since tag: 6 (UNVERIFIED)  [UNVERIFIED]

    RECONCILE: the search index and local git history disagree.
      in the API, absent from local history: #1403

#1403 **is** the release PR. Its merge commit `a6c3a8b` is the commit the tag points at, so it is not in `v0.34.0..master` — correctly, by definition. The search index returns it because it merged before the tag existed and the index has no notion of the boundary; local history excludes it because the boundary *is* that commit.

So the op is right that the two sources disagree, and right to refuse rather than assert — that part is the three-state contract working. What it cannot do is tell **this** disagreement, which is structural and will happen at every single release, from a genuine one where a merged PR is missing from history.

The cost is small and certain: the first `gh-since-tag` after any tag reports `UNVERIFIED`, and the count is off by one (it said 6; five PRs merged after the tag). A number that is always wrong in the same way at the same moment is the number a reader stops checking.

- **Exclude the tagged commit itself from the "absent from local history" reconciliation**, since `A..B` is exclusive of `A` by construction. A PR whose merge commit *is* the boundary is accounted for, not missing.
- **Keep the refusal for every other disagreement.** The point of this render is that a PR present in the index and absent from history is usually a real problem — a force-push, a wrong default branch, a stale local ref. Narrowing must not turn that into silence.
- Worth checking whether the same reasoning applies to a **squash-merged PR whose commit was later amended**, which would land in the same bucket for a different reason and does deserve the refusal.

Found by running the op immediately after cutting v0.34.0 — the render named the exact PR, which is why this took one read rather than an investigation.

fdaviddpt:
Closes #1405

`gh-since-tag` folds into `gh-prs:merged-since=TAG`. Florian's call, twice — the first time it produced #1411 (shared parser, shared query, op left standing), this is the rest of it.

## The measurement that makes a tag name the mechanism rather than a convenience

`merged-since=` could not express a second-precision boundary **at all**. The `:` tokenizer rejects an ISO instant, so the only typable spelling was a bare date:

```
merged:>2026-08-11T00:00:00+00:00  -> 75 PRs
merged:>2026-08-11T18:57:19+00:00  -> 20 PRs
```

**75 where the answer is 20**, silently, in the direction that fires a release. A tag name contains no `:`, so it is the only colon-free spelling of the boundary. The friction that started this — hand-resolving a tag to an instant, then watching the op refuse the instant — was the op telling you it structurally could not accept the thing it needed.

## #1405's own mechanism: the issue's headline is wrong

Not a boundary-rule disagreement. A **one-second clock skew firing on a coin flip** — GitHub stamps `merged_at` after writing the merge commit:

| tag | PR | tagged commit | `merged_at` | |
|---|---|---|---|---|
| v0.30.0 | #1161 | `06:29:26Z` | `06:29:26Z` | equal |
| v0.31.0 | #1198 | `15:13:43Z` | `15:13:43Z` | equal |
| v0.32.0 | #1250 | `22:35:33Z` | `22:35:34Z` | **+1s** |
| v0.33.0 | #1289 | `00:39:57Z` | `00:39:57Z` | equal |
| v0.34.0 | #1403 | `13:34:34Z` | `13:34:35Z` | **+1s** |

Two in five, not every release. `merge_commit_sha` equalled the tagged sha in **all five**, so the boundary is applied as identity on `mergeCommit.oid`:

```
  · boundary PR: #1403 merged AS the tagged commit a6c3a8b — inside the release, not
    after it. Excluded by identity rather than by clock … (#1405)
  · cross-check: RAN and AGREED — 39 PR reference(s) in v0.34.0..refs/remotes/origin/master
merged since tag: 39  [EXACT]
```

`tests/test_github_since_tag_1209.py:61` pinned `mergedAt == TAG_LOCAL` **exactly** — the 3-in-5 case that passes. A test shaped by the implementation rather than by the defect, which is why this shipped. Replaced.

## The three blockers I named before building — closed vs relocated

| | |
|---|---|
| **The reconcile** | **Closed.** Runs only under a tag boundary; declines by name otherwise. And the agent found the version I did not ask about: its own first cut hardcoded `("author","assignee","reviewer","label")`, so a key added to `_FILTER_KEYS` and forgotten there would make the cross-check report `RAN and AGREED` about two populations that were never comparable — the house defect, inside the guard written against it. Now `_NARROWING_KEYS = _FILTER_KEYS - _NON_NARROWING_KEYS`, pinned by a test on the complement. |
| **Fragments + CONTRADICTION** | **Closed.** Both in the footer. `test_the_contradiction_survives_the_fold` is the test that would have caught the loss — that render is the only reason #1209 was ever found. |
| **The field set** | **Expressible, but only by changing the render**, and this is the thing I asked to hear before it was built. The board's columns are fed by `statusCheckRollup` / `reviewDecision` / `mergeable`; keeping the narrow field set *and* the board render is impossible — the columns would render as fabricated passes. So the boundary slice gets its own render: merge order and merge instants, which the board never carried. On a merged PR the rollup is historical, approval is spent, `mergeable` is null. `failed` beside `merged-since=` is now **refused** rather than answered from a field nobody fetched. |

**Relocated, not closed, and it belongs in the record:** the `changelog.d` read and the `repo:` refusal did not become unnecessary. They moved onto the most-called op in the repo. Narrower than before — only `merged-since=<tag>` triggers the target refusal, a date follows a `repo:` target whole — but the hazard now lives on `gh-prs`.

## What the fold loses

- **The default boundary, permanently.** Bare `gh-since-tag` chose the newest version-shaped reachable tag. A filter value cannot: that choice has three outcomes and a listing has nowhere to carry *"two boundaries are defensible here"*. The tag must be named. This is the irreducible loss.
- **`AMBIGUOUS` stops printing a count** — it refuses. Right for an op owning its render; for a filter, a real board under a warning is a real answer to a question nobody asked.
- **`gh-prs:merged-since=<date>` renders differently for existing callers**: merge-order list plus `release gate: NOT APPLICABLE`, not the triage board.

## Every conditional read says whether it ran

`gh-prs` is what `radar`'s tier calls and what every tick opens with. It was one `gh` call; it now conditionally does a `git tag` read, a `git log` and a `changelog.d` scan. The failure mode would be a narrowed population rendering like a complete one — which `gh-prs` has already had once (#1207, inherited by radar, paid for twice in #1230).

So each read states its own state on the boundary slice: the reconcile `RAN and AGREED` / `RAN and DISAGREED` / `DID NOT RUN` with the reason; the fragment count read or not; the boundary resolved or refused between named candidates. A caller who passes no `merged-since=` sees exactly today's behaviour.

## Review

Sonnet, read-only, against the committed diff. Two load-bearing findings, both real, both accepted:

- **The exit code silently degraded to always-0.** `gh-since-tag` exited 0 only on RESOLVED+EXACT; a board always exits 0. `merged since tag: 5 (UNVERIFIED)` would have shipped the strongest available "go" beside a number the tool had just refused to stand behind. Restored as `gate_exit`, tag slice only — verified live: narrowed → 1, clean gate → 0, ordinary board → 0.
- **`presets/github.json`'s `gh-prs` description still read *"It takes an ISO date or instant, never a tag name"*** — the exact sentence this change falsifies, shipped to every plugin user in the registry. Five prose surfaces were updated and the one that ships was skimmed past.

Also fixed, unasked: `README.md:46` claimed *"`radar`'s GitHub tier still narrows to `author=@me`"*. False since #1230 — the same inverted-capability shape the JIT file records for this exact row.

## Rebase, by me, and what it touched

Branched before #1460 landed; both change `presets/github/pr.py` and `presets/github.json` heavily. Rebased onto `2acd666` with two conflicts, both resolved by **keeping both sides**:

- `docs/presets/github.md` — the `gh-pr` row is #1460's verbatim; the `gh-prs` row is #1460's with #1405's tag-boundary sentence appended.
- `presets/github.json` — the `gh-prs` description is #1405's with #1460's `('threads?' when the per-PR thread call declined — never counted as zero, #1445)` clause spliced back in. Verified by diffing the two strings rather than by reading them: that clause was the *only* HEAD-unique content.

178 tests green across both features' suites after the resolution, including #1460's two thread files.

## Tests

Three red stages, each proved separately — 8 failed / 2 passed for the identity fix, 24 failed / 3 passed for the fold, then the derived-narrowing set. The 3 passing at stage 2 are the acceptance bar: board argv unchanged, `_LIST_FIELDS` still has no `mergeCommit`. Correct before and after.

Full suite before the rebase: `2 failed, 11516 passed, 93 skipped`, both failures reproducing on `f045b6a` in a clean worktree.

**Windows unverified.** No new path-separator logic, no POSIX literals in assertions, spawn already routed through `_run`'s `FileNotFoundError` arm (#997), no new exception catching.

## One decision for the maintainer

`gh-since-tag` survives as a **retirement stub** — `syntax: "gh-since-tag -- RETIRED, use gh-prs:merged-since=TAG,state=merged"` — rather than being removed from the registry. That is a deprecation, not a deletion: the name still resolves and teaches its substitute instead of answering `unknown operation`. If the intent was removal, the stub comes out in a follow-up and `_OP_SYNONYMS` is the existing mechanism for the pointer.


fdaviddpt:
### Code review

Found 4 issues, all stale prose or a dead reference — no correctness findings. Four of five review lenses (CLAUDE.md adherence, shallow bug scan, git history, prior-PR comments) returned nothing; the git-history lens specifically confirmed neither `_build_list_cmd` drift, #1207 or #1230, is reintroduced.

1. `**Instants only. A tag name is deliberately not accepted.**` is now false in the file that implements the opposite, and it is the argument against what this PR builds

https://github.com/Digital-Process-Tools/claude-supertool/blob/bd3ca9ab6d7458ac126c0fae3e5716097487f7f0/presets/github/prs.py#L117-L126

2. Same block: `since_tag.repo_target_refusal` names a module this PR renamed to `_release_gate.py`, and "the state `gh-since-tag` exists to report" describes a deleted op

https://github.com/Digital-Process-Tools/claude-supertool/blob/bd3ca9ab6d7458ac126c0fae3e5716097487f7f0/presets/github/prs.py#L120-L126

3. `_build_list_cmd`'s docstring is stale about its own callers — there is no `gh-since-tag` to call it; the second caller is now a second call site in this same file. That paragraph exists because this function drifted twice, so being wrong about who calls it is the failure mode it was written to prevent

https://github.com/Digital-Process-Tools/claude-supertool/blob/bd3ca9ab6d7458ac126c0fae3e5716097487f7f0/presets/github/prs.py#L311-L317

4. `docs/presets/github.md` lost the release gate's exit-code contract in the `@@ -1778,7 +1782,53 @@` rewrite and nothing restated it — `supertool 'grep:Exit code is:docs/presets/github.md:5:1'` returns `0 results ... scanned 1 files`. The identical contract is documented in code, so the one fact a release script gates on went from documented to undocumented in the doc most likely to be read for "should a release fire"

https://github.com/Digital-Process-Tools/claude-supertool/blob/bd3ca9ab6d7458ac126c0fae3e5716097487f7f0/presets/github/_release_gate.py#L923-L927

Also, and noted because the reason it was set aside is wrong: `import prs  # noqa: E402  (the one gh pr list argv — see read_merged_prs)` at `_release_gate.py:101` cites a function this PR deletes, and `prs.` appears in that file exactly once afterwards — inside a comment. A lens declined to report it as "CI would catch F401". CI is green here, so CI did not catch it.

https://github.com/Digital-Process-Tools/claude-supertool/blob/bd3ca9ab6d7458ac126c0fae3e5716097487f7f0/presets/github/_release_gate.py#L100-L104

Issue #1347, opened 2026-08-11
supertool should own the raw-command block: each op declares what it replaces, and one shipped PreToolUse hook enforces it


fdaviddpt:
Florian, 2026-08-11: *"we should do a prehook that auto-block all raw variant, it should be automatic and belong to supertool. Each supertool op can list what it replaces, array of raw op, and then on PreToolUse we parse, check if it is replaced by something else and prevent the use with the description of the op as a response."*

Today the enforcement is hand-written markdown rules under `.claude/jit-context/tools/00-manual/`, each with its own regex, each maintained separately from the op it protects. That layer is failing in both directions at once, measured this morning.

## The measurement that decided it

**The same rule blocked and did not block the same command, minutes apart, in one session.** `gh-pr-view-merge-have-ops.md`, `mode: block`, matching `~(^|[;&|\n] *)(rtk +)?gh[[:space:]]+pr[[:space:]]+(view|merge|create)`:

    09:03  gh pr view 1321 --json number -q .number    -> RAN. printed 1321. no block.
    09:03  (same command inside a `for` loop)          -> RAN. no block.
    09:04  gh pr merge --help                          -> BLOCKED
    09:05  gh pr view 1321 --json number -q .number    -> BLOCKED

Identical text, identical session, opposite outcomes. Whatever the mechanism — a timeout, a cache, an index read that races — **a gate that intermittently does not run is indistinguishable from a command that complies**, which is this repo's own defect class sitting inside the thing meant to enforce its conventions. It let seven raw `gh pr view` calls through in one tick before anyone noticed, including two whose output went into this tracker.

Two more, already filed, both from the same layer:

- **#1336** — `gh-list-limit.md`'s `require: --limit` refuses commands that carry `--limit`, whenever a double-quoted argument precedes it. Four-command discriminator in that issue.
- **#1221** — `supertool-no-cut.md` fires on the string `supertool` anywhere in the command, which for this repo is usually the **directory name**. Six false positives across four agents on 2026-08-11 alone, including `pytest --no-header` (the substring `head`) and a plain `git worktree add`.

So: one rule that under-fires non-deterministically, one that refuses compliant input, one that fires on commands it was not written for. Three hand-written regexes, three different failure modes, and `CLAUDE.md` already states the consequence — *"A blocking rule that fires on commands it was not written for teaches people to route around the block."*

## The proposal

[...]

PR #1608, opened 2026-08-13
The guard forwarded a dying rung's half-written envelope as a verdict, and the Windows disclosure named only one of the two dead hooks (#1377, #1401)

### LINKED ISSUE #1377: The #1347 guard costs ~0.15s of wall clock on every Bash call, for every plugin user, and nothing is cached
Measured on the #1347 branch by the implementing agent: **0.15s wall for the full wrapper**, per Bash tool call. Three costs, all paid every time:

1. an interpreter probe that **executes** each candidate python to find a working one
2. a fresh import of `_supertool.py` — 22.7k lines
3. a full registry walk to build the replaces table

None of it is cached between calls, and a PreToolUse hook runs on **every** Bash invocation, including the ones the guard has nothing to say about. This is a permanent tax shipped to every plugin user, not a cost this repository pays alone.

For scale, the maintainer skill already records why this matters more than the raw number suggests: a helper that spawns 27 subprocesses is ~300ms and invisible where a spawn costs 10ms, and seconds where one costs 150–800ms — under QEMU, under an antivirus filter driver, on hardware nobody on the team has. **The platform the code is written on is the one that cannot see its own constraint**, and macOS is where this was measured.

## What is worth deciding

- **Cache the interpreter probe.** Executing candidates to find a working python is the most expensive of the three and its answer changes approximately never.
- **A cheap pre-filter before the import.** Most Bash calls contain no command any op replaces. A string test over the replaces table's command words, done before `_supertool` is imported, would skip the two expensive steps entirely for those — at the cost of a second matcher whose disagreement with the real one would be silent. Whether that trade is acceptable is the real question, and this repo's history says a second copy of a matching rule is exactly how a gate starts lying.
- **Whether a slow guard is worse than no guard.** A hook that hangs inside a command reads as a slow network, not as a bug report — the failure mode named in the maintainer skill under making failures announce themselves.

Not fixed in #1347's PR: it is a caching design with its own failure modes and deserves its own review.

### LINKED ISSUE #1401: session-start.sh is not tool-gated, so a native Windows host with no Git for Windows loses the symlink and the op roster every session

fdaviddpt:
Closes #1401
Closes #1377

Two changes in `hooks/`, bundled because the second is a release blocker found in the same files.

## #1377 — a rung that dies mid-envelope has its fragment forwarded, and the wrapper exits 0

Found by the v0.41.0 round-1 security audit, and **introduced by #1377 itself** an hour earlier. `attempt()` was rewritten to identify a working Python by prefix-matching stdout against `{"hookSpecificOutput"`; `rc` was still captured but only fed the decline message. So an interpreter that writes a truncated envelope and then dies — OOM kill, SIGKILL mid-`sys.stdout.write`, a launcher wrapping the child — had its partial JSON printed verbatim and the wrapper exited 0.

Claude Code cannot parse that, so there is **no decision and no `additionalContext`** — the silent fail-open the file's own docstring exists to prevent: *"failing open silently is the exact defect #1347 opens with."*

Reproduced with a shim printing `{"hookSpecificOutput"` then `exit 7`; `json.loads` fails at char 21.

**The fix, and why it is not simply the pre-#1377 shape returning:** forward an envelope only from a rung that also exited 0. Every `main()` return in `hooks/pre_bash_guard.py` is `0` (lines 269, 275, 288, 296, 303, 318, 324, 336), so a rung that wrote a *whole* envelope exited 0 by construction. A failing rung is still only recorded and the walk still continues — pinned by two rows, `..._still_falls_through_to_the_next_one` and `..._writes_nothing_and_dies_still_falls_through`. That is what #1377 deleted the exit check for, and it stays deleted in that direction.

The cost, stated: a whole envelope followed by a non-zero exit for an unrelated reason has its answer dropped. Bash cannot distinguish that from a fragment, and dropping it buys a **disclosed decline** where forwarding it buys silence. The decline text now separates "began writing a verdict and then exited N" from "exited N without writing a verdict".

## #1401 — accept the gap and complete the disclosure

The issue named two repairs. **The Python-with-exec-form rewrite is not implementable**, and that is a refutation rather than a preference:

- Exec form resolves `command` on `PATH` — the `CreateProcess` search #1401 was mis-filed about — and `tests/test_hook_interpreter_windows_1401_1402.py` already refuses `args`.
- It would still have to name an interpreter, and none is portable: `python` / `python3` are the App Execution Alias stubs #572 banned for blocking rather than erroring; `python3.9`–`python3.14` are absent on a standard Windows install (#1402); `py -3` is absent everywhere else.
- Resolving that is exactly what `hooks/python-ladder.sh` is for — **and it is a shell script, so it cannot bootstrap a shell-less host.**

So this ships the second option properly: the disclosure existed for one of the two dead hooks and pointed at itself as though that were the whole cost. It now names `session-start.sh` as well, says what the session loses, and says the way back.

## A load-bearing claim in #1401's own refutation is false, and it was pinned by a test

The reviewer caught this in the new prose; checking it showed the sentence was **inherited, not invented**.

`hooks/hooks.json:5` is `"matcher": "Bash|PowerShell"`, and `hooks/pre_bash_guard.py:28-34` records why — #1413 widened it because Claude treats PowerShell as the primary shell wherever that tool is enabled. So on a native Windows host with no Git for Windows the guard hook **is** matched, fires, and cannot run. Inert, which is #1378's silent state — not un-asked.

The same claim sits in `docs/configuration.md:251` and in `tests/test_hook_interpreter_windows_1401_1402.py:23-27`, pinned by #1412, where it composes into *"the only host where the hook string reaches PowerShell … is exactly the host with no Bash tool, so a PreToolUse hook matched on Bash never fires there at all."* **That composition is false**, and it is the load-bearing step in #1401's refutation of the guard half — including in the issue comment itself.

What survives: `args` is still the wrong repair, and #1401's severity claim is still wrong, because only exit 2 blocks a `PreToolUse` call and the WSL launcher exits 1, so the command proceeds. What does not survive: "the branch is empty" — it is **bounded**, not empty. All three sites corrected in place; no assertion changed, docstrings and prose only.

The distinction that actually carries #1401 is **silent failure versus visible failure**, not asked versus unasked, and the shipped text now says that.

## Tests

Red first, both files:

```
3 failed, 1 passed in 5.78s
FAILED ...::test_the_no_bash_state_names_the_session_start_hook_as_well
FAILED ...::test_the_no_bash_state_names_what_the_session_loses_and_the_way_back
```

```
E  json.decoder.JSONDecodeError: Expecting ':' delimiter: line 1 column 22 (char 21)
FAILED ...::test_a_fragment_of_an_envelope_is_never_forwarded_as_a_verdict
3 failed, 1 passed in 3.05s
```

Green: `4 passed`, and `59 passed` across the five neighbouring hook and guard files. **Full suite run twice** (README and docs are read by meta-tests): `12703 passed, 97 skipped, 2 subtests passed in 320.10s`.

## Windows evidence grade, per claim

| Claim | Grade |
| --- | --- |
| Both `hooks.json` entries name `bash`; `SessionStart` carries no matcher | **observed** — file on disk |
| A bare `bash` under `CreateProcess` on a Windows runner is the WSL launcher | **observed** — this repo's own pytest, PR #1399 |
| Every `main()` path in `pre_bash_guard.py` returns 0 | **observed** |
| A partial envelope is forwarded verbatim and the wrapper exits 0 | **observed** — reproduced on macOS with a shim |
| Shell form reaches PowerShell where Git Bash is absent | **reasoned** — Claude Code docs |
| The `Bash` tool is absent on that host | **reasoned** — the docs are not explicit |
| `SessionStart` there loses the wrapper and the op roster | **reasoned** |


fdaviddpt:
## Third commit — the fixture was wrong, and fixing it exposed a second product defect

**Why the first version went red on all four ubuntu legs and passed on macOS and Windows.** The test installed the dying shim at the ladder's *first* rung and left the rest of the ladder alone. `hooks/python-ladder.sh` then walked on:

| | `/usr/bin` contents relevant to the ladder |
| --- | --- |
| the author's Mac | `python3` only — the bare name, **never a rung** (#572) |
| ubuntu runner | real `python3.10` / `python3.11` |

So on ubuntu the walk stepped past the dying shim, found a genuine interpreter, and returned a genuine `deny`. The assertion was measuring the real guard wearing the shim's name. Reproduced locally by standing a versioned interpreter after the shim directory on `PATH` — byte-for-byte the CI envelope, not inferred.

`tests/test_hook_interpreter_windows_1401_1402.py` already carried `_blind_the_whole_ladder`, whose docstring says exactly this: *"A rung left unshadowed is the host's own interpreter answering the ladder, and the test then measures the runner rather than the code."* The helper existed, documented, and was not reached for.

**The fixture now shadows every rung with `exit 9` and overwrites only what each row is about**, with two additions beyond the existing helper:

- **The rung list is parsed out of `hooks/python-ladder.sh`** rather than copied. The existing helper hardcodes `_VERSIONED`, so a seventh rung added to the ladder would silently un-blind it on any host carrying that interpreter — the same failure one turn later. The parse refuses to return nothing rather than blinding zero names and reporting green.
- **A control row that asserts the blinding itself** — with every rung exiting 9 the wrapper must decline and never decide. Without it, a renamed rung or a non-executable shim puts the other rows quietly back to measuring the runner image.

## The second defect, which only a correct fixture could see

With six rungs tried *after* the dying one, the decline named whichever rung was last:

```
supertool raw-command guard did not run: py -3 exited 9 without writing a verdict.
```

That buries the only actionable fact — `python3.14` began writing a verdict and exited 7 — under a rung that was never going to answer, and sends the reader hunting a silent interpreter instead of a crashing one. `PARTIAL_TRIED` / `PARTIAL_RC` are now sticky and reported in preference to the generic last-rung failure.

The first version of the test could not have caught this: with one rung installed there was no later rung to overwrite the message. **The bad fixture was hiding a real defect, not merely failing on CI.**

## The coverage leg was a consequence, not a floor

From job #94496301714's own log:

```
coverage gate: the suite failed — coverage of a red suite is not a number anyone should act on.
```

No floor missed, no file short — it declined rather than reporting a number, which is the three-state contract working. It goes green when the four pytest legs do.

Local: `5 passed` targeted, `12704 passed, 97 skipped, 2 subtests passed in 253.07s` full. The `py` rung is shadowed for Windows for the same reason. Local green remains the weakest evidence about a platform nobody here runs — but the specific ubuntu shape is now reproduced and asserted against rather than assumed away.

Issue #259, opened 2026-05-29
edit/paste/replace silently drop the file's executable bit (atomic write loses mode)


fdaviddpt:
## Summary
Mutating ops (`edit`, `paste`, `replace`, `replace_lines`, `vim`) silently drop the target file's permission bits. An executable script (`0755`) becomes `0644` after editing, breaking it. No error is surfaced — the next invocation just fails with `permission denied`.

## Repro
1. `chmod +x cb.sh` (mode `0755`)
2. `./supertool 'edit:@-' <<< '{"path":"cb.sh","old":"X","new":"Y"}'`
3. `./cb.sh` → `permission denied` (file is now `0644`)

`bash -n` / syntax validators still pass — only the mode is lost, so it's easy to miss until the script is run.

## Cause
The atomic-write path creates a temp file (default umask → `0644`) and renames it over the original. The original file's mode is never copied to the temp file, so the executable bit (and any non-default mode) is lost. This affects every op that shares the atomic-write helper.

## Fix
Copy the original mode onto the temp file before the rename:
```python
import shutil
if os.path.exists(path):
    shutil.copymode(path, tmp_path)
```
(or capture `os.stat(path).st_mode` before writing and `os.chmod(tmp_path, mode)` after). For `paste` creating a brand-new file, keep the default mode.

## Impact
Silently broke `cb.sh` in real use, and would break git hooks, any `*.sh`, or other executables edited through supertool.

## Env
macOS, supertool invoked via symlink. Validators ran clean (bash-check OK) — only perms were lost.

PR #1667, opened 2026-08-14
paste and append created files at mkstemp's 0600 and never said so; the example config named a variable no runner sets (#1275, #1021)

### LINKED ISSUE #1021: .supertool.example.json tells readers to set SUPERTOOL_DASHBOARD_LANE_PREFIX, which the runner never sets
Found while reviewing #1008 before merging it. The code is correct and pinned by a test; the **example config a user copies** names an environment variable that does nothing.

## What the file says

`.supertool.example.json`, `ops.dashboard._doc`, last sentence:

> Passes through as `SUPERTOOL_DASHBOARD_LANE_PREFIX`.

## What the runner does

`_supertool.py:2085`:

```python
env[f"SUPERTOOL_{k.upper()}"] = v if isinstance(v, str) else json.dumps(v)
```

The key alone is uppercased. It is **not** namespaced by op. So `ops.dashboard.lane_prefix` arrives as `SUPERTOOL_LANE_PREFIX`, which is exactly what the preset reads:

```python
LANE_PREFIX_ENV = "SUPERTOOL_LANE_PREFIX"
```

`test_the_config_key_reaches_the_preset_under_the_name_it_reads` pins the code side. Nothing pins the prose.

## Why the wrong name is plausible

The sibling `_doc` immediately below it is correct — `ops.radar.radar_tiers` really does arrive as `SUPERTOOL_RADAR_TIERS`, but only because the *key* is `radar_tiers`, not because the op name is prepended. Copying that sentence to a key called `lane_prefix` silently changes what it claims.

`ops.git-diff.red_flags_extra` is the third instance of the same route and would read `SUPERTOOL_RED_FLAGS_EXTRA` by the same rule.

## Harm

Misreports class, and it is the silent kind: someone who prefers an env override to a config edit exports `SUPERTOOL_DASHBOARD_LANE_PREFIX`, gets no error, and the lanes section correctly prints `!! unread — no lane vocabulary configured`. The op is right and the reader has been told the op is broken.

## Fix

One sentence in `.supertool.example.json`. Worth pairing with:

- a test asserting every `SUPERTOOL_*` name appearing in a `_doc` in `.supertool.example.json` matches `SUPERTOOL_<KEY.upper()>` for the key it documents — this class is copy-paste-shaped and will recur;
- a check of the other `_doc` entries for the same mistake while in there.

### LINKED ISSUE #1275: paste creates a file 0600, so a script written through it lands non-executable and git records the wrong bit
Hit while an agent wrote a new shell script into a repo where every sibling script is 0755.

## What happens

`paste` creates the file with mode **0600**. Nothing in the receipt says so. Every sibling in that directory was 0755, and git records only the exec bit — so had it not been caught by hand, the commit would have carried a shell script that cannot be executed, with no local symptom until someone ran it.

## Why this is worse than a wrong default

The receipt is otherwise complete: it names the path, the line count and the byte delta. **A mode that is never mentioned is a fact the reader has no reason to check**, and the one thing that would have surfaced it — running the script — is exactly what the next person does after the merge rather than before it.

It also fails asymmetrically. 0600 on a data file is invisible and harmless. 0600 on a script is a broken artefact, and `paste` cannot tell which it just wrote from the content alone. But it can tell something the caller cannot: what the *directory* looks like.

## Directions, none free

1. **Report the mode in the receipt.** Cheapest, honest, and does not guess. It puts the fact where the caller is already reading, and leaves the decision with them.
2. **Match the umask** rather than forcing 0600. The obvious answer, and it makes `paste` behave like every other file-creating tool — but it is a behaviour change for anyone who has come to rely on the tight default, and payload-written files sometimes carry secrets.
3. **Match the neighbours** — if every file in the directory shares a mode, use it. Most likely to be right, and the most surprising when it is wrong.

I lean toward 1 plus 2: disclose always, default to the umask. **Whoever builds this should say what happens on Windows**, where the bit is meaningless and the disclosure line would be noise or a lie depending on wording.

## Adjacent, found in the same session

There is **no op named `write` or `create`**. `paste` is the answer, and an operator who greps the op list for the word they were reaching for gets a clean zero — the absence-read-as-absence shape, in the op-discovery surface. `help:paste` then answers that it *"has no documented help… run 'ops'"*, and `ops` lists it under "Also accepted, no reference" with no signature: two calls to learn nothing. Worth fixing alongside, since both are about `paste` being unfindable rather than about what it does.

fdaviddpt:
Closes #1275
Closes #1021

## #1275 — the root cause was `mkstemp`, not a policy

`_atomic_write` captured the target's mode inside a `try`. On a **create** the `os.stat` raised, `orig_mode` became `None`, the chmod was skipped, and `mkstemp`'s `0600` survived the rename. So the mode was never a decision — it was a temp-file default leaking through, applying only to creates, never to rewrites, undocumented and absent from the receipt.

Now three states: `FileNotFoundError` → `0666 & ~umask`; **any other `OSError` → `None`**, leaving mkstemp's mode, because a mode that could not be read must not be replaced by a create's default; otherwise the target's own mode, so #259 is intact.

**The widening is real and bounded, and the fragment says so rather than hiding it.** On `umask 022` this is group/world-readable where it was owner-only, and a payload can hold a token. Three bounds: the mask is the caller's own declared policy and is now honoured instead of overridden; the mode is disclosed on every create; overwrites are untouched, so an existing `0600` file stays `0600`. `umask 077` is the spelling for owner-only, and it now works for creates too — which it did not before.

**`paste` infers nothing.** No shebang sniffing, no neighbour-matching. The receipt states the file is not executable and names `chmod +x`; the mode on disk is identical either way, so a wrong read of the content costs one line of prose rather than a permission nobody asked for.

## The issue's second claim was false, and it was mine

I filed #1275 asserting "git records the wrong mode in the index". **It does not.** Git stores only the executable bit, so a `paste`d `0600` file and a `>`-redirected `0644` file are byte-identical in the index — `100644` for both, a *correct* record of a non-executable file. Verified with `git ls-files -s` over one of each in a scratch repo. The defect was the mode on disk and the receipt's silence, and the changelog fragment says that instead of shipping my framing.

## The reviewer found the half I did not ask about

`append` creates through the same chokepoint, so its created files were widened too — and `op_append` never disclosed it. That is precisely the silent case this fix exists to remove, while the README sentence I had just written claimed the widening is never silent. Accepted in full, nothing argued down: `append`'s created-file mode is now disclosed too, red first (`assert 'mode 0644' in 'created …'`), with a companion test that appending to an *existing* file states no mode.

## #1021 — the prose was the bug, not the runner

Three sibling entries document the flat form correctly, every preset reads the flat name, and an existing test pins that the key reaches the preset under the name it reads. Namespacing the runner would have broken all of them, so `SUPERTOOL_DASHBOARD_LANE_PREFIX` → `SUPERTOOL_LANE_PREFIX`.

Pinned as a **rule swept over both config files** rather than as the one sentence, because the defect is copy-paste-shaped. The sweep found exactly one instance. A counter-example drafted into the config prose ("never `SUPERTOOL_DASHBOARD_LANE_PREFIX`") was dropped: the example file is *copied*, not read as a lesson, and leaving the wrong string in it would defeat the sweep guarding it. That explanation lives in the fragment.

## Windows

No mode line and no `chmod` line is printed there — `os.chmod` honours only the read-only flag, so `st_mode` reads back `0o666` whatever was asked, and a mode line would be a false claim about access.

The mode assertions gate on a **probe** — create a file with an explicit `0666` and read it back — not on a platform name, so a filesystem mounted without permission bits skips for its real reason. Skips carry `posix-file-modes(#1275)`. One test asserts the cross-platform fact directly: `("mode 0" in out) == (os.name != "nt")`.

Full suite in a disposable clone: `13084 passed, 97 skipped`.


fdaviddpt:
## Update — the Windows red was the guard, not the fix

CI went red on all four Windows legs and on `lint`. 12429 passed on each leg, so the product was never disarmed; the harness was.

**One cause, three symptoms.** All three failures already called `require_posix_modes()` as their first statement — they were gated, and the gate was wrong. The probe created a file at `0o666` and compared against `0o666 & ~umask`. Windows' umask is `0` and a writable file there reads back `0o666`, so **the value the probe wanted was the platform's unconditional answer**. A probe whose positive result is indistinguishable from the platform's default answers nothing.

**The strongest evidence was a test that did not fail.** `test_a_created_file_lands_at_the_umask_default` *passed* on Windows — possible only if `_umask()` returned 0 and `want == got == 0o666`. A vacuous pass on the platform whose reds are load-bearing, invisible in the log until you look for what is missing from it.

**Why no POSIX run could ever have caught it.** The first attempt to simulate the failure went green: faking only `os.stat → 0o666` is caught by macOS's `umask 022`, because `want` becomes `0o644 != 0o666`. Both halves are load-bearing — umask 0 **and** a constant `0o666`. The regression test now fakes both and says so.

The probe now creates at `0o600` and requires exactly `0o600` back. **Clearing the group and other bits is precisely what a filesystem without permission bits cannot do.** Its docstring carries why `0o666` was the wrong value to test with, and notes that a umask clearing an *owner* bit also lands in the skip — the safe direction, named in the reason string.

**Observed, not reasoned:** a pytest plugin forcing `umask 0` and every regular file to `0o666` — the Windows filesystem, reproduced on this box:

```
SKIPPED [4] posix-file-modes(#1275): a file created with mode 0o600 read back 0o666 --
  this filesystem does not carry POSIX permission bits (a umask of 0o0 would also land here)
8 passed, 4 skipped
```

Forcing `os.name = "nt"` too was abandoned — pathlib raises `UnsupportedOperation: cannot instantiate 'WindowsPath' on your system` at collection. So the filesystem half is **observed** and the `os.name == "nt"` disclosure half is **reasoned** from `_created_mode_note`'s own branch.

## A second defect, found while fixing the first

`require_posix_modes()` answers *does this filesystem carry permission bits*. A receipt assertion depends on `_created_mode_note`'s `os.name == "nt"` branch. **Those coincide on every CI runner and are not the same claim** — a POSIX host on an exFAT or SMB mount discloses the mode while the probe declines, and the receipt test would skip there for a reason that is not its own. Receipt assertions moved to `require_mode_disclosure()`, mirroring the product's actual condition. One test lost its gate entirely: it asserts an absence true on every platform.

Product code is unchanged in this round — the last commit touches only the two test files.

Issue #1239, opened 2026-08-09
gh-issues accepts `per=` and its own signature omits it, so four documents read as wrong because the registry is


fdaviddpt:
Found by `claims` (#1220) on its first real run over this repo's docs. It is the only reason the corpus does not come back `contradicted 0`, and the finding is in the registry rather than in the prose.

## The defect

`per=` is a real filter key — `_FILTER_KEYS` at `presets/github/issues.py:90` — and `issues.py:595` prints it to the user by name:

```
capped at --limit 50 — more may exist, raise with per=N
```

But `presets/github.json`'s `syntax` string for `gh-issues` does not list it. `gh-prs`'s does list its own `per=`.

So anything that validates a documented invocation against the registry concludes that `gh-issues:per=100` is wrong. **Four documents use it, correctly, and all four are reported as contradicted.** The op told them to write it, and the registry says they may not.

## Why this matters beyond four rows

The `syntax` field is **parsed, not just displayed** — `docs/contributing.md` §"Custom ops" and `tests/test_at_file_route.py::TestPayloadRoutePin`, ref #770. For any op whose `syntax` contains `:::` it derives the `@file`/`@payload` field names, so the string is load-bearing rather than documentation. An omission there is not cosmetic; it is the registry disagreeing with the code about what the op accepts.

It also makes the registry unusable as an oracle. A reference checker, a completion, or a person reading `ops` all get the same wrong answer, and the *most* accurate documents are the ones that get flagged.

## Fix

Add `per=` to `gh-issues`' `syntax` in `presets/github.json`, matching how `gh-prs` spells it.

## The judgment call, which is bigger than the one-line fix

**Nothing compares `_FILTER_KEYS` to the `syntax` string.** This is one drift instance of a class: two hand-maintained copies of one contract, which is exactly `docs/validators.md`'s SCHEMA drift problem (#1042) one layer over. Worth deciding whether a test should assert that every key an op accepts appears in its declared syntax, for the whole preset registry — that turns "somebody noticed" into "it cannot ship".

If that test is written, expect it to find more than this one. Report the count rather than fixing silently.

Issue #1293, opened 2026-08-11
A test swept a changelog fragment, so it reddened the release commit on all 13 legs — and would have reddened every release after


fdaviddpt:
Found by the v0.33.0 release itself, which is the only thing that could have found it.

## What happened

`tests/test_ops_roster_1231.py::test_quoted_byte_figures_are_this_checkouts` listed `changelog.d/1231.added.md` among the files it sweeps for two stale byte figures. `assemble_changelog.py` **consumes** fragments at the tag, so the release commit deleted the file the test reads:

    FAILED tests/test_ops_roster_1231.py::test_quoted_byte_figures_are_this_checkouts
      FileNotFoundError: … /changelog.d/1231.added.md

13 of 22 legs, on a failure with nothing to do with the figures it guards.

## Why it is worth an issue rather than just a fix

**A fragment exists only between the PR that writes it and the release that consumes it.** Any test keyed to one is green for exactly that window and red forever after — and the window is invisible from inside the PR that adds it, because the file is there the whole time CI runs.

That is a latent class, not a one-off. Anything referencing `changelog.d/<n>.<section>.md` by name has the same lifetime problem: a doc example, a test fixture, a jit-context citation. Worth one sweep for other references keyed to a fragment path, and worth stating the rule where fragment authors will see it — a fragment is not addressable after the tag.

## The fix taken, and its measurement

Swept `CHANGELOG.md` instead, which is where the fragment's prose lands and is permanent. Measured before swapping: **zero occurrences of either stale figure across the whole assembled history**, so the swap neither weakens the sweep nor manufactures a failure on old entries.

## What this says about the release gate

The release is the only place this could surface, because it is the only operation that deletes fragments. That is an argument for the release rehearsal being a real run rather than a checklist — the version-site sweep and the four guards were all green, and none of them could see this.

Issue #1231, opened 2026-08-09
The SessionStart op listing is truncated below `grep`, so every gh-*, git-* and radar op is invisible at the moment it matters most


fdaviddpt:
## What happened

The SessionStart hook prints the op listing. It is capped, and it says so:

```
⚠ Output is 8766 bytes, exceeds the ~7168-byte SessionStart hook cap.
The tail of this listing will be truncated — ops below the cut-off are hidden.
Run `./supertool 'ops'` to see the full listing.
```

The listing stops at `grep:PATTERN:PATH`. **Everything alphabetically after that is hidden** — the entire `gh-*` family, the entire `git-*` family, `radar`, `watches`, `read`, `paste`, `tree`, `validate`, `workspace`.

The disclosure is honest and the behaviour is still bad, because what gets hidden is *existence*, and a reader cannot miss what they do not know to look for.

## Measured cost, one session, 2026-08-09

| Reached for | Existing op | Cost |
| --- | --- | --- |
| `worktrees` (guessed) | `git-worktrees` | error listing 40 builtins, none the answer; second call |
| raw `gh pr create` ×4 | `gh-pr-create` | skipped the op's closing-reference check at creation — the exact `Closes #948 880` defect it exists to catch |
| hand-rolled cohort jq | `gh-labels:tally=` | the burn-down the maintainer skill orders reported every tick |

Three agents in the same session independently reported `write:` is not an op without being told `paste:` is.

## No listing form fits the cap

```
ops           47,260 bytes
ops-compact    9,073 bytes
cap            ~7,168 bytes
```

`ops-compact` misses by 1,905 bytes, so **truncation is guaranteed today** and will stay guaranteed as ops are added.

## What does fit

A names-only roster: **83 names, 711 bytes** — under a tenth of the cap.

```
append around around_line batch between channel cwd dashboard diag diff edit format …
gh-branch gh-issue gh-issues gh-job gh-labels gh-pr gh-pr-create gh-pr-merge gh-prs …
git-blame git-checkout git-commit git-conflicts git-diff git-push git-status git-worktrees …
```

**A complete list of names beats a truncated list of descriptions**, because every failure above was about existence, not semantics. Names alone would have caught all three.

## Shape

An `ops:names` form (or a flag) that emits the roster and nothing else, for the hook to call. Descriptions stay available on demand.

[...]

PR #1674, opened 2026-08-14
Three registry claims the dispatcher contradicts, and a conflicted config that reports itself as absent (#1245, #1162)

### LINKED ISSUE #1162: A conflicted .supertool.json disables every preset op — including git-conflicts, the one op you need to resolve it
Hit live on 2026-08-09 resolving a real rebase conflict in `.supertool.json` itself (#1140's
`html-check` block against #1132's `changelog-fragment` block, both registering a validator).

Git writes conflict markers into the file. The file is now invalid JSON. Config loading fails, so
**no preset ops load at all**:

```
$ python3 supertool.py 'git-conflicts'
supertool: skipped /Users/.../st-wt/1132/.supertool.json: JSONDecodeError: Expecting property name enclosed in double quotes: line 318 column 1 (char 16595)
--- git-conflicts ---
ERROR: op 'git-conflicts' is unavailable here, not unknown — it is provided by the shipped preset 'git'.
       No .supertool.json was found from ... or any parent, so no preset ops and no project ops are loaded — only the built-ins.
       Fix: run it from a project that enables the 'git' preset, or make this call's first op 'cwd:<project-path>'.
```

Both halves of that message are individually accurate and together they are unhelpful, because the
suggested fixes do not apply:

- **`cwd:<other-project>` does not help.** The git ops act on the working directory, so pointing at
  another project root reports *that* repo's conflicts, not this one's. It would answer a question
  nobody asked — the `cwd:`/`repo:` defect shape (#678) arriving through the remedy line.
- **There is no other project root to move to.** The conflict is in the repo you are standing in.

So the failure mode is circular: `git-conflicts`, `git-resolve` and `git-status` are exactly the ops
for this situation, and the situation is what removes them. The fallback is raw `git diff
--diff-filter=U` plus reading markers by hand, which is what this preset exists to replace.

**The message is also slightly untrue in a way that matters.** It says "No .supertool.json was found
from X or any parent" — but one *was* found, and skipped, and the line above says so. A reader who
sees only the ERROR concludes the file is missing and goes looking for why, when the file is present
and merely unparseable. Two lines of one render disagreeing about whether the config exists.

## What might be right

Not sure, and the judgment is the interesting part rather than the fix:

- **Detect conflict markers specifically.** If the config fails to parse *and* contains
  `<<<<<<< `/`>>>>>>> ` at line starts, say so — "your .supertool.json is mid-conflict; the git
  preset is what resolves that" — rather than reporting it as absent.
- **Load the git preset from the last good config, or from defaults, when the config is unparseable
  for this reason.** Attractive and possibly wrong: silently running ops from a config the user
  cannot see is its own trap, and this repo's whole contract is against guessing.
- **Or just fix the message** and accept the circularity, since the git preset's own ops are the
  narrow case where an unparseable config is the *expected* state rather than an error.

The first and third are cheap. The second needs an argument I do not have.

### LINKED ISSUE #1245: Three registry and doc claims about the op set that the dispatcher contradicts
Collected while building #1231, which had to read the op registry as an oracle and found it disagreeing with the code three ways. Filed together because they share one shape — **a declaration that other code or a reader treats as authoritative, and that nothing compares against reality.**

## 1. `status` is documented as informational and is not

`docs/contributing.md`'s op schema describes `status` as `"experimental"` / `"stable"`, informational only.

The code reads it as **truthy**, and `0` **suppresses the op from every listing**. So the doc is wrong about the type *and* about the effect, and the effect is invisibility — an op that ships, dispatches, and cannot be found. That is #1231's defect arriving through a second door.

## 2. `.supertool.json`'s `builtin-ops` carries two names that dispatch as nothing

`grep-count` and `read-grep` are documentation entries for *forms* of `grep` and `read` — `grep:…:count` and `read:PATH:::grep=`. They are not dispatchable names. Anything enumerating the registry to learn what exists gets two names that do not resolve, and #1231's roster had to exclude them by hand.

Either they belong in the registry as real aliases, or they belong in the docs and not the registry. Right now they are in the place that machines read.

## 3. `docs/operations/index.md` opens with "~40 ops across five categories"

The dispatcher accepts **84** here. Measured 2026-08-09 by enumerating the resolved op set, not by counting headings.

That number is the first sentence a reader meets, and it is roughly half. It is also the kind of claim that was true once — which is what makes it worth a guard rather than an edit.

## The judgment call, which is bigger than the three fixes

**Should any of these be a test rather than a correction?** #1239 is the same class (`gh-issues` accepts `per=` and its declared syntax omits it) and #1241 is the same class one layer up (a guard reading string literals only). A test asserting that every key an op accepts appears in its declared syntax, that every registry name dispatches, and that a documented count matches the resolved set would retire all of them and stop the next one.

If that test is written, **report the count it finds rather than fixing silently** — the interesting number is how many sites are wrong today, and this repo has been surprised by that count twice (#1202's "13 adapters, not 1", #1159's "nine directories, not seven").

#1231 already ships one guard of this family: shipped preset ops must declare `safety`, built-ins must have a class, and the roster must stay under half the cap. That is the precedent to copy.

fdaviddpt:
Closes #1245
Closes #1162

One file and one load path, which is why they are one PR.

## #1162 — the config that disables the ops you need to fix it

A rebase conflict in `.supertool.json` makes the file unparseable, so no preset ops load — including
`git-conflicts`, `git-resolve` and `git-status`, which are the ops for exactly that situation. Both
remedies the old message printed were dead: `cwd:<other-project>` answers about a different
repository, and there is no other root to move to.

Worse, the message **contradicted itself**: it said "No `.supertool.json` was found from X or any
parent" one line under a line saying one *was* found and skipped.

Now `_skipped_config()` re-derives from disk why a present-but-unused config was skipped, and conflict
markers are named with their line numbers. The `cwd:` remedy is not printed on the conflict path,
because it is not a route to anything.

**Direction 2 from the issue — loading the git preset from defaults — is refused, and the fragment
says so** so the decision survives the release. Running ops out of a config the user cannot see is
precisely what `docs/validators.md` §"Declining instead of guessing" forbids. The circularity is real
and it stays; the message stops inventing a way around it.

One pre-existing bug surfaced by this change: `_unknown_op_message` read `_CONFIG_PATH` without
forcing a load, and `None` means both "none loaded" and "none attempted" — so a **good** config was
reported as unloadable. Fixed at the caller and in the helper.

## #1245 — a guard, and the count it found is three

`grep-count`, `read-grep` and — found by enumerating the files rather than reading the issue —
**`grep-no-exclude` in `.supertool.example.json`, the template users copy**, are argument *forms*, not
dispatchable names. Anything enumerating the registry got names resolving to `unknown operation`.

They now carry an explicit `form` key rather than being inferred from the syntax head, and
`_configured_op_names` and `_build_at_file_registry` both skip forms. They are **kept, not deleted**:
each carries `hint: true`, and `ops-compact` keeps the `example` line only for those — the SessionStart
listing is where a payload shape gets discovered, and deleting three examples from a ~7KB budget to
fix a naming problem is the wrong trade.

`docs/contributing.md` described `status` as informational; the code reads it truthy, and a falsy value
**hides the op from every listing while it still dispatches**. `docs/configuration.md` had it right, so
two pages disagreed.

**The issue's third item needed no change, and none was made.** `docs/operations/index.md` already
reads "44 ops across seven categories", and `tests/test_ops_index_complete_1371.py` already ships the
guard the issue proposes, both directions plus a phantom-name check. The issue's "84" is the *resolved*
op set (94 today); the page's 44 is `len(_valid_op_names())`, the right denominator for a builtins
reference. Two correct numbers about different sets, read as one.

**The guard is scoped deliberately.** "Every registry name dispatches" and "a documented count matches
the resolved set" are both held now. "Every key an op accepts appears in its declared `syntax`" (#1239)
was **not** written — it needs a parser for each op's real kwarg surface and is a different PR.

## Test evidence

Nine failures shown red before implementation, verbatim in the worktree note, plus one adjacent finding
proved red separately by removing its five-line guard. Green: 20 passed targeted; 340 passed across the
neighbouring at-file/614/1231/1371/meta/1124/custom-op/1551 files.

**Full suite run, exit code 0** — core paths changed, `_build_at_file_registry` runs on every dispatch.
The agent lost the pass/skip tally by piping its own capture through `tail -25`; the verdict is the
exit status, not a count anyone read. **Windows unverified.**

## Adjacent, fixed in this PR

- `docs/configuration.md:19` — the illustrative `builtin-ops` block shows `read-grep` **without** a
  `form` key, 70 lines above the new paragraph saying every key there is an op name unless it declares
  one. That block is the shape people copy.
- `_AT_FILE_DROPPED_ROUTES` carried `read-grep`, a standing #770 report that "a payload route was
  intended here and discarded" about a name that never was an op. Nothing reads the list at runtime,
  which is why it survived. Pinned by a test proved red by removing the guard.

## Review — and this is the one thing to read before merging

The Sonnet reviewer **produced no substantive finding, and its first round was malformed** (category
headers reading "see findings above" with nothing above them). Round 2 restated as `clean` for both
files. It wrote nothing to the worktree.

So this PR did **not** get a working independent review, and a review that did not execute must not
render as a review that found nothing. **Both adjacent findings above came from the author's own sweep
afterwards.** I am running a second review against this diff before merging rather than treating the
`clean` as evidence.


fdaviddpt:
## Round 2 — the replacement review's findings, and what the fixes turned up

The review that produced these findings was a **replacement**: this PR's own reviewer returned a
malformed pass and then a bare `clean`, so a second reviewer was run against the committed diff. Two
new commits, `928d8226` and `9883b9ac`.

| Finding | Verdict |
| --- | --- |
| `_skipped_config()` uncached, multiplies inside `batch` | **Argued down** — see below |
| A directory named `.supertool.json` still reported as absent | **Fixed**, and dangling symlinks with it |
| Three untested branches, incl. the `\|\|\|\|\|\|\|` diff3 marker | **Fixed** — and the first test disproved this PR's own changelog fragment |
| A comment over-claiming why the generic branch is reached | **Fixed** |
| Two tests pinning config data rather than behaviour | **Kept as data pins**, one synthetic behaviour test added alongside |

### The cache was refused, and the reason is this PR's own subject

A cache would have to be invalidated on `cwd:`, which retargets the working directory mid-call with no
hook to hang an invalidation on. **A cached answer read after a chdir is a statement about the wrong
tree** — #678's shape, and one of the two dead remedies this issue exists to delete from the message.

The cost is also bounded rather than asserted to be small: `_skipped_config` returns above the walk
whenever `_CONFIG_PATH` is set, so it is unreachable once any config has loaded. That bound is now a
test, not a claim:

```python
assert supertool._skipped_config() is None
assert calls == [], f"walked the tree with a config already loaded: {calls}"
```

And if the repetition ever does matter, the fix is to pass the note into `_unknown_op_message` and have
`dispatch` derive it once — written into the docstring so the next reader does not reach for the cache
first.

### The test disproved the fragment that asked for it

The changelog fragment claimed that matching only `<<<<<<<` / `>>>>>>>` would classify a diff3 conflict
as ordinary broken JSON. **That is false** — `merge`, `diff3` and `zdiff3` all emit the opening and
closing markers, so detection never depended on the base marker. What `|||||||` actually buys is a
**complete line list**: a reader told to go resolve markers and handed a list missing one goes back to
the file for it. Fragment, constant comment and test all say that now, and the test asserts the list:

```
assert "(lines 3, 5, 7, 9)" in out
```

The reviewer had confirmed the detection worked. Nobody had checked the sentence explaining why. That
is the argument for demanding the test rather than accepting a hand-verification.

### The full suite found a red only the full suite could find

`test_no_file_in_this_checkout_names_a_pending_fragment` (#1293): a docstring in the new test file
cited `changelog.d/1162.fixed.md` by path. **The release deletes that file**, so the reference is green
for the window between this PR and the next tag, and red on that tag and every tag after — the shape
#941, #953 and #1231 each shipped, invisible from inside the PR by construction. Nothing in the
targeted run or the ten neighbouring files reaches it. Fixed.

Eleven other errors in that same run were self-inflicted (a `git-commit` run inside the worktree while
the non-hermetic suite was reading the tree); all twelve named files pass in isolation. Clean re-run on
an untouched tree: **`13238 passed, 97 skipped in 319.11s`, exit 0**. Windows still unverified.

### Two corrections of the record, both the author's own

- `_build_at_file_registry` does **not** run on every dispatch — `_AT_FILE_REGISTRY_BUILT` guards it,
  once per process, `_supertool.py:23029`. That claim was in the first report and in my review brief;
  it never reached the code or the note. It does not change the decision to run the full suite (the
  config loader is still shared surface, and the #1293 red earned the run), but the stated reason was
  wrong.
- The first report said "exit 0, suite green". It was exit 0 because the run had been piped through
  `tail -25` and the tally never reached the author — **a verdict reported without being read**. The
  figure above is read.

Issue #1681, opened 2026-08-14
Ten presets/git renders split an unquoted log subject, so a U+2028 inflates a counted list — and split_lines alone would make it worse


fdaviddpt:
Found by the #1654 agent while repartitioning the two `splitlines` registers, and deliberately left out
of that PR with a reason worth reading before anyone picks this up.

## The measurement that makes it real

`core.quotePath` quotes **pathnames only**. Measured on git 2.46.2 during #1654:
`log --format=%s`, `for-each-ref`, `branch -vv`, `show` content and `blame --line-porcelain` all hand
back U+2028 **raw**. So a commit subject or author name written by whoever made the commit reaches
these splits intact.

## The sites, ~10 across six files

`presets/git/checkout.py::main`, `diverge.py::main`, `push.py::_discarded_by_force`,
`push.py::_incoming_commits`, `status.py::main`, `trail.py::main`.

Each splits a log stream and renders **N lines**, counted. A U+2028 in one subject inflates the count
and adds a row nobody committed — a list whose length is the thing a reader acts on.

## Why the obvious fix is the wrong fix, and this is the whole reason it is a separate issue

`_untrusted.split_lines` is the repair everywhere else in the codebase because those sites **select
one line** out of a stream; consuming the separator is exactly right there.

**These sites render every line.** Narrowing the split alone means the separator is no longer consumed
— it goes to the terminal live, inside a row the tool is presenting as its own output. That trades an
inflated count for a cursor command, which is worse.

The repair is `split_lines` **plus** `visible()` per rendered line, across six files. That is a design
decision about where flattening belongs in a multi-line render, and a diff roughly twice the size of
the one it was found in.

## What a PR closing this must do

Account for all ten sites — the class body names them so nobody fixes six and closes it. And state
which of `split_lines` / `visible()` / both each site gets, because "we applied the helper" is the
sentence that produced this issue's own precondition.

## Class

`forges`, at the low end of it: text a third party wrote reaches column 0 of a rendered, counted list.
Nothing is destroyed and nothing leaves the machine; what is chosen by someone else is how many rows
the tool appears to have found. Whoever takes it should say whether `forges` still fits after reading
the sites, since the eleven-class table in this repo has been refused by seven audits running.

Filed 2026-08-14; the measurement is #1654's, re-derivable from the commands above.

Issue #1130, opened 2026-08-08
presets/git/ has ~60 unaudited str.splitlines() sites — no claim either way, which is the point


fdaviddpt:
Named by the #1119 agent as explicitly out of scope, and filed rather than left implicit because the alternative is a silence that reads as coverage.

## What is known and what is not

The forged-boundary class — `str.splitlines()` splitting on terminators the input's author controls, at a position the parser treats as structural — has now been audited and fixed twice:

- **#1105 / PR #1118** — `presets/github/`, 12 sites, 2 narrowed, 10 left with a written reason
- **#1119 / PR #1129** — `presets/gitlab/`, 9 sites, 3 narrowed, 6 left with a written reason

`presets/git/` holds roughly **60 more sites and has never been looked at**. That is not a claim that any of them are defects. It is the absence of a claim, which is the honest state and worth recording as such — the twin register shipped in #1119 covers the two twins only, so nothing on the board currently says whether this directory is clean.

## Why it is not simply "run the same sweep again"

Both prior audits narrowed a **minority** of sites, and #1105's central finding was that four of its ten untouched sites would have been made *worse* by narrowing: they lift a line out of a subprocess's stderr and print it, where consuming the separator is the safer half. #1119 then found the same thing empirically — `splitlines()[0]` consumes the separator, so the split is itself the disclosure.

So a blanket narrowing of 60 sites is a regression, not a fix. The deliverable is per-site: narrowed, or left with the reason written down.

## What makes this directory different from the two already done

`presets/git/` reads output from **git itself**, and #1119 established something useful about that: `git check-ref-format` **rejects** every ASCII control character in a refname while **accepting** U+2028. So for refname-bounded output the threat surface is genuinely narrower than for a CI log — but that argument covers refnames, not blob content, not commit messages, not stderr, and not `merge-tree` output.

Each of those is a different input author with different escaping rules. Expect the answer to be "mostly fine for a stateable reason, with two or three real ones", which is what both prior audits found.

## Also worth checking in the same pass

[...]

PR #1703, opened 2026-08-14
Three git-push fields that contradict their own [result], and a blamed file's own line forging the row that names its author (#1669, #1693)

### LINKED ISSUE #1669: git-push names a hook stdout continuation line as the failure while the actual transport error sits on the last line it printed
Hit live pushing a one-line docs commit to `master` this afternoon.

## What it printed

```
Status: PUSH REJECTED X
First error: ^ counts skips carrying that token only, not every symlink-dependent test: one held off
this runner by an unrelated collection-time marker (no O_NOFOLLOW, a posix-only class) skips without
it, and a symlink call inside an `except OSError` arm does not skip at all. Full population:
tests/test_symlink_gating_register_1232.py
Hint: git reported no ref status for origin/master - the push was stopped before it reached the
remote (local pre-push hook, or transport). Not a divergence: a rebase would not help.
```

and, on the last two lines of the relayed git output:

```
> ========== 13259 passed, 51 skipped, 15 warnings in 416.00s (0:06:56) ==========
> X Tests passed. Pushing.
>
> Connection to github.com closed by remote host.
```

`[result]` carries the same wrong sentence as `First error:`.

## What is actually wrong

The suite **passed** — 13259 passed, 51 skipped, and the hook printed `Tests passed. Pushing.` The
push then died in transport: `Connection to github.com closed by remote host.` A plain retry of the
same command is the correct response, and nothing in the receipt says so.

The line promoted to `First error:` is not an error at all. It is the **second line of a two-line
disclosure block** the hook prints on success — one of five such blocks (`core-timeout(#1501)`,
`symlink-capability(#1143)`, `lint-budget(#1360)`, `live-gh(#1568)`, `git-status-decline(#705)`),
each a headline followed by a `  ^ counts ...` continuation explaining what the denominator does and
does not include. Those blocks exist to make a *healthy* run legible. The receipt selected one of
them, mid-sentence, as the cause of the failure.

## Why this matters more than a cosmetic mis-pick

The `Hint:` line immediately below is **correct and well written** - it names both possibilities,
"local pre-push hook, or transport", and rules out divergence. So the receipt contains the
information needed to reach the right answer, one line under a sentence that points hard at the wrong
one. A reader who trusts `First error:` goes looking at `tests/test_symlink_gating_register_1232.py`
and the symlink register - a file that has nothing to do with anything here and whose disclosure said
`0 of 51`, i.e. *nothing was skipped*.

The consequence is real: the hook run costs ~7 minutes. Diagnosing "which symlink test broke" against
a suite that passed, before noticing the last line, is the whole cost of the defect.

Which selector picks that line. I have not read the code, and I would rather someone measured it than
took a guess from me. Two shapes worth ruling in or out, in the order I would check them:

1. The relay keeps only `first 5 and last 30` lines of a long hook output (it says so), and the
   `First error:` pick may be running over the *elided* view rather than the full stream.
2. A "looks like an error" heuristic over hook stdout - the chosen line contains `does not`, `without`
   and a `tests/...py` path, which is roughly what a naive matcher would like.

Either way the transport line was **present, last, and unselected**, which is the fact any fix has to
account for.

## What the fix has to preserve

- The hook's own output stays relayed. Truncating it would trade this defect for a worse one.
- A genuinely failing suite must still surface its failing assertion as `First error:` - that is what
  the field is for and it works. This is about the case where the hook **succeeded** and the failure
  came from git itself.
- The success/failure boundary is already printed by the hook (`Tests passed. Pushing.`), so a fix
  has something to anchor on that is not a keyword match.

## Class

`misreports`. Nothing is destroyed, nothing is lost, and the push correctly did not happen - the
receipt is honest that it failed. What it is wrong about is *why*, and it is wrong in the specific
direction that sends the reader into unrelated test code. Related but not the same: #297, a closed
issue where `git-push` misreported a pre-push hook's amend-and-push as `PUSH REJECTED`.

## Reproduction

Not deterministic - it needs a transport failure after a passing hook run. What is deterministic is
the input: any successful pre-push run on this repo prints those five two-line disclosure blocks, so
the selector's behaviour over that text can be exercised without a real network failure.

Measured 2026-08-14 on master at 402cf00, pushing one commit.

### LINKED ISSUE #1693: The two relay seams the presets/git splitlines register still cannot settle: five undocumented-but-sound, one genuinely open
Two seams the #1130 splitlines register cannot settle, left by PR #1692 (#1681/#1638) with a one-line reason each and deliberately not folded in. A PR closing this must account for both — they need opposite work.

## Instance 1 — `resolve.py` renders `path` raw at five receipt seams, on a ground nothing states

The `✓` / `⊘` / `✗` lines interpolate `path` with no `flat`. The ground that makes this safe is `QUOTED PATH`: the paths come from `diff --diff-filter=U`, which `core.quotePath` quotes, or from the caller's own argv. **That ground is written down nowhere** — not in the register, not in the file, not in `docs/presets/git.md`.

So this is not a defect to fix, it is a claim to record. What makes it worth an entry rather than a shrug: the next person auditing `presets/git` finds five unflattened interpolations and either repairs them (churn — `visible()` on octal-quoted output is a provable no-op) or leaves them and cannot say why. PR #1692 hit exactly that fork on three other sites and resolved it by reclassifying them `QUOTED PATH` in the register; these five never got the same treatment.

## Instance 2 — `investigate.py::main` is the register's only `NOT QUOTED, open`, and it parses file *content*

`blame --line-porcelain` output is not a path stream and not a subject stream — it interleaves porcelain headers with **the file's own lines**. A crafted line in a blamed file can therefore forge a header, and what that buys is a misattributed author or date in a receipt someone is reading precisely to answer *who changed this and when*.

Pre-existing, unchanged by #1692, and the only entry the register still carries in that state. Unlike instance 1 this one is a real hole, and its blast radius is the interesting part: the attacker needs a line in a file you blame, which is a much lower bar than a commit subject.

## Why one issue

Both are the same question pointed two ways — **what is the ground for each relay in `presets/git`, and where is it written down?** One turns out to have a sound ground nobody recorded; the other turns out to have none. The register is the shared artifact and the fix for both lands in it. Filed as one entry with both instances named so nobody repairs the cheap half and closes it.

## Class

Instance 1: none — the code is correct and the documentation is absent. Instance 2: `forges`, and worth re-checking against the eleven-row table by whoever takes it, since blame content is a stream shape none of the existing findings used.

Found by the #1681 agent, 2026-08-14, as adjacents it declined to take on — correctly.

fdaviddpt:
## #1669 — three fields that contradict their own `[result]`

All three instances from the body and its two comments:

| Field | Was | Now |
| --- | --- | --- |
| `First error:` on a rejected push | picked a `^ counts skips…` footnote out of the hook's **success** disclosure | picks on the `To <url>` boundary `_split_hook_stdout` already draws — git's channel, then the hook's, then git's contentless `failed to push some refs` epilogue |
| the verdict word | `REJECTED`, which reads as the remote refusing the ref | `_ref_status` empty ⇒ git never reached the remote ⇒ **`STOPPED BEFORE THE REMOTE`**. Headline, hint and verdict now come from that one bool |
| `Status:` on a no-op push | `pushed ✓` three lines above `[result] NOT PUSHED` | `_success_receipt` buffers its body and prints `Status:` from the same `moved` the `[result]` verdict uses |

**The issue offered two hypotheses for the selector and both were wrong.** It is not the `first 5 / last 30` elision and not a proximity heuristic: `_git_common`'s error test was `"error" in low`, which matches inside **`OSError`**. Measured, then word-anchored via `_ERROR_WORD_RE`.

## #1693 — and the obvious fix would have been the loud-bug-for-quiet-bug trade

Instance 1 (`resolve.py`'s five `QUOTED PATH` interpolations) is documentation: the ground is recorded in the register and **pinned**, so the next audit does not re-litigate five sound sites.

Instance 2 is the real hole, and **`_untrusted.split_lines` would not have closed it.** `_git` runs `subprocess.run(text=True)`, so Python rewrites a lone CR to LF before any preset sees the stream. Verified against a real repository: a blob containing `x = 1<CR>author Mallory<CR><TAB>I did this` arrives as three lines, two of which read as git's own porcelain. **Same forgery, plain ASCII, and no splitter undoes it** — shipping `split_lines` here would have published a register with 0 open entries over a hole reachable with one keystroke.

So `_blame_entries` takes a new `_git_verbatim` reader (same budget, same `TIMEOUT_RC`) and parses on git's own LF. Two readers, deliberately.

## The reviewer found this issue's own defect on the one caller the fix did not touch

`push.py:2543` — `_recover_by_rebase` printed `Status: pushed ✓ (rebased onto remote)` **and** called the now-self-printing `_success_receipt`: two contradictory `Status:` lines, which is #1669 itself. Reproduced and fixed with `status_suffix`.

It also reddened `test_main_non_ff_rebase_clean_pushes`, which had been asserting `pushed ✓` against a fixture whose own `[result]` said `NOT PUSHED - already up to date` — **this issue's contradiction, sitting green in the suite since the test was written.**

One finding refused: that `_push_error_line`'s epilogue tier is only reachable through an indirect regex match. `test_gits_epilogue_is_still_answered_when_it_is_all_there_is` drives that tier directly and asserts exact equality.

## Adjacents fixed

`investigate.py`'s log render and both diff renders — #1681's class one function on, where the count is the product — got `split_lines` + `visible()`, pinned by tests asserting a commit subject cannot add a row to the commit count. The blame render's `author` and `content` were interpolated raw; `visible()` now precedes truncation, taking the census 4 → 2. And a **CRLF-blob regression this fix would itself have introduced** is closed and pinned: one trailing CR consumed, a mid-line CR kept.

## Left for filing rather than taken

- `_untrusted.split_lines`' CR arm is dead for every `_git` consumer in the repo, because `text=True` already normalised it — while its docstring and the #1130 register both reason as though CR reaches them. Harmless today, load-bearing the moment another reader goes verbatim.
- `_looks_like_success` still substring-matches `failed`/`fatal`/`rejected`/`aborted`/`declined` — the same class as the bug fixed here, pointed the other way (it suppresses rather than selects). Churn without a demonstrated instance.

## Tests

Red first for both files, verbatim, before either implementation existed. Green `16 passed`, then `19 passed` after the review round. **Full suite run six times**, final `13334 passed, 97 skipped in 370.82s`.

**Windows unverified** — no `os.name` branch, no path comparison and no new spawn arm; `_git_verbatim` decodes bytes itself, so it is platform-identical.

Closes #1669
Closes #1693


fdaviddpt:
**Red on `pytest (windows-latest, 3.10)`, and it is outside this PR's blast radius — recording the evidence before the re-run destroys it.**

```
FAILED tests/test_html_check.py::test_two_script_blocks_are_paired_and_checked_independently - assert 2 == 1
1 failed, 12625 passed, 868 skipped, 5 warnings, 2 subtests passed in 644.25s
```

This PR touches `presets/git/{push,_git_common,investigate,resolve}.py` and their tests. `tests/test_html_check.py` exercises `validators/html-check`, which no path in this diff reaches.

Two facts that matter and point in opposite directions:

- **`master` at `bb0ae0b` was GREEN 22 minutes earlier**, all 16 test legs including `windows-latest, 3.10`. So this test passed on the parent and failed on the branch, which is exactly the shape that is *not* supposed to be dismissed.
- **12625 passed and only this one failed**, so nothing is disarmed, and the failing assertion is about html-check's own block pairing rather than anything a git preset could influence.

Re-running the leg as the discriminator, deliberately and with the evidence already captured here: a pass means a Windows-only flake in `test_html_check`, which is then worth its own issue rather than being forgotten; a second failure means it is real and this PR does not merge until it is understood.

Stating that in advance so the outcome cannot be read backwards into whichever answer is convenient.

Issue #1793, opened 2026-08-18
A subagent classified the supertool-required jit-context block as a prompt-injection payload and refused it


jbkkz:
Found while running the `oss` maintainer loop on an unrelated repo (`jbkkz/requivo`). Reported by a subagent, verified by the maintainer. Ranked **`misreports`** against the oss ranking table — a first-party tool constraint that reads as an attack. Not embargoed.

## What happened

A developer agent spawned two concurrent reviewers against its own diff. One of them — a stock `Explore` subagent with **no supertool briefing** — reported that reading files through its `Read` tool returned, in its words:

> a fabricated JIT Context block claiming Read/Edit/Write/Glob/Grep are blocked … a prompt-injection payload (planted content masquerading as a system directive)

It **refused to comply** and fell back to `git show` to read the files.

That block is `supertool-required.md`, the supertool plugin's own PreToolUse hook. There was no injection.

## Why the agent was right to refuse

This is the part worth sitting with: given only what that agent could see, refusing was **correct reasoning**. From inside the subagent, the block arrives as:

- unsolicited text appearing inside a tool result,
- asserting that its own tools are unavailable,
- and directing it to run a differently-named executable instead.

That is a textbook description of a tool-result injection, and "refuse and find another route" is exactly the behaviour anyone would want. The agent applied the right rule to the wrong object, and it could not have known which object it had.

## The cost, and why it is not zero

Here the review completed, so the cost was one workaround. The failure modes it generalises to are not bounded that way:

- an agent that refuses a **legitimate** instruction because it arrives through the same channel,
- an agent that spends its budget investigating a non-existent attack rather than the question it was spawned for,
- and — the inverse, which is worse — a maintainer who reads *"my reviewer detected a prompt-injection payload"* in a report and treats a real one as this false positive next time.

The block is doing its job; nothing about the guard's behaviour is wrong. What is missing is that **the text does not identify itself as a first-party tool constraint**, and identity is precisely the thing an injection cannot forge cheaply for a reader who knows to look for it.

## Direction, and a caveat about the obvious fix

The straightforward repair is a line in the block naming the plugin, the hook and where the text comes from. Worth noting before that gets written: *asserting* first-party origin is also what an injection would do, so a self-declaration helps a reader who can check it and is worth little to one who cannot. Anything verifiable from inside a subagent — the plugin root path, a pointer to the on-disk file the reader can `read` — is worth more than a claim.

[...]

PR #2207, opened 2026-09-03
chore: refresh oss-owned files from /oss:scaffold 0.20.0

fdaviddpt:
`.oss/statusline.py` is plugin-owned and replaced wholesale by `/oss:scaffold`, so a fix shipped upstream reaches this repo only when the command is re-run here. `/oss:doctor` at plugin 0.20.0 reported the drift as `would change what it does -- _channel_field, _default_branch_marker, render, _gh_default_branch_state, and 2 more`.

No issue: this is furniture, filed by the diagnostic rather than by the tracker (same shape as #2171).

The delta is claude-oss #856: a glyph glued to the repository name saying whether the default branch head is green right now. Four states (`green`, `bad`, `running`, `no-run`), with `running` and `no-run` deliberately sharing one glyph, and `None` rendering nothing rather than `?`.

The changelog trio stays declined -- this repo runs its own gate under `.github/workflows/changelog.yml`.

The `01-oss` tools layer is left as it was. Scaffold rewrote its three `.md` files and added three rows to `00-index.tsv`; all of that is reverted, per `CLAUDE.md` and the precedent in #2171 -- `.claude/jit-context/tools/00-manual/harness-tools-blocked.md` already does that job locally, and a tracked second copy reproduces the read that goes wrong (#1793). None of the three `.md` files has ever been tracked here.

Verified: `tests/test_statusline_wiring_documented_1964.py`, `test_version_sites_agree_1854.py`, `test_jit_tools_index_seven_columns_1992.py`, `test_harness_tools_do_not_ship_1791.py`, `test_jit_index_round_trips_1579.py`, `test_jit_paths_rules_declare_no_mode_1442.py` -- 42 passed locally.

https://claude.ai/code/session_01ADypoTKtynbzy81zM2znbf

### DIFF
--- a/.oss/statusline.py
+++ b/.oss/statusline.py
@@ -1102,6 +1102,48 @@ def _channel_field(channel, symbols, color=False):
     return shade + text + RESET
 
 
+#: `gh-branch`'s own four states, folded onto `_symbols`' four render glyphs (#856).
+#: `"bad"` gets its own glyph -- a leg has actually failed, the one state that is a
+#: finding rather than "not settled yet". `"running"` and `"no-run"` share `run` on
+#: purpose: both mean "nothing to act on, look again later", and collapsing them is
+#: the deliberate call the issue asked for rather than an oversight -- see
+#: `_gh_default_branch_state`'s own docstring for why they are still told apart
+#: before they reach this table, so `"no-run"` never silently becomes `"running"`.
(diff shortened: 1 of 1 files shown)


fdaviddpt:
Superseded by #2220 (the 0.21.0 scaffold refresh), which produces a byte-identical statusline.py plus more, already merged. Closing without merge; nothing here is lost.

PR #2213, opened 2026-09-03
chore: refresh .oss/statusline.py from /oss:scaffold 0.20.0

fdaviddpt:
`/oss:doctor` at oss 0.20.0 reported the owned statusline as behind the plugin:

```
WARN .oss/statusline.py: re-running /oss:scaffold would change what it does --
_channel_field, _default_branch_marker, render, _gh_default_branch_state, and 2 more.
```

This is that re-run, restricted to the one file. The local copy carried no edits of its own: the only lines the old file had that the plugin's does not are the four-line `Managed by the oss plugin` header, which `scaffold.py:854` writes itself.

## What changes

The default-branch marker (#856 upstream). One glyph glued onto the repository name saying whether the default branch's head commit is green, read off GitHub's combined-status endpoint rather than by enumerating every workflow run on the head SHA. `total_count == 0` is read separately from `state`, so "nothing has reported on this commit yet" no longer renders as "checks are running" -- which is the state right after this loop's own merge.

## What is deliberately not in it

`/oss:scaffold --apply` also rewrites the whole `01-oss` rule layer. The paths and vocabulary halves came back byte-identical. The tools half is deliberately empty in this repo -- `00-index.tsv` is empty and `00-README.md` is the only tracked file there -- so the three rule files scaffold wrote (`merge-gate.md`, `pr-create-gate.md`, `supertool-required.md`) and their three index rows were dropped again, exactly as they were at 0.19.0 in fa0b8afd. `CLAUDE.md` documents the reason for `supertool-required.md`; it does not document one for the other two, and the only evidence they should be dropped is that precedent. Worth a line in `CLAUDE.md` either way.

## Verified

- `python3 .oss/statusline.py` renders against a minimal stdin payload, on the branch.
- `git status --porcelain` in the worktree shows one modified path.

https://claude.ai/code/session_01ADypoTKtynbzy81zM2znbf

### DIFF
--- a/.oss/statusline.py
+++ b/.oss/statusline.py
@@ -1102,6 +1102,48 @@ def _channel_field(channel, symbols, color=False):
     return shade + text + RESET
 
 
+#: `gh-branch`'s own four states, folded onto `_symbols`' four render glyphs (#856).
+#: `"bad"` gets its own glyph -- a leg has actually failed, the one state that is a
+#: finding rather than "not settled yet". `"running"` and `"no-run"` share `run` on
+#: purpose: both mean "nothing to act on, look again later", and collapsing them is
+#: the deliberate call the issue asked for rather than an oversight -- see
+#: `_gh_default_branch_state`'s own docstring for why they are still told apart
+#: before they reach this table, so `"no-run"` never silently becomes `"running"`.
+_DEFAULT_BRANCH_GLYPH_KEY = {
+    "green": "ok",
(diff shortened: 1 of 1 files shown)


fdaviddpt:
Superseded by #2220 (the 0.21.0 scaffold refresh), which produces a byte-identical statusline.py plus more, already merged. Closing without merge; nothing here is lost.

PR #2349, opened 2026-09-06
chore(oss): re-scaffold .oss/statusline.py from oss 0.26.0

### LINKED ISSUE #2350: claude-oss#1035 landed: re-scaffolding .oss/statusline.py reddens 20 tests that pin the pre-guard behaviour
`/oss:doctor` under oss 0.26.0 reports `.oss/statusline.py: re-running /oss:scaffold would change what it does -- parse_channel_report, channel_status, board_from_cache, git_release_progress, and 31 more`. Re-running it is correct, and it turns four CI legs red.

## What happened

PR #2349 (`chore/scaffold-statusline`) is `scaffold.py --root . --config .oss.json --apply` under oss 0.26.0, one tracked file changed. Result on `c0d34cf`:

```
checks: 23 total: 9 passed, 4 failed, 10 pending
  failed: pytest (ubuntu-latest, 3.9)  job #101569064126
          pytest (ubuntu-latest, 3.10) job #101569064047
          pytest (ubuntu-latest, 3.11) job #101569064152
          pytest (ubuntu-latest, 3.12) job #101569064023
```

`20 failed, 16524 passed, 223 skipped` on each leg. Every failure is `assert 0 == 2` or `assert 0 == 1`, `where 0 = len([])` -- the `gh api` call was never made:

- `tests/test_statusline_api_ref_validation_2245.py` -- 7 cases of `test_malformed_repo_or_branch_still_reaches_the_api_call`
- `tests/test_statusline_repo_only_api_ref_validation_2278.py` -- 13 cases across `test_gh_count_still_reaches_the_api_call_for_a_malformed_repo`, `test_gh_external_issue_count_still_reaches_the_api_call_for_a_malformed_repo` and `test_latest_release_now_reaches_the_api_call_for_a_shape_invalid_repo`

## Why

#2299 gave the plugin its statusline back and filed this repo's local guards upstream as [claude-oss#1035](https://github.com/Digital-Process-Tools/claude-oss/issues/1035). It then repointed both test files at the plugin's unguarded behaviour, and said so in the docstring of the first one:

    Mocked to capture the call args rather than raise, so this fails loudly (not
    silently) the day a guard is reintroduced and starts short-circuiting these
    before `_run`.

That day is today. **claude-oss 0.26.0 ships the guards.** Measured on the two copies:

```
$ grep -cE '_malformed_repo|_malformed_api_ref|_BRANCH_UNSAFE_RE' .oss/statusline.py           # master
0
$ git show chore/scaffold-statusline:.oss/statusline.py | grep -cE '...'                        # 0.26.0
17
```

The scaffolded copy carries `_malformed_repo` (line 1419), `_BRANCH_UNSAFE_RE` (1459) and `_malformed_api_ref` (1462), citing `#1035, #1051` in its own comments, and eight call sites now return before `_run`.

So the product is right and the tests are stale. The gap #2299 documented as accepted is closed upstream, and the tripwire it left is the thing reporting it.

## What to do

Invert the 20 parametrised cases in the two files: a malformed `repo` or `branch` must now be refused **before** `_run`, so the assertion becomes `len(calls) == 0`. The names say the opposite and go with them (`..._still_reaches_the_api_call` becomes `..._is_refused_before_the_api_call`). The module docstrings of both files are the #2299 rationale in full and need rewriting to the post-#1035 state, citing the version that carries it rather than the one that did not.

Keep both must-not-fire controls as they are -- `test_a_normal_repo_and_branch_still_reach_the_api_path` and the three `..._a_normal_repo_still_reaches_the_api_path` siblings are the positive half of the pair, and a suite asserting only that nothing fires passes when nothing runs at all.

One case needs a decision rather than an inversion: `test_latest_release_still_refuses_a_falsy_repo_before_the_api_call` already asserted a refusal for a falsy `repo` and is passing. Check whether the new guard subsumes it before merging the two files' coverage.

#2349 carries the scaffold half and is red on exactly this. Either land the test change on that branch or close it and do both in one.

[AI-generated]

fdaviddpt:
Closes #2350.

`/oss:doctor` reported the owned file as `would change what it does -- parse_channel_report, channel_status, board_from_cache, git_release_progress, and 31 more`. Owned files are replaced wholesale by `/oss:scaffold`, so a fix shipped in the plugin reaches this repo only when the command is re-run here. This is that re-run.

Written by `scaffold.py --root . --config .oss.json --apply` under oss 0.26.0. What the run reported:

- `.oss/statusline.py` replaced (the only tracked change: 395 insertions, 172 deletions)
- the changelog trio declined, because this repo already runs its own gate (`.github/scripts/assemble_changelog.py`, `.github/workflows/changelog.yml`)
- the 11 files of the `01-oss` rule layer rewritten byte-identical, so `git status` shows none of them
- 12 templates already present, left untouched

No changelog fragment: `.oss/` is outside the user-visible path set `changelog.yml` gates on, so that leg reports `skipped`.

Re-running `/oss:doctor` after this drops the `owned files` warning; the `clone HEAD` warning in the same report was a stale local master and is unrelated to this diff.

## Update: #2350 landed on this branch

The scaffold reintroduced `.oss/statusline.py`'s malformed-`repo`/`branch` guards (`_malformed_repo`, `_BRANCH_UNSAFE_RE`, `_malformed_api_ref` -- filed upstream as `claude-oss#1035`), which reddened 20 test cases in `tests/test_statusline_api_ref_validation_2245.py` and `tests/test_statusline_repo_only_api_ref_validation_2278.py` that had been repointed at the pre-guard behaviour by #2299. `a6a4027` (fast-forwarded onto this branch) inverts those 20 assertions to `len(calls) == 0`, rewrites both module docstrings to the post-guard state, merges `test_latest_release_still_refuses_a_falsy_repo_before_the_api_call`'s coverage into the shared `MALFORMED_REPO` parametrization, and keeps every must-fire positive control unchanged. Red: `20 failed, 6 passed`. Green: `26 passed`. Self-review (Explore + oss:auditor, spawned against the diff) returned `NO FINDINGS`; the auditor's mutation check confirmed the inverted assertions are genuinely exercised, not vacuous. Full detail in the developer's own note: `/Users/floriandavid/Documents/st-wt/notes/fix-2350-20260906T225530Z.md`.

https://claude.ai/code/session_01XKeLFSWdSpSTMasdJsPZSE

[AI-generated]


fdaviddpt:
Red on 4 legs, and it is a real failure, not a flake: `20 failed, 16524 passed` on each of `pytest (ubuntu-latest, 3.9|3.10|3.11|3.12)`, every one of them `len(calls) == 0` where the test expects 1 or 2.

Cause and remedy are in #2350. Short version: oss 0.26.0 ships the guards this repo filed upstream as claude-oss#1035, so `.oss/statusline.py` now refuses a malformed `repo`/`branch` before the `gh api` call. The two test files repointed in #2299 pin the pre-guard behaviour and say so in their own docstrings. The file in this diff is right; the tests are stale.

Part of #2350. Do not merge until the test half lands.

Issue #2486, opened 2026-09-09
the claude-channel consumer cannot start on a fresh plugin install: its dependency is never installed


fdaviddpt:
#520 declares the consumer for every plugin install via `${CLAUDE_PLUGIN_ROOT}`:

```json
{"mcpServers": {"claude-channel": {"command": "node",
  "args": ["--experimental-strip-types", "${CLAUDE_PLUGIN_ROOT}/notifiers/claude-channel/channel.ts"]}}}
```

`channel.ts:32-33` imports `@modelcontextprotocol/sdk` at load. `node_modules/` is gitignored and untracked -- `git ls-files notifiers/claude-channel/` returns eight files and no dependency tree. The only `npm install --omit=dev` lives in `install.sh`, which the README added by the same commit tells plugin users to skip:

> A plugin install needs none of this. The install script and the manual wiring below are only for a git clone you are running outside the plugin.

`package.json` declares no `postinstall`; `.claude-plugin/plugin.json` declares no install hook.

Observed rather than reasoned -- `channel.ts` copied to a directory outside any `node_modules` and run exactly as `.mcp.json` runs it:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@modelcontextprotocol/sdk'
```

Why nobody has felt it here: the installed cache at `dpt-plugins/supertool/0.58.0/notifiers/claude-channel/` carries every shipped file stamped 2026-09-08 18:01 and `node_modules` stamped 19:53 -- an hour and fifty-two minutes later, from a manual `install.sh` run on this machine, not from anything the marketplace shipped.

Found by the v0.59.0 release audit, round 2 (token `rel-59-r2-m4tx`), ranked `misreports`, non-blocking, so the gate carried it forward. The row undersells it: the consequence is not a misleading message, it is a declared server that never starts for anyone installing fresh.

## The constraint any fix has to respect

**stdout is the JSON-RPC stream.** Anything an installer prints there corrupts the MCP handshake, so a fix that shells out to npm must send every byte of its output to stderr.

## Not the fix

Reimplementing the three SDK surfaces this file uses (`new Server`, `connect(new StdioServerTransport())`, `notification()`) as raw newline-delimited JSON-RPC. It is perhaps eighty lines and it replaces a dependency problem with a protocol-compatibility problem, in the one component whose failures are hardest to see from inside a session.

## Also

[...]

PR #2538, opened 2026-09-13
fix(claude-channel): spawn npm via a shell so the auto-install works on Windows

biglat1595:
## Problem

`start.mjs`'s auto-install (added for #2486) runs `spawnSync("npm", [...])` without a shell. On Windows the executable on `PATH` named `npm` is actually `npm.cmd`; `spawnSync` cannot execute a `.cmd` file directly and fails with `ENOENT`, regardless of what `PATH` contains. So on every fresh Windows install, the auto-install itself fails, and `claude-channel` never starts — the exact failure #2486 set out to close, still open there.

Naming the executable explicitly as `npm.cmd` does not fix it either — it throws `EINVAL` instead, a documented Node/libuv limitation: spawning a `.bat`/`.cmd` file on Windows requires a shell no matter how the executable is named.

## Fix

Pass `shell: true` to the `spawnSync` call. The args passed (`install`, `--omit=dev`, `--no-audit`, `--no-fund`) are a fixed, static list with no spaces or shell metacharacters, so the unescaped-argument concatenation that makes `shell: true` risky in general (and that triggers Node's `DEP0190` deprecation warning) does not apply here — and that warning, if it prints, goes to inherited stderr, never to the stdout JSON-RPC stream the file is careful to protect elsewhere.

## Verification

On Windows: removed `node_modules`, ran `node --experimental-strip-types start.mjs` directly. Before the fix: `ENOENT` (or `EINVAL` with the `npm.cmd` variant), fatal, exits 1. After: auto-install runs, `@modelcontextprotocol/sdk` resolves, server starts.

https://claude.ai/code/session_01MSbhQgFJmc1cYb1vXKsG5C


fdaviddpt:
Thanks, this is a real bug and the diagnosis is right. `CreateProcess` cannot execute a `.cmd` whatever `PATH` holds, the `EINVAL` you get from naming `npm.cmd` explicitly is the documented libuv behaviour, and the argv here is a fixed list with no metacharacters, so `shell: true` adds no quoting surface. The DEP0190 warning going to inherited stderr rather than the protocol stream is the right thing to have checked.

I pushed two commits to your branch rather than asking for them, since `maintainer_can_modify` was on.

**A positive control in `tests/test_channel_dependency_2486.py`.** The fix had no guard, and the existing test could not have caught the bug. It asserts that a failed install exits non-zero, names a remedy and writes nothing to stdout. A launcher that never spawned npm at all satisfies all three: `npm could not be started (ENOENT)` is loud, names `install.sh`, and touches no stdout. So on Windows that test has been passing through the unreached arm since #2487, and the `npm.cmd` stand-in it writes at line 102 was never executed -- the comment beside that stand-in says the opposite, and I have corrected it. Asserting `npm exited 127` separates the two arms: the stand-in's own exit code can only reach stderr if the spawn actually reached it.

Worth being explicit about what that assertion has and has not shown. It is green on darwin both before and after your fix, because on POSIX `shell: true` changes nothing about resolving `npm`. I could not get it red on any platform I have. The windows legs of the `pytest` matrix are the only place it discriminates, which is the whole reason it needed to exist.

**A changelog fragment**, `changelog.d/2486.fixed.windows-npm-cmd.md`. One per pull request is required here and `oss-changelog.yml` refuses without it.

Two things left before this can merge:

- No checks have run. You are a first-time contributor, so the workflows need approving from our side. I will do that.
- The body declares no closing reference. #2486 is already closed, so this is a follow-up on the same defect rather than a new fix for it, and I have keyed the fragment to that issue.

Separately, your PR found a class rather than an instance. Seven validator and formatter adapters do the same thing from Python -- `shutil.which()` consults PATHEXT and returns `eslint.cmd` or `tsc.cmd`, then `subprocess.run()` spawns the bare name and cannot execute it. Filed as #2540. `prettier-check` is the bad one there: it reports a finding on every file rather than declining. Not your problem to fix, but you found it.

[AI-generated]

PR #2539, opened 2026-09-14
Potential fix for code scanning alert no. 4: Incomplete URL substring sanitization

fdaviddpt:
Potential fix for [https://github.com/Digital-Process-Tools/claude-supertool/security/code-scanning/4](https://github.com/Digital-Process-Tools/claude-supertool/security/code-scanning/4)

Use URL parsing (`urllib.parse.urlparse`) and assert on the parsed hostname rather than checking for a URL literal substring in `out`.

Best minimal fix in `tests/test_secret_redaction_760.py`:
1. Add `from urllib.parse import urlparse` to imports.
2. In `test_bearer_header`, keep the `"curl" in out` assertion.
3. Replace `assert "curl" in out and "https://api.example.com" in out` with logic that extracts URL tokens from `out`, parses them, and asserts at least one parsed URL has `hostname == "api.example.com"`.

This preserves test intent (“redaction should not remove the endpoint host”), avoids substring-based URL matching, and satisfies CodeQL’s recommendation to parse before checking host.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


fdaviddpt:
Closing. The alert this fixes is a false positive, and the fix is weaker than the assertion it replaces.

`py/incomplete-url-substring-sanitization` targets a trust decision made from a substring check -- `if "example.com" in url: allow`. Line 140 is a test assertion that `_secrets.redact()` preserved the endpoint, on a string literal written two lines above it. Nothing is sanitized, nothing is authorized, and no input in that function is attacker-controlled. The `high` severity is the rule's class, not this instance.

The rewrite loses something real. The old assertion pins the exact substring `https://api.example.com`. The new one accepts any whitespace-delimited token starting `http://` or `https://` whose parsed hostname is `api.example.com` -- so a redactor that downgraded the scheme to `http://` fails the old test and passes the new one. The scheme is part of the endpoint the test exists to protect. It is also brittle the other way: punctuation glued to the URL lands inside `urlparse(...).hostname` and fails for a reason unrelated to redaction.

And it does not remove the pattern from the file. `tests/test_secret_redaction_760.py:171` is `assert "example.com/repo.git" in out` -- the same construct, one line away, unflagged. CodeQL caught one of two identical shapes, so silencing that one leaves the file exactly as it was and four lines longer.

Alert 4 dismissed as a false positive with that reasoning. No `.github/codeql*` config exists here (default setup), so there is no `paths-ignore` to add without switching to advanced setup, which is not worth doing for one test-file false positive.

Nothing against the autofix as a prompt -- it is a suggestion to weigh, and this one did not survive weighing.

[AI-generated]

Issue #1311, opened 2026-08-11
grep refuses a bounded pattern as catastrophic backtracking, and help:paste documents only half the op


fdaviddpt:
Two from the #1299 and #1293 agent runs, both costing a round-trip each time.

## 1. The backtracking guard is over-eager on `?`

`grep` refused:

    [0-9]+\\.(added|fixed)(\\.[a-z-]+)?\\.md

with *"nested unbounded quantifiers — would risk catastrophic backtracking"*.

`(...+)?` is bounded to **one** repetition of the group. Catastrophic backtracking needs `(...+)+` or `(...+)*`, where the outer quantifier can re-partition the same input unboundedly. Refusing `?` costs a legitimate pattern and teaches the caller that the guard is noise, which is the expensive part.

The fix is in which outer quantifier the check looks for, not in the check existing.

## 2. `help:paste` documents only the colon form

It shows `paste:::PATH:::CONTENT`. The `@-` payload route and its field names (`path`, `content`) are absent — the agent found them by guessing, then from an error message. `literal_backslashes` is discoverable only by tripping the refusal.

Every mutating op takes a payload route and it is the recommended one for anything containing code, so `help:` omitting it points readers at the weaker form.

## 3. The backslash refusal names a count but shows one offender

`(2x)` tells you how many are left; it shows one truncated excerpt. With several offenders you fix one, resubmit, and learn about the next. Same shape as #1303's second item: a count says how much work remains, the list says what it is.

## 4. `git-commit:::MESSAGE` with a multi-line message misdiagnoses

It fails with `no PATHS were given` where the real problem is that the message's newlines break pathspec parsing. The recovery block is excellent — it got both agents to `git-commit:@-` in one hop — but the first line names the wrong cause. Related to #1303's second item and to #1011, which is about the same form and colons.

PR #2546, opened 2026-09-16
fix(#2535): grep backtracking guard now checks every group's own quantifier

### LINKED ISSUE #2535: grep backtracking guard misses a nested unbounded group inside an unquantified outer group
Found by the gate-3 release-audit round 2 for the v0.60.0..HEAD delta (dispatch token `b79cbfb18a3f94b1`), class B (a guard that did not run), row `unranked`. Non-blocking -- the round-2 disposition (`carry-forward-and-proceed`) let the release proceed with this filed rather than fixed.

**What's missing.** `_has_outer_wrapped_unbounded_group` in `_supertool.py` (~lines 6932-7011, feeding `_op_grep`'s catastrophic-backtracking refusal, #1311) resolves an outer `( ... )` group and then sets `i = j`, skipping past the whole group -- so an inner unbounded-quantified group nested inside an *unquantified* outer group never gets its own trailing-quantifier check.

Measured at HEAD (read-only import of `_supertool`, no CI needed to reproduce):

```
'(x(a+)+y)'            new refuses=False   old (pre-#1311) refuses=True
'(?:(a+)+)'            new refuses=False   old refuses=True
'foo(bar(a+)*)?'       new refuses=False   old refuses=True
elapsed on '(x(a+)+y)' against 26 a's: 2.04s (doubles per character)
```

`_op_grep("(x(a+)+y)", ".", limit=1)` does not refuse at HEAD -- it should, and does not: the old (pre-#1311) regex-based guard caught this exact shape, so #1311's depth-aware replacement is *narrower* than what it replaced on this one case.

**Impact.** A caller's own grep pattern (not remote/untrusted text) can hang the core supertool process with exponential backtracking. No existing test in `tests/test_grep_backtracking_guard_outer_wrap_1311.py` covers a nested unbounded group inside an unquantified outer group (all 5 existing cases are top-level shapes).

**Fix shape.** Recurse into a resolved group's own body (or continue the scan from `i + 1` instead of jumping to `i = j` after resolving the outer group), so every group's own trailing quantifier is examined regardless of nesting depth.

[AI-generated]

fdaviddpt:
Closes #2535

`_has_outer_wrapped_unbounded_group` (the ReDoS guard feeding `_op_grep`'s catastrophic-backtracking refusal) resolved a `(...)` group by depth-counting to its matching close-paren, then unconditionally jumped the scan index past the whole group (`i = j`) -- so a nested group's own trailing quantifier was never independently checked when the *outer* group was not itself quantified (e.g. `(x(a+)+y)` was never refused, even though the inner `(a+)` is immediately followed by its own `+`).

One-line fix: `i = j` -> `i += 1`, so the scan revisits every `(` character-by-character and every group at any nesting depth gets its own check.

## Tests
- New file `tests/test_grep_backtracking_guard_nested_unquantified_outer_2535.py`: 5 tests -- 3 refusal cases straight from the issue's own repro table, plus 2 positive controls carried over from the sibling #1311 test to guard against regressing that fix.
- Red confirmed two ways: a standalone repro script against `_has_outer_wrapped_unbounded_group` directly, and the pytest file itself run against unfixed code (stash/pop `_supertool.py`) -- "3 failed, 2 passed".
- Green after the fix: "10 passed" (5 new + the 5 sibling #1311 tests), no regression.
- Bare `pytest` not run locally, per repo convention -- CI is the matrix that answers.

## Self-review
Two reviewers (`Explore`, `oss:auditor`) spawned concurrently against the committed diff, both returned no findings. Performance checked directly: 30-deep nested group and 50 sibling groups both under 1ms; the character-by-character rescan is not a new DoS vector since `_op_grep` already caps pattern length at 1000 chars.

## Adjacent finding, not fixed here
`_op_around` and the shared `_pattern_gate` (used by `read:PATH:::grep=` and `grep_filter`) never call this guard at all -- the same ReDoS class is reachable through a different call path this diff does not touch. Filing separately.

[AI-generated]


fdaviddpt:
Maintainer review (tick wait+review step): CI is red on this PR -- CodeQL flagged 1 high-severity alert on `tests/test_grep_backtracking_guard_nested_unquantified_outer_2535.py:27`, "Inefficient regular expression", against the string literal `"(x(a+)+y)"` used as the *input* to `_op_grep` in `test_nested_unbounded_group_inside_unquantified_outer_group_is_refused`.

Read as a false positive: that string is test data passed to `_op_grep`, and `_has_outer_wrapped_unbounded_group` (the guard this whole PR fixes) runs and refuses it *before* the pattern ever reaches `re.compile`/`re.search` -- confirmed by reading `_op_grep`'s own body (the ReDoS-guard block sits ahead of any `re` call on `pattern`). CodeQL's static regex analysis has no visibility into that guard; it is scoring the literal as if Python were about to execute it as a live regex.

I re-ran the check independently rather than trusting the report: with the one-line fix (`i = j` -> `i += 1`) reverted, `pytest tests/test_grep_backtracking_guard_nested_unquantified_outer_2535.py` gives `3 failed, 2 passed`; with the fix restored, `5 passed` -- matches the PR description exactly, so the test is real and not vacuous.

Rest of the check arithmetic and premise look clean: 23 legs (8 passed / 1 failed / 14 pending at read time), blast radius is the guard function plus its own test file plus a changelog fragment, and the adjacent-finding claim ("`_op_around`/`_pattern_gate` don't call this guard, filing separately") checks out -- #2547 exists and matches.

This needs a fix before it can go green, not a re-run: either a scoped CodeQL suppression on that line with the guard-runs-before-compile justification inline, or restructuring the test so the pattern string isn't visible to CodeQL's static scan (e.g. built from parts rather than a single literal). Routing as needs-fix, not merging.

[AI-generated]

PR #2559, opened 2026-09-16
fix(#2556): probe lock_dir writability before paying the full retry deadline

### LINKED ISSUE #2553: serialize_once still fails open: an absent lock file means the holder released it, not that lock_dir is unusable
`master` is red at ffe3bb1, the #2552 merge itself, on `tests / pytest (macos-latest, 3.9)`, job #104913047946. The failing test is the positive control #2552 added for its own fix:

```
FAILED tests/test_go_warmup_lock_2331.py::test_two_racing_calls_still_serialize_when_contention_raises_permission_error
  AssertionError: serialization did not hold when contention raised PermissionError instead of FileExistsError (max concurrently active: 2)
assert 2 == 1
```

1 failed, 16489 passed. Not Windows this time, and not the test #2551 was filed from.

## Mechanism

#2552 disambiguates "someone holds it" from "the lock file is unusable" by asking whether a lock file exists:

```python
if not isinstance(exc, FileExistsError) and not _exists_or_assume_so(lock_path):
    time.sleep(0.05)
    if not _exists_or_assume_so(lock_path):
        return fn()
```

Absence has two causes, not one. The lock file is absent when `lock_dir` is unwritable and no lock was ever created. It is *also* absent when the holder has just released it -- `os.close(fd)` then `lock_path.unlink()` in its own `finally`. In the second case absence means the lock is **free**, and the right move is to retry `os.open`, not to bypass the lock and run `fn()`.

The losing interleaving, which is what the macOS leg hit:

1. thread A holds the lock, runs `fn`, unlinks the file
2. thread B's `os.open` raises the translated `PermissionError`; `_exists_or_assume_so` says gone; it sleeps 0.05s; still gone; it returns `fn()`
3. thread C's `os.open` succeeds against the now-free path and enters `fn`

Two callers inside `fn`. `max_active == 2`.

The 0.05s recheck narrows the window, it does not change the inference drawn from absence. That is why this reproduces on a platform the original issue never involved: the test translates `FileExistsError` to `PermissionError` on every contended open, so every waiter takes the new branch, and the branch is wrong whenever the holder releases while a waiter is inside it.

## Direction (a hint, not a patch)

Absence is not evidence of an unusable `lock_dir` -- retrying the create is. Retry the `os.open` on that branch and only fall back to `fn()` when the create keeps failing with no lock file appearing, bounded by the existing `deadline`. An unwritable `lock_dir` fails every attempt, so it still reaches the fallback; a holder mid-release succeeds on the next attempt, which is the behaviour the lock exists for.

Keep both existing assertions. A third is needed for this shape specifically: a waiter that takes the not-FileExistsError branch at the exact moment the holder releases must still serialize, not fall through.

Filed by the maintainer loop from a board-level `went_failed` on master, not by a lane. Related: #2551, #2552.

[AI-generated]

### LINKED ISSUE #2556: serialize_once's fallback-on-timeout retry pays the full timeout_s even when lock_dir is unusable
**Symptom.** #2553's fix (merged in #2555) replaced a single 0.05s absence-recheck with a bounded retry loop on `os.open()` in `tests/_go_warmup_lock.py::serialize_once`, correctly fixing the race where an absent lock file was misread as "lock_dir unusable" when it could also mean "the holder just released it." But the retry loop is bounded by the *same* deadline used for genuine lock contention, so a `lock_dir` that is actually unusable (unwritable, wrong permissions, missing) now waits the full `timeout_s` before falling back to running `fn()` directly, instead of the ~0.1s the old single-recheck code took.

**Location.** `tests/_go_warmup_lock.py::serialize_once`, the retry loop added in #2555 (commits d09a349e, 6a44c7e9 on the now-merged `fix/2553` branch).

**Mechanism.** Real call sites pass `GO_WARMUP_S` scaled by `platform_factor()` as the deadline -- minutes, not seconds. So every `go-vet`-family test that happens to hit an unusable `lock_dir` (a real, if rare, condition -- see the pre-existing `test_an_unwritable_lock_dir_falls_back_to_running_fn` case) now burns minutes of wall clock per call before falling back, where it used to burn ~0.1s. CI cannot see this cost today because no current test exercises an unwritable `lock_dir` under the new retry path with a realistic (minutes-scale) deadline -- the existing unwritable-dir test uses a short deadline, so the cost is real but currently invisible to the suite.

**What would settle it, and the proposed fix.** Absence of a lock file is genuinely ambiguous (racing a holder vs. a dead directory), which is exactly what #2553 fixed by retrying. But *writability of `lock_dir` itself* is not ambiguous, and can be answered with zero patience spent: one probe `os.open()` of a unique throwaway filename in `lock_dir` at the very start (or the first time a real create attempt fails) tells you immediately whether the directory is usable at all, independent of whether the *specific* lock file is currently held. Usable-but-no-lock-file means genuinely racing a holder (retry, as #2555 already does correctly); unusable means fall back to `fn()` immediately, without waiting out the deadline.

This is a performance/cost fix layered on top of #2555's already-correct control flow, not a correctness bug in the merged fix -- #2555 should not be reopened or blocked on this. Filed as its own issue per the maintainer's own review of #2555.

[AI-generated]

fdaviddpt:
Fixes the fallback-on-timeout retry added in #2555 so it no longer pays the full `timeout_s` when `lock_dir` is genuinely unusable.

`serialize_once`'s retry loop (added in #2555 to fix #2553's real race between "holder just released the lock" and "lock_dir is dead") was bounded by the same deadline used for genuine contention. A `lock_dir` that is actually unusable now waits the full `timeout_s` -- minutes, at real call sites -- before falling back to `fn()`, instead of the ~0.1s the pre-#2555 code took.

`_lock_dir_usable()` probes `lock_dir` writability once, up front, with a unique throwaway filename (uuid4) that cannot collide with a real lock file or the Windows delete-pending window `_exists_or_assume_so` tolerates. Any `OSError` on that probe means the directory is genuinely unusable, so `serialize_once` returns `fn()` immediately instead of retrying out to the full deadline. Writability of `lock_dir` is not ambiguous the way absence-of-lock-file is (which #2555 correctly retries on) -- it can be answered with zero patience spent.

A self-review follow-up commit fixes two findings from a spawned `oss:auditor`: the probe's own failure modes (`os.open`, `os.close`) previously either swallowed silently or could raise past the module's "never fails" contract; both now warn (mirroring the file's existing unlink-failure warning pattern) and neither can raise out of the function.

Closes #2556

[AI-generated]


fdaviddpt:
Maintainer review (#2556): check arithmetic sums (22 passed, 0 failed, 1 superseded excluded per #1458 = 23 total), premise matches the linked issue, blast radius is confined to `tests/_go_warmup_lock.py`/`tests/test_go_warmup_lock_2331.py`, and the two self-review fixes (probe `os.open`/`os.close` no longer able to raise past the module's "never fails" contract) are load-bearing and read against the code, not just the report.

On the open github-code-quality thread (tests/_go_warmup_lock.py:142, empty `except OSError: pass`): below the bar for a filed issue. The unlink is best-effort cleanup of the probe's own uniquely-named (`uuid4`) throwaway file; a failed delete leaves an orphaned probe file with no other holder and no effect on `serialize_once`'s correctness, and the pattern already exists uncommented at line 291-293 of the same file (best-effort stale-lock unlink before a `continue`). Not adding a comment here is consistent with that existing convention rather than a new gap. Decision: ready-to-merge as-is; not filing a new issue for this thread.


[AI-generated]

PR #2577, opened 2026-09-17
fix(#2575): never spawn a repo-planted file resolved only via the cwd

### LINKED ISSUE #2575: validators/formatters resolve a Windows adapter via shutil.which(), which prepends the inspected repo's own cwd -- a repo-planted .cmd now executes
Filed by the maintainer loop's release-auditor (dispatch token `95999d02ddecfeac`, round 2, delta `v0.61.0..HEAD`), ranked here as **blocking** (row `executes` on the ranking table: "a file supplied by the repository under inspection is run as a program by a tool pointed at that repository" -- unconditional block, embargo yes). This stops the v0.62.0 tag.

**Composition defect, assembled from two individually-clean halves in this same delta.**

1. `#2540`/`#2542` (already merged, correctly fixes a real Windows spawn bug) changed `validators/common/spawnable.py:82` (`argv0`) and `validators/common/bin_resolve.py:46` (`_spawnable`), consumed by roughly 18 adapters (`validators/ruff/ruff.py:177,211`, `markdownlint.py:207`, `tsc-check.py:200`, `git-status.py:348`, `prettier-check.py:118,169`, `go-vet.py:214`, `actionlint.py:168`, `ruby-check.py:121`, plus `mypy`, `eslint`, `stylelint`, `terraform-check`, `hadolint`, `gitleaks`, `psr`, `pyright`, `gofmt-check`, `html-check`, `new-file-lint`, `cargo-check`), to spawn the **path `shutil.which()` returned** instead of the bare adapter name.
2. CPython's own `shutil.which()`, on `win32`, does `path.insert(0, os.curdir)` unless `NoDefaultCurrentDirectoryInExePath` is set (gated by `_win_path_needs_curdir` / `NeedCurrentDirectoryForExePath`, default TRUE). Grepped: that env var is set nowhere in `_supertool.py`, `validators/` or `presets/`. `which()` also honours `PATHEXT`, so a `.cmd`/`.bat` in the current directory matches ahead of every real PATH entry.
3. `_supertool.py:24973` (`_validator_resolve`) and `_supertool.py:26777` (`_formatter_run_one`) spawn every adapter with **no `cwd=`**, so the adapter subprocess inherits supertool's own cwd -- the repository the tool is pointed at.

**What an attacker gets**: a cloned repository containing `ruff.cmd`, `markdownlint.cmd`, `git.cmd` or `tsc.cmd` at its root has that file resolved ahead of every PATH entry and executed, on Windows, the moment a user runs the corresponding validator or formatter against the clone.

**Why this delta specifically makes it reachable, not a pre-existing condition**: at `v0.61.0` the adapters gated on `shutil.which(name)` and then spawned the **bare name** (verified: `git show v0.61.0:validators/markdownlint/markdownlint.py:199-206` -- gates on `which`, spawns `["markdownlint", "--", file]`). `CreateProcess` appends only `.exe`, so the repo-planted `.cmd` was found by the `which()` gate and then failed to actually spawn, landing in `refusal.absent()`. This delta replaces argv[0] with `which()`'s own return value, converting that dead end into real execution. The plant moves from needing a compiled `.exe` (unusual to commit) to a committable text file (`.cmd`).

**The regression guard has no positive control for this shape**: `tests/test_spawnable_register_2540.py` asserts adapters *use* `argv0`; `tests/test_windows_cmd_spawn_2540.py` asserts a `.cmd` **on PATH** is spawnable. Neither test plants a `.cmd` in the current working directory and asserts it is refused, so nothing in the delta would redden against this class. (Reasoned, not observed -- the auditor did not run the suite.)

**The one fact that would settle it directly** (needs a Windows host): `os.chdir(tmp)`, write `ruff.cmd` into `tmp`, call `validators/common/spawnable.argv0("ruff")`, and check whether the returned path resolves inside `tmp`.

**Fix direction** (a design call, not prescribed here): pass an explicit `path=` to `shutil.which` that excludes `os.curdir`/the cwd, or reject a resolution that lands inside the tree being inspected, at the shared `argv0`/`_spawnable` chokepoint rather than per-adapter.

This issue is on the v0.62.0 release's critical path per gate 3 (`commands/release.md` / `skills/manager/phases/release.md`): round 2 is the audit's hard cap, so no third full-delta audit round will run, but this specific blocking row's fix must land and be verified before the tag moves.

[AI-generated]

fdaviddpt:
Closes #2575.

## What was wrong

`shutil.which()` inserts `os.curdir` at the front of its search path on
Windows unless `NoDefaultCurrentDirectoryInExePath` is set -- even when a
caller passes an explicit `path=`, since the insertion happens after the
`path is None` branch, not inside it. Every validator and formatter
chokepoint (`_validator_resolve` / `_formatter_run_one` in `_supertool.py`)
spawns its adapter subprocess with no `cwd=`, so the subprocess inherits
supertool's own cwd -- the repository under inspection. A clone shipping
`ruff.cmd` / `markdownlint.cmd` / etc. at its own root had that file resolved
ahead of every real PATH entry and executed on Windows the moment the
matching validator or formatter ran against it.

A second, narrower path to the same shape: the raw `os.path.isfile(name)`
fallbacks in `_already_a_path`/`_is_executable` also resolved a bare name
relative to cwd directly, with no `which()` call involved at all.

Filed by the round-2 release auditor as a blocking security finding
(#2575), ranked on the "executes" row of the ranking table -- a file
supplied by the repository under inspection running as a program by a tool
pointed at that repository. It stopped the v0.62.0 tag.

## The fix

`which_excluding_cwd()` in `validators/common/spawnable.py` is the shared
chokepoint now used by `argv0()`, `spawnable()`, and `bin_resolve.py`'s
`_is_executable()`/`_spawnable()`: it walks PATH in `shutil.which()`'s own
order, skipping any entry whose directory is the current one, so a genuine
PATH-installed copy still resolves and only the implicit, attacker-reachable
cwd match is refused. The bare-name `isfile` fallbacks are now gated on the
name containing a directory component, so a bare name is never resolved
relative to cwd.

`tests/test_which_excludes_cwd_2575.py` is the new regression guard -- 8
cases, including the positive control (a real PATH entry still resolves)
paired with the negative one (a cwd-only match is refused), the shape
#2575's own report named as missing from the existing suite.

## Verification

- `python3 -m pytest tests/test_which_excludes_cwd_2575.py -q` -- 8 passed
- `python3 -m pytest tests/ -k "spawnable or bin_resolve or 2540 or 2542 or windows_cmd" -q` -- 23 passed, 4 skipped
- (Reasoned, not observed: the cwd-insertion behaviour itself is Windows-only
  and this was run on macOS. The tests construct the PATH list directly
  rather than depending on the host OS, so they exercise the same code path
  `which_excluding_cwd()` runs on Windows.)

[AI-generated]


fdaviddpt:
Scheduler review, reasoned from the code and documented Windows behaviour rather than executed -- I am on darwin and cannot run the Windows path.

The diff is right and it closes the escalation #2575 was filed for: a committable `ruff.cmd` at a repo root no longer reaches a spawn. The `_already_a_path` / `_is_executable` gating on a directory component is a genuine catch, and it closes a second route that never involved `which()` at all.

Two things worth recording before this merges, neither of them a reason to hold it.

## 1. `or name` leaves the narrower route open, so the title is stronger than the code

Both chokepoints still end the same way:

```python
return which_excluding_cwd(name) or name
```

On Windows, `CreateProcess` performs its own search when handed a bare command line, and that search includes the current directory of the calling process. So when `which_excluding_cwd` correctly refuses a cwd-only match and returns `None`, the fallback hands Windows the bare name and Windows resolves it against the repo under inspection anyway.

The plant has to be `ruff.exe` rather than `ruff.cmd`, because `CreateProcess` appends only `.exe` -- which is the same asymmetry this issue's own body relies on when it explains why the `.cmd` plant was a dead end at v0.61.0 and this delta made it live. That means the `.exe` route was reachable at v0.61.0 too: pre-existing, not introduced here, and correctly out of scope for a fix that has to unblock a tag.

But "never spawn a repo-planted file resolved only via the cwd" (this PR's title, and the shape of the new docstrings) is not what the code delivers on Windows. It delivers "never *resolve* one" -- and then hands the unresolved name to something that resolves it the same way. Worth a follow-up issue naming the `or name` fallback, with the honest framing that closing it means deciding what an adapter does when resolution fails, which is a behaviour change well outside this diff.

## 2. The gate and the spawn now search different paths

`validators/ruff/ruff.py:200` still gates on raw `shutil.which(TOOL)`:

```python
if not shutil.which(TOOL):
    emit(absent(TOOL, file, INSTALL_HINT, ...))
```

That call keeps the curdir insertion this PR exists to remove, so on Windows a repo-planted file satisfies the *gate* while `argv0` refuses to resolve it. The adapter then proceeds past "the tool is present" into a spawn that cannot find it, landing on the `which` said yes and exec said no arm at line 217 -- a confusing diagnosis for a condition that is actually "a repository tried to plant a binary".

Not a vulnerability on its own, and arguably the safe direction. But two searches that disagree about what is on PATH is the kind of split that reads as a flake later. Worth checking how many of the ~18 adapters gate on raw `shutil.which` while spawning through `argv0`.

Neither point blocks. #2575 is a tag blocker and this closes what it was filed for.


[AI-generated]

Issue #2540, opened 2026-09-15
Seven validator/formatter adapters cannot spawn a node CLI on Windows: which() resolves .cmd, subprocess cannot run it


fdaviddpt:
Seven adapters resolve a node-packaged CLI with `shutil.which()` and then spawn it with a bare name through `subprocess.run()`. On Windows those two calls disagree, deterministically, on every install.

`shutil.which()` consults `PATHEXT`, so it finds and returns `eslint.cmd`, `tsc.cmd`, `npx.cmd`. Python's `subprocess` without a shell goes through `CreateProcess`, which searches `PATH` but appends only `.exe`, and cannot execute a `.cmd` file at all. So `which()` says present and the spawn raises `FileNotFoundError`.

This is the same constraint PR #2538 hit from the Node side in `notifiers/claude-channel/start.mjs`: `spawnSync("npm", ...)` failed with `ENOENT` because the executable on `PATH` is `npm.cmd`, and naming `npm.cmd` explicitly raised `EINVAL` instead. Node and Python both bottom out in `CreateProcess`. That PR fixes one call site. These are the rest of the population.

## The sites

| File | Spawns | Line |
| --- | --- | --- |
| `validators/eslint/eslint.py` | `eslint`, then `npx --no-install eslint` | 133-138, 196 |
| `validators/stylelint/stylelint.py` | `stylelint`, then `npx --no-install stylelint` | 71-74, 116 |
| `validators/markdownlint/markdownlint.py` | `markdownlint` | 199, 205 |
| `validators/tsc-check/tsc-check.py` | `tsc` | 192, 198 |
| `validators/pyright/pyright.py` | `pyright` | 100, 105 |
| `validators/prettier-check/prettier-check.py` | `prettier` (`PRETTIER_BIN`) | 151, 171 |
| `formatters/prettier-write/prettier-write.py` | `prettier` (`PRETTIER_BIN`) | 63, 98 |

`node-check` and `html-check` are clean: they spawn `node`, which is `node.exe`. So are every adapter reaching `git`, `gh`, `glab`, `ruff`, `mypy`, `shellcheck` and the rest of the real executables. The defect is specific to CLIs shipped as npm packages, which are `.cmd` shims on Windows and nothing else.

## What it costs

[...]

PR #2585, opened 2026-09-17
fix(#2579): route every absent-tool gate through the cwd-safe spawnable() chokepoint

### LINKED ISSUE #2579: ruff.py's which() gate and its argv0() spawn now search different paths -- audit the ~18 adapters for the same split
Follow-up from #2575 / PR #2577's own review (maintainer, reading the diff directly, 2026-09-17 -- reasoned from the code and documented Windows behaviour, not executed; on darwin).

`validators/ruff/ruff.py:200` still gates on raw `shutil.which(TOOL)` before spawning through `argv0()` (which now routes through `which_excluding_cwd()`). PR #2577 removed the cwd-insertion from the *spawn* path but not from this *gate* path, so the two now search different PATH lists: a repo-planted `ruff.cmd` passes the gate (raw `which()` still finds it via cwd) and then fails to actually spawn (the corrected `argv0()` refuses it) -- landing in whatever `refusal.absent()` does today, rather than being silently prevented from running at all as #2575 originally described.

This is not a security regression -- the fix still holds -- but it is a correctness/consistency gap: the gate can say "found" while the spawn says "not found", for reasons a caller debugging a validator failure would have no way to connect without reading both call sites.

#2575's report named "roughly 18 adapters" consuming the shared `argv0`/`_spawnable` chokepoint (ruff, markdownlint, tsc-check, git-status, prettier-check, go-vet, actionlint, ruby-check, mypy, eslint, stylelint, terraform-check, hadolint, gitleaks, psr, pyright, gofmt-check, html-check, new-file-lint, cargo-check). Worth a sweep counting how many of those still gate on a raw `shutil.which()` call independent of the chokepoint the actual spawn now uses, and routing every one of them through `which_excluding_cwd()` (or the chokepoint itself) for the gate too.

[AI-generated]

fdaviddpt:
Follow-up from #2575/#2577's own review: #2575 fixed the *spawn* side of validator/formatter adapters (`argv0()`/`resolve_bin_cmd()` now route through `spawnable.which_excluding_cwd()`, excluding a repo-planted cwd match), but left each adapter's *absent-tool gate* calling raw `shutil.which()`, which still finds a cwd-planted shim -- so the gate could say "found" while the spawn said "not found", landing on the wrong error rather than being cleanly prevented from running.

## What changed

Routed every absent-tool gate through the same `spawnable()` chokepoint the spawn already uses, across:

- 23 files in `validators/`: the ~18 named in the issue (ruff, markdownlint, tsc-check, git-status, prettier-check, go-vet, actionlint, ruby-check, mypy, eslint, stylelint, terraform-check, hadolint, gitleaks, psr, pyright, gofmt-check, html-check, new-file-lint, cargo-check) plus 3 recon found the issue's own count missed (shellcheck, ci-lint) plus `jit-index`'s own separate, previously-unprotected `awk = shutil.which("awk")` presence check -- same defect shape but gate AND spawn were both unprotected via the same raw call, never routed through any chokepoint at all.
- All 4 files in `formatters/` (ruff-format, phpcbf, php-cs-fixer, prettier-write) -- same compound-gate shape as `validators/ci-lint.py`, discovered during the sweep, never part of either #2575's or #2540's prior fix.
- Removed now-dead `import shutil` from `ci-lint.py`, `jit-index.py`, and the 4 formatters.
- Updated ~20 existing test files whose fixtures faked tool-presence via `monkeypatch.setattr(mod.shutil, "which", ...)` with a companion `monkeypatch.setattr(mod, "spawnable", ...)` patch, since the gate no longer reads the old attribute at gate time.
- Added `tests/test_gate_matches_spawn_2579.py`: an AST-based register test (same shape as `tests/test_spawnable_register_2540.py`) that walks every file under `validators/`+`formatters/` and flags any adapter whose gate calls raw `shutil.which()` while its spawn already uses the cwd-excluding chokepoint. Verified red-before-fix against master's `ruff.py`, green after.
- Updated the stale jit-context rule `.claude/jit-context/paths/00-manual/validators.md`, which prescribed the old raw `shutil.which(TOOL)` snippet as canonical.
- `changelog.d/2579.fixed.md` added.

## Deliberately NOT changed: `validators/phpstan/phpstan.py`

`phpstan.py`'s `PHPSTAN_BIN` value is spliced as a literal script-path argument to `php` (`cmd = ["php", ..., phpstan_bin, "analyse"]`), never resolved through `argv0()`/`spawnable()`/`resolve_bin_cmd()` on either side -- so its gate has nothing to be inconsistent WITH, and fixing it alone would not close anything against this issue's own scope. This is a DIFFERENT, unconfirmed vulnerability shape: a cwd-planted file literally named `phpstan` (the default `PHPSTAN_BIN` value) could in principle be opened by `php`'s own file-open logic if it resolves the literal string relative to cwd, independent of both the pre- and post-#2575 chokepoint fixes -- no `which()`/`shutil.which()` call is even involved in the actual exploit path. This was not reproduced or confirmed (no `php` binary was invoked to test PHP's own relative-path resolution), so it is not filed as a new issue here and is not claimed as a confirmed finding -- flagged as an open question for a human to decide whether it merits its own issue, worded exactly as considered during review rather than asserted as a defect.

## Review

Two spawns per `agents/developer/review.md`: `Explore` (4 findings: 2 fixed real test regressions caused by the migration, 1 argued down, 1 fixed-partial with residue) and `oss:auditor` (1 unranked finding, fixed: hardened the new register test's own detector against a latent, non-live coverage gap for module-style `import spawnable` usage).

Below-bar residue from Explore's finding #4, recorded here as its own pr_anchor: the ~11 other adapters' generic "which said yes and exec said no" comments (tsc-check, ruff, markdownlint, ruby-check, actionlint, mypy, cargo-check, hadolint, stylelint, gofmt-check, terraform-check, shellcheck, pyright) read as materially less misleading than the two I fixed (eslint.py's docstring and git-status.py's comment, which specifically named `shutil.which()` for a branch this diff rewired), and removing 18 files' dead `import shutil` statements purely for tidiness was judged not worth the added surface given the register test is the real regression guard.

## Tests

Targeted run across the ~52 touched/adjacent test files: 799 passed, 25 skipped (pre-existing tool-absence skips). New register test `tests/test_gate_matches_spawn_2579.py`: verified red against master's `ruff.py` before the fix, 5/5 green after. Two test regressions the migration caused (`test_go_vet_stall_1461.py`, `test_adapter_wall_is_not_a_verdict_1604.py`'s ruby tests) were found by review and fixed, both re-verified passing under a genuinely empty `PATH` environment variable, not just a mock. Full repo suite deliberately not run locally, per this repository's own policy; CI is the source of truth.

Closes #2579

[AI-generated]


fdaviddpt:
## Maintainer review — #2585

**Check arithmetic:** 23 legs, 9 passed / 6 failed / 8 pending at last read; the decisive leg is
`lint (files this PR adds, renames or modifies)`, which is a genuine FAIL, not pending noise.

**Finding:** `lint-new` reports 20 of 50 touched `.py` files with a new `F401 'shutil' imported but
unused` finding (`validators/actionlint/actionlint.py:21`, `validators/cargo-check/cargo-check.py:27`,
and 18 more) — each one is a validator where this diff replaced its only `shutil.which(...)` call
with `spawnable(...)`/`spawnable.which_excluding_cwd()` but left the module's `import shutil` in
place. This is a **modified-path finding measured against the merge base** (`lint_new_files.py`'s own
scope note: "checked only for findings this PR introduces"), so it is caused by this diff, not
pre-existing debt.

**Disagreeing with the argued-down disposition.** The PR body's own Review section reasons: "removing
18 files' dead `import shutil` statements purely for tidiness was judged not worth the added surface
given the register test is the real regression guard." That call was made without checking it against
`lint-new`'s own repo-wide gate (F401/F841/F541 are re-enabled and enforced tree-wide for exactly this
class of finding, per `.github/scripts/lint_new_files.py`'s own comment: "an added path has no share
of that debt, so it is checked whole; a modified path has one, so it is checked only against its own
base revision"). It is not cosmetic here — it is the reason 20 of the 23 legs never got to run pytest
at all. Per `skills/manager/phases/review.md`, "a load-bearing argued-down finding still gets checked,
by hand, against the code" — checked, and it does not hold.

**Decision: needs-fix.** Remove the now-dead `import shutil` from the 20 files `lint-new` names
(the same cleanup already applied to `ci-lint.py`, `jit-index.py` and the 4 `formatters/` files in
this same diff, per the PR body's own "Removed now-dead `import shutil`" bullet — this is completing
that already-started cleanup consistently rather than a new ask).

No other finding from this pass: premise, blast radius (validators/formatters + their own tests,
consistent with the stated scope) and the `phpstan.py` scope note all check out; the `#2579`-vs-
`#2575` distinction the PR draws is accurate. Not merged — checks are red.

[AI-generated]

PR #2627, opened 2026-09-18
feat(#1318): add ops:grep=PATTERN, a filter for the op roster

### LINKED ISSUE #1318: `ops` has no filter, so finding the right op means piping it through grep — the one pipe a block rule forbids
From the #1252 agent, and I hit the same thing twice this session.

`supertool 'ops'` prints the whole roster — 85 ops. Answering "which op creates a file" means reading all of it or piping it through `grep`, and piping a supertool op is what `supertool-no-cut.md` blocks. So the tool's own discovery surface pushes the caller at the one motion the tool's own rules forbid.

    supertool 'ops:grep=PATTERN'

would close it: match on the op name, its syntax string and its description, print the matching rows whole, and say how many of the total matched — `4 of 85`, so a small result set cannot be misread as a small roster.

Related and cheaper than it looks: **#1222** already asks for a near-miss line on `unknown operation`, which is the same information arriving one call later. A filter would make the near-miss trivially derivable rather than a second lookup table.

The count is the argument for both: this session ran `ops` and then re-read it for a name three separate times.

fdaviddpt:
Closes #1318.

`supertool 'ops'` printed the whole op roster with no way to narrow it, and piping the output through `grep` is exactly what the repo's own raw-command guard blocks -- leaving redirect-to-a-temp-file as the only route, reported independently by three agents in one week (see the issue thread).

Adds `ops:grep=PATTERN` (scoped alongside `ops:roster`/`ops:full`/`ops:session`), matching op name, syntax and description, and always stating `N of M ops matched` -- including `0` -- so an empty result can never be misread as a short roster.

Two review rounds found and fixed:
- colon truncation: `ops:grep=PATTERN` silently truncated `PATTERN` at its first colon (the tokenizer's own limitation), so a search for `read:PATH`-style syntax -- the op's own named use case -- ran a narrower, wrong pattern with no error. Fixed by reading the pattern off the raw dispatch string.
- the filter only searched ops with a `.supertool.json` entry, so a real dispatchable op with none (`introduction`, `version`, etc) read `0 of 101 ops matched` -- the exact absence-as-answer defect this issue was filed to remove, reintroduced one level down.
- an empty-config state answered `0 of 0 ops matched`, literally claiming zero ops exist.
- the filter was unconditionally case-insensitive, unlike every sibling pattern-taking op (`grep`, `around`, `between`, `read`'s own `grep=`).
- a stale illustrative op-count figure in the docstring and `docs/operations/meta.md`, replaced with generic `N of M` language.
- a test docstring overclaimed which fix it pinned; renamed and reworded to say what it actually checks.

Also rewrites the `docs/operations/meta.md` paragraph that had explicitly documented "deliberately no `ops:PATTERN` filter" two days before this issue was filed.

[AI-generated]


fdaviddpt:
**Maintainer review (tick-review): merge-ready, one below-bar finding.**

Verified: check arithmetic (23/23 SUCCESS), blast radius (docs/README both updated to reflect the new `ops:grep=` mode, no unrelated files touched), `fix_commit_scope.py --base 7d66693d --head 576e82fe` -> `needs-second-pass` (4 files) with the second pass present and recorded (`7d58f3b0`, itself 1 file, docstring-only), and a targeted red re-run of `tests/test_ops_filter_1318.py` against `origin/master` with the fix absent (9 of 12 tests fail as expected).

**Below-bar finding, real but non-blocking:** of the four named edge cases (colon-containing patterns, undocumented ops, empty config, case sensitivity), three discriminate correctly against the pre-fix commit. `test_ops_grep_empty_config_lists_names_not_zero_of_zero` does not -- it passes unchanged against `origin/master` (fix absent) for an unrelated reason: with the fix absent, `ops:grep=version` falls through to the generic unknown-argument refusal, whose text echoes the argument back verbatim (quoting `grep=version`), which contains the substring `version` and satisfies `assert "version" in out` and `assert "0 of 0" not in out` without ever reaching the new empty-config branch. The feature itself is correct -- manually confirmed against the PR branch: setting the config empty then dispatching `ops:grep=version` returns `5 of 106 ops matched` including `version`, not `0 of 0` -- so this is a weak/vacuous test, not a functional defect. Worth tightening (assert on a phrase only the empty-config branch emits) whenever this file is next touched; not worth blocking this merge over.


[AI-generated]

Issue #1867, opened 2026-08-20
edit takes one edit per payload and the batch shape is not discoverable from ops, so seven mechanical edits to one file cost seven heredocs


fdaviddpt:
Reported from `Digital-Process-Tools/claude-oss`, which declares supertool as a dependency and routes every file operation through it. Found by an agent working under a brief that mandates supertool for all writes — so this is the cost paid on the supported path, not on a workaround.

## Symptom

`edit` takes exactly one edit per payload. A payload using an `[[edits]]` array is refused with `unknown field 'edits'`.

## Cost, measured on one run

Seven mechanical edits to one test file cost **seven separate heredocs**, plus one refused round-trip and a re-send of the whole seven-edit payload as seven singles.

## Why this is not a complaint about the refusal

The refusal is correct and its message names the accepted keys. The gap is that **the accepted shape is not discoverable from `ops`** — an author reaching for a batch does not learn it is unsupported until the write is attempted, and by then the payload is composed.

## What would settle it

Either of these closes it; they are different amounts of work and it is your call which:

- `edit` accepts `[[edits]]`, applying them in order with the existing post-write validator and rollback over the whole batch. That is the shape the author reached for.
- Or it stays one-per-payload and `help:edit` / the refusal says so as a positive statement — "one edit per payload; send N payloads" — rather than leaving it inferable only from the unknown-field list.

The second is cheap and removes the surprise; the first removes the cost.

## Not claimed

- That the one-edit-per-payload contract is wrong. It may well be right for rollback semantics, and if so the fix is the documentation half.
- That this blocks anything. It is per-multi-edit-file overhead, paid in tokens and round-trips.

Issue #1773, opened 2026-08-16
`help:OP` answers for builtin ops only inside this checkout — every other project gets "no documented help" for read, grep, paste, edit


fdaviddpt:
Observed while working in a `claude-oss` worktree, which is a plain supertool consumer: three presets in its `.supertool.json`, no `builtin-ops` block of its own.

```
$ cd ~/Documents/claude-oss-wt/204-205
$ supertool 'help:paste' 'help:read' 'help:grep' 'help:gh-pr'
--- help:paste ---
ERROR: op 'paste' has no documented help in .supertool.json. It's a valid operation here — run 'ops' for the list, or see docs/operations.
--- help:read ---
ERROR: op 'read' has no documented help in .supertool.json. It's a valid operation here — run 'ops' for the list, or see docs/operations.
--- help:grep ---
ERROR: op 'grep' has no documented help in .supertool.json. It's a valid operation here — run 'ops' for the list, or see docs/operations.
--- help:gh-pr ---
gh-pr:NUMBER_OR_BRANCH[:status|:full|:diff[:PATH]|:threads]
...
```

The same four calls from `~/Documents/claude-supertool` all answer in full.

## Mechanism

`op_help` (`_supertool.py:16532`) reads `builtin-ops`, `ops` and `aliases` out of whatever `_load_config()` returned, and `_load_config()` (`_supertool.py:1131`) is the config found by walking up from **cwd**, plus the presets that config names. Preset ops therefore carry their help wherever they are installed — `gh-pr` above proves it. Builtin ops do not: their `builtin-ops` documentation lives in **this repository's own `.supertool.json`**, which is not a preset and is not merged into anyone else's config.

So the split is not "documented / undocumented". It is **preset op / core op**, and the core ops are exactly the ones a new user reaches for first. Every plugin install outside this checkout has a dead `help:` for `read`, `grep`, `glob`, `paste`, `edit`, `replace`, `vim`, `between`, `batch` and the rest of the core surface.

## Why the error text makes it worse

The refusal is well-formed and still misleads twice:

- *"has no documented help in `.supertool.json`"* reads as a gap in the project's own config, i.e. something the reader could fix by writing docs locally. It is not — the docs exist, upstream, and are simply not shipped.
- *"see docs/operations"* points at a path that does not exist in a consumer repo. There is no `docs/operations` next to a plugin install; that directory is here.

The reader is told the answer is somewhere they can look, and both places named are wrong.

[...]

Issue #1435, opened 2026-08-11
A quoted `grep` pattern matches the quote characters, so ordinary shell hygiene returns a well-formed zero


fdaviddpt:
Reproduced by hand just now, on `master` at `260468e`, one call, both forms side by side:

```
$ supertool 'grep:"Valid operations":_supertool.py:3' 'grep:Valid operations:_supertool.py:3'
--- grep:"Valid operations":_supertool.py:3 ---
(0 results in 0 files, scanned 1 files, limit 3)

--- grep:Valid operations:_supertool.py:3 ---
(1 results in 1 files, scanned 1 files, limit 3)
_supertool.py
  1976:           f"Valid operations: ..."
```

The quotes are taken as **literal characters in the pattern**. Nothing in the render says so.

## Why this is the house defect rather than a papercut

`scanned 1 files` is the disclosure this repo added so that a zero says whether the tool looked — and here it says *yes, I looked, and it is not there*. That is the strongest possible statement of an absence, and it is wrong. The reader's next move is to conclude the string is not in the file.

**Quoting a pattern is not a mistake anyone gets warned out of.** It is what you do with a phrase containing a space, in every other tool a shell user has ever typed into. The op that most needs a pattern quoted is the one that punishes it.

**It can also mask itself.** An agent hit this tonight with a quoted `blame` and got hits, because the file genuinely contains that word wrapped in quotes — so the first probe corroborated the wrong mental model before a second one disproved it.

## Nearest existing op and what it cannot answer

The op is `grep`; nothing else is nearer. What it cannot do today is distinguish *the caller quoted a phrase* from *the caller wants a quote character in the match*. Both are legitimate, so stripping quotes unconditionally is not obviously right — a pattern hunting a quoted word is exactly the second case, and it is the one that hid this.

Three shapes worth weighing, and choosing is the work:

1. **Strip a symmetric outer pair and say so in the render** — one line, the same shape as `around`'s auto-correction to `around_line`, which explains itself where the reader is looking. Costs the ability to search for a quoted phrase *including* its quotes, unless an escape exists.
2. **Do not strip; disclose.** When a pattern begins and ends with the same quote character and returns zero, add a line naming the other reading. Keeps the grammar and closes the silence, at the cost of a second call whenever the guess was right.
3. **Refuse the ambiguous form** and name both spellings, the way #1417's `grep` gate now declines an empty PATH slot rather than widening the population silently.

[...]

PR #2629, opened 2026-09-18
fix(#1777): repeat read's window note at the foot, name case-insensitivity on grep's 0 results

### LINKED ISSUE #1777: Four small findings from one session: a refusal that names no alternative, two disclosures positioned where they cannot be read, and two undocumented write ops
Affected version: `supertool 0.46.0`. Four findings from one delegated agent's session in `claude-oss`, filed together because each is small and none is the same bug. The serious one from that session is filed separately.

Common thread worth naming: **three of the four are cases where the disclosure is genuinely good and is positioned where the caller cannot read it.** That is a different defect from a missing message, and it may deserve a single answer.

## 1. `edit:@-` refuses a TOML array-of-tables payload with `unknown field(s) edit`

The batching instruction agents are given says to pipe edits in as a TOML payload, so `[[edit]]` is the obvious first guess for editing one file several times. It is refused, and the refusal is correct and well-argued — but it names no batching form, so a batch of edits to one file costs one call each.

Either the array shape should work, or the refusal should name the form that does.

## 2. `read:PATH:195:300` silently means OFFSET:LIMIT, while `between:PATH:200:300` refuses and names the fix

`between` gets this exactly right:

```
ERROR: between does not take line ranges — it is between:SYMBOL:PATH ...
  For lines 30-70 use: read:commands/release.md:30-70  (inclusive, 1-based)
```

`read` with the same colon shape accepts it as offset-and-limit. It does disclose which form it read, how many lines it skipped, and what to type for the other one — the disclosure is excellent. **It sits above the file content**, so with a 189-line window the correction scrolls off and the caller reads the wrong lines believing they are the right ones.

Consider putting that line at the foot as well as the head, or refusing the colon form the way `between` does.

## 3. `grep` is case-sensitive and the refusal path surfaces no flag

The instruction agents work under says to confirm a write a second way. Grepping back a sentence that is on disk with different capitalisation returns `0 results in 0 files` — a false negative on a *successful* write, produced by the exact step that exists to catch a failed one.

`0 results` is the one output where a case-insensitivity hint is worth printing, because it is the only one where the caller cannot tell a real absence from a spelling mismatch.

## 4. `help:paste` and `help:append` answer `no documented help`

Two ops that write files — one of which truncates — with no help text, listed under *"Also accepted, no reference in .supertool.json"*. This is also the root of the separately-filed issue: a caller with no help to read guesses the convention from the documented ops.

## Not established

All four are from one session, on macOS, `supertool 0.46.0`, one caller. Frequency is unknown; each is reproducible on demand.

fdaviddpt:
Two of the four findings in #1777 had already shipped before this lane started (finding 1 via #1867, finding 4 via #1773) -- this pull request implements the remaining two.

## Finding 2: `read:PATH:OFFSET:LIMIT`'s disclosure sits only above the content

`read` already discloses which colon form it read, how many lines it skipped, and what to type for the other form -- the disclosure itself is good, it is only positioned where a long window (189+ lines in the reported case) scrolls it off before the caller reaches the lines it corrects. The identical note string is now also appended at the foot of the output, gated on the same `offset > 0` condition as the existing header copy, so the two copies cannot drift apart.

## Finding 3: `grep`'s `0 results` gives no case-insensitivity hint

The instruction agents work under says to confirm a write a second way. Grepping back a sentence that is on disk with different capitalisation returned `0 results in 0 files` -- a false negative on a successful write, produced by the exact step that exists to catch a failed one. A new `_case_insensitive_note` helper (modelled on the existing `_quote_pair_note` from #1435) probes with an inline `(?i)` prefix and only prints when that flips a zero to a match, wired into all three `op_grep` zero-result branches (count_only, context, plain).

Guarded so the probe -- a second full-corpus scan -- never runs on a pattern with no letters at all (numbers, punctuation), where case can never matter; an earlier version of this fix paid that cost unconditionally and measured a ~3x slowdown on a genuinely-absent, letter-bearing pattern over ~1700 files. The suggested rerun is deliberately unquoted (`grep:(?i)PATTERN:PATH`) -- supertool's grep does not strip a quote pair, so a quoted suggestion would have sent the caller straight into the #1435 quote-pair zero instead of resolving this one.

## Review

Two self-review rounds (Explore + oss:auditor in parallel, round 1; Explore alone, round 2, triggered by `fix_commit_scope.py`'s file-count threshold). All 4 findings across both rounds were fixed: an incorrect quoted rerun suggestion, an unconditional probe cost on letter-free patterns, and two tests that did not actually pin the behaviour they claimed to (one vacuous single-element self-comparison, one that could not distinguish "probe skipped" from "probe ran and (?i) was a no-op").

## Tests

Red then green, both new test files written and run standalone before any production code changed. Full suite deliberately not run locally -- CI is the merge gate.

Closes #1777

[AI-generated]


fdaviddpt:
Maintainer review (tick-review, oss:sub-manager delegate) on the final head commit (d828daca): check arithmetic 23/23 passed, blast radius (2 new test files + `_supertool.py` + changelog fragment) matches the two findings claimed, `fix_commit_scope.py --base 468f9e41 --head d828daca` reports `within-scope` (no second review round required), and the targeted red re-run against master (01425ee, fix absent) fails 8 of the 17 new assertions -- both new test files are non-vacuous.

One below-bar finding, real but non-blocking: **Finding 2's description above is stale relative to the shipped code.** It says the footer echo is "gated on the same `offset > 0` condition as the existing header copy," but the last commit (d828daca, "CI: gate the footer echo on window size, not on offset alone") changed that gate to `FOOTER_ECHO_MIN_LINES` (printed-line count), precisely because an `offset > 0` gate broke `tests/test_hint_points_at_the_fix_1489.py` and `tests/test_read_range_382.py` on CI. The code and the code comments are correct; only this PR-body paragraph was never updated after the fix. Not asking for a body edit -- routing this as a comment rather than gh-pr-edit to avoid re-triggering the closing-reference re-parse over a wording-only change.

Decision: ready-to-merge.

[AI-generated]

Issue #2593, opened 2026-09-17
Finish the YouTube preset: youtube_reply, youtube_like, youtube_status_since


fdaviddpt:
Split from #227, which is being closed as superseded now that the read half (#2527: `youtube_search`/`youtube_list`/`youtube_read`) and the OAuth write half (#2543: `youtube_auth`, `youtube_comment`) have both merged, plus a bugfix pass (#2584/fix#227) against the merged OAuth code.

Three ops from #227's original proposal remain unwritten on purpose (see `docs/presets/youtube.md`'s own Scope section and `changelog.d/227.added.youtube-writes.md`):

- `youtube_reply:COMMENT_ID|TEXT` -> `comments.insert`, 50u, OAuth2
- `youtube_like:VIDEO_ID` -> `videos.rate?rating=like`, 50u, OAuth2
- `youtube_status_since[:ISO]` -> `commentThreads.list` on own videos filtered by `publishedAt > ISO`, ~1u/video, API key

All three share the OAuth/quota/rate-cap machinery `youtube_auth`/`youtube_comment` already built (#2543) -- this is additive within an existing module, not a fresh design. #227's own body has the full op table, auth model and guardrails (no-batch-writes, 5 writes/hour rate cap, `sent.jsonl` audit log, visibility != HTTP-success warning) if starting from scratch here.

[AI-generated]

PR #2637, opened 2026-09-19
fix(#2636): flatten \n and \r out of injection-hit joins at the remaining 8 call sites

### LINKED ISSUE #2636: youtube/_sanitize.py wrap() and read.py render() still forge column 0 on a newline-carrying injection hit (7 unfixed call sites, 4 preset dirs)
## Symptom

`_sanitize.py::wrap()` and each preset's own `read.py::render()` join an
injection-scan hit straight into a printed warning line with no newline
stripping. A comment or post body containing a bare `key:`-shaped line (one
of `_sanitize.detect()`'s own patterns, `(?:^|\n)\s*system\s*:`, matches
across an embedded newline) can put attacker-chosen text at column 0 of the
tool's own `⚠ POSSIBLE INJECTION` banner, indistinguishable from a line the
tool wrote itself. #2593 found and fixed this exact shape at a third call
site (`presets/youtube/status_since.py::render()`, new in that diff) but left
the two pre-existing call sites per preset directory untouched, and it is not
one preset -- the same two-line shape is duplicated, unfixed, in at least
four preset directories.

## Location

`presets/youtube/_sanitize.py:102`:
```python
lead = f"⚠ POSSIBLE INJECTION — review carefully ({\', \'.join(hits[:3])})\n"
```
`presets/youtube/read.py:91` (same shape):
```python
warning = f"⚠ POSSIBLE INJECTION in this video\'s text — {\', \'.join(inj_hits[:3])}\n"
```
The identical `_sanitize.py:102` / `read.py` pair is duplicated, unfixed, in
`presets/bluesky/_sanitize.py:102` + `presets/bluesky/read.py:80`,
`presets/devto/_sanitize.py:102` + `presets/devto/read.py:96`, and
`presets/hashnode/_sanitize.py:102` (its own `read.py` not checked in this
pass) -- 4 preset directories, `_sanitize.py:102` confirmed identical in all
four by direct read.

## Mechanism

`_sanitize.detect()` returns the matched substring itself
(`m.group(0)[:60]`), not a canned pattern name, and one of its own patterns
matches across an embedded newline in third-party text. `wrap()` and
`render()` both join that raw hit into an f-string with no `\n` stripping,
so the untrusted text can start at column 0 of the printed line, which is
exactly the shape #2593's fix addressed at the one call site the issue
actually opened (`presets/youtube/status_since.py::render()`).

## What would settle it

Flatten each hit (`[h.replace("\n", " ") for h in hits[:3]]`) before the
`join()`, the same one-line fix #2593 applied to
`status_since.py::render()`, at each of the 7 remaining call sites (2 per
preset directory x 4 directories, minus the one already fixed). A fixture
per site with an embedded newline in the matched text, mirroring
`tests/test_youtube_reply_like_status_2593.py::test_status_since_injection_warning_never_carries_a_raw_newline`,
is what would pin each fix -- the existing fixtures carry no newline, which
is why this shape shipped unnoticed at every site except the one #2593
touched.

Originally logged as `trap.d/2593.injection-hit-join-forges-a-line.md`;
filed here per `skills/manager/phases/findings.md`'s ranking table --
`forges` (text somebody else wrote reaches column 0 of a receipt the loop
parses) blocks a release unconditionally, so a trap.d fragment alone
under-routes it.

[AI-generated]

fdaviddpt:
AI-generated fix for #2636.

## What this closes

Closes #2636

## Symptom

_sanitize.py::wrap() and each preset's own read.py::render() (bluesky, devto, hashnode,
youtube) joined a prompt-injection scan hit raw into the printed "POSSIBLE INJECTION" warning
line. _sanitize.detect() returns the raw matched substring, not a canned pattern name, and one
of its own regex patterns matches across an embedded newline in untrusted text. Joining that raw
hit into an f-string with no stripping let attacker-controlled text land at column 0 of a
second, forged-looking line.

#2593 fixed this exact shape at one other call site (presets/youtube/status_since.py), but that
fix lives only on the still-unmerged branch fix/2593 (open PR #2635, all 23 checks green, not
yet merged) -- it is not on master. This PR closes the 8 remaining call sites: 2 per preset
directory (the shared _sanitize.py::wrap() and the preset's own read.py::render()) across all 4
preset directories, including presets/hashnode/read.py, which #2636's own issue body says was
"not checked in this pass" but which this lane confirmed by direct read carries the identical
unfixed shape.

## Fix

Two commits:
1. Flattens \n in each of the 8 sites (hits[:3] with newlines replaced by spaces before the
   join), one regression test per site.
2. A self-review finding (independent oss:auditor spawn): the \n-only flatten still let a raw
   \r survive, because _sanitize.py's pattern (?:^|\n)\s*system\s*: can match across \n\r
   together -- the \s* right after the literal \n anchor consumes a trailing \r. Fixed by also
   flattening \r at all 8 sites, pinned by a new red-then-green regression test.

Both fixes were watched red before green; see the report for full command output.

## Not in scope for this PR

tests/test_unanswerable_checks_693.py's SANITIZE_PRESETS tuple omits youtube from its own
drift-parametrized behavioral tests -- correctly out of scope for this diff, not fixed here.
This is a real, reachable gap in the same subsystem, but not part of this diff's own mechanism
(the column-0 banner-forgery defect); recorded for a future pass.

## Testing

- `pytest tests/test_sanitize.py tests/test_youtube.py tests/test_bluesky.py tests/test_devto.py tests/test_hashnode.py tests/test_unanswerable_checks_693.py -q --no-cov` -> 332 passed
- Full suite left to CI (12-leg matrix) per repo policy.
- Two independent reviewers (Explore, oss:auditor) ran read-only after both commits; both returned clean, both re-verified test and byte-identity claims independently.

[AI-generated]


fdaviddpt:
Maintainer review (tick pass, #2636/#2637): merging as ready-to-merge — checks 23/23 green, blast radius confined to the sanitize/read subsystem across all 4 preset dirs (matches the issue), and a manual red re-run on the pre-fix base commit confirmed the 6 new `\n`-flatten regression tests are non-vacuous (they fail without the fix, pass with it).

One below-bar, non-blocking observation from this pass, named rather than silently dropped: the second commit's `\r` flatten is applied at all 8 call sites, but the dedicated regression test for the `\r` case (`test_wrap_injection_warning_never_carries_a_raw_carriage_return`) only exercises the shared `_sanitize.py::wrap()`, not the 4 presets' own `read.py::render()` inline flatten. I verified by hand (against `presets/hashnode/read.py`'s `render()`, reusing the PR's own `\n`-test fixture with a `\r` inserted) that the fix is correct there too — no raw `\r` reaches the banner — so this is a coverage gap only, not a live defect. Not filing an issue for it; noting it here as the record.


[AI-generated]

Issue #1868, opened 2026-08-20
A paste payload whose content is source code pays a full re-send for quoting collisions the author cannot see until the write is attempted


fdaviddpt:
Reported from `Digital-Process-Tools/claude-oss`, which declares supertool as a dependency and routes every file operation through it. Found by an agent writing test fixtures under a brief that mandates `paste` for file creation.

## Symptom

Two `paste` payloads were rolled back before landing, each costing a full re-send of a ~6KB `content` field:

1. A TOML literal block carries no escapes, so a **Python source fixture containing a nested triple-quoted string** could not be written the obvious way.
2. A single doubled backslash inside a Python string literal being written — an escaped newline in a test assertion — triggered the `literal_backslashes` refusal.

**This issue body reproduced case 2 while being written.** The first attempt to `paste` this text was refused, because the sentence above quoted the offending sequence literally. That is not a complaint — it is the cleanest available demonstration that the collision is structural rather than a mistake anyone can learn their way out of.

## Why this is not a complaint about either refusal

**Both refusals are correct and both messages are precise.** The `literal_backslashes` one in particular states the two opposite fixes and says nothing can tell which was meant, which is the right behaviour: guessing would produce a file that passes every validator and is wrong only in string contents.

The cost is structural and worth naming on its own: **a payload whose `content` is source code in the same language family as the payload format pays a re-send for quoting collisions the author cannot see until the write is attempted.** Writing Python that contains quotes and backslashes, inside TOML that assigns meaning to quotes and backslashes, is the common case for anyone whose brief says "create files with `paste`" — and test fixtures are the densest instance of it.

## What would settle it

Stated as options rather than a request, since the trade is yours:

- A dry-run or lint mode that reports the refusal **without** requiring the payload to be re-sent. The re-send is the whole cost; the refusal is cheap.
- Or a `content_file` / stdin-delimited form where the content is not TOML-parsed at all, so quoting in the payload format cannot collide with quoting in the content.
- Or nothing, and this is filed as a known cost. That is a legitimate outcome and recording it is the point.

## Not claimed

- That either refusal is wrong. Both prevented a wrong file from landing.
- That a workaround does not exist. The author re-sent and it worked; the re-send is the cost.

fdaviddpt:
Triage sweep (2026-09-16) proposes this issue as the parent of two clusters, both resting on the TOML literal-string payload mechanism this issue already reasons through:

[...]

PR #2653, opened 2026-09-19
gh-issue-comment: add :edit=COMMENT_ID to correct a posted comment

### LINKED ISSUE #2643: No gh-issue-comment-edit op: correcting a posted comment goes through raw `gh api PATCH` with no byte-identical read-back
## Observed

2026-09-19, a design pass posting to #1868: the comment went out with a wrong figure (`7.4 KB` where the measured value was 6,376 characters). The only route to fix it was `gh api -X PATCH repos/.../issues/comments/ID -f body=@file`, hand-verified afterwards. `gh-issue-comment` re-reads what it posted and reports `byte-identical to what was sent (N characters)`; the raw PATCH does not, and it does not append the AI-authorship disclosure either, so an edited comment can silently lose the marker the create path added.

## What an op would answer

`gh-issue-comment-edit:COMMENT_ID:@FILE | :@-` (or `gh-issue-comment:NUMBER:@-:edit=COMMENT_ID`): PATCH the body, re-read it, report byte-identical or not, keep the disclosure line. The comment id is already in `gh-issue-comment`'s own `[result]` line (`comment id=5741496503`), so the caller has it without another lookup. Same for PR review comments if `gh-pr-comment` grows the same need.

This replaces one raw `gh api` call plus one raw read-back, and answers a question the raw call does not: whether what landed is what was sent.

[AI-generated, filed by the scheduler session of /oss:run]

[AI-generated]

fdaviddpt:
Closes #2643.

`gh-issue-comment` had no way to correct an already-posted comment -- the only route was raw `gh api -X PATCH repos/.../issues/comments/ID`, hand-verified afterwards, with no byte-identical read-back and no re-application of the AI-authorship disclosure marker if the corrected body dropped it. Observed 2026-09-19 on a comment to #1868 that went out with a wrong figure.

A trailing `:edit=COMMENT_ID` now switches the op from POST (new comment) to PATCH against `repos/{repo}/issues/comments/{edit_id}`, reusing the same payload parsing, the same four-state read-back verdict (EXACT/NORMALISED/MISMATCH/UNKNOWN), and the same disclosure step the create path already gives. The comment id is already in the ops own `[result]` line from when the comment was first posted, so the caller has it without another lookup.

`parse_args` grew from a 3-tuple to a 4-tuple `(number, payload_path, edit_id, error)`, mirroring `gh-pr-edit`s own `unlink`-token shape. A non-numeric edit id is refused before any gh call, and an edit id combined with an unrelated trailing token is still refused rather than silently swallowed.

## Self-review

Explore and oss:auditor were spawned concurrently against the first commit. oss:auditor returned NO FINDINGS across its four classes. Explore returned 3 findings:

1. A stale docstring ("this op has one shape") contradicted the same diffs own new edit-mode paragraph -- fixed.
2. The pre-existing create/POST branch had no positive-control test asserting it actually sends POST rather than PATCH -- fixed, with a new test that was confirmed red against a manually-broken create path before going green.
3. The raw route the issue is about (`gh api -X PATCH .../issues/comments/ID`) remains unguarded by the `replaces` guard system, the same pre-existing gap `gh-pr-edit`s own PATCH route already has. **This is not fixed here** -- it spans the whole github preset family rather than this issues blast radius, and should be filed as its own cross-cutting issue rather than patched piecemeal in this PR.

## Tests

Red before fix (implementation stashed against an already-updated test file): 9 failed, 20 passed. Green after: 33 passed. Targeted runs only (`tests/test_gh_issue_comment_2078.py`, `tests/test_gh_issue_comment_help_text_2444.py`, `tests/test_forge_disclosure_issue_comment_691.py`, plus the full registry-consistency group, 11 files) -- the full suite is left to CI.

[AI-generated]


fdaviddpt:
Maintainer review (tick-review spawn).

**CI: needs-fix.** `pytest (ubuntu-latest, 3.11)`, `(ubuntu-latest, 3.10)` and `(windows-latest, 3.10)` all fail the same way, in `tests/test_meta_doc_figures_1783.py`:

```
FAILED tests/test_meta_doc_figures_1783.py::test_the_footer_reports_the_render_and_the_doc_quotes_the_difference - AssertionError: meta.md does not quote the descriptions' cost (84,778)
FAILED tests/test_meta_doc_figures_1783.py::test_no_other_figure_claims_a_different_descriptions_cost - AssertionError: meta.md states ['84,372'] bytes for the descriptions' cost somewhere in the file, but the live figure is 84,778
```

This diff (`presets/github.json`) grows `gh-issue-comment`'s own `description` by the `:edit=COMMENT_ID` paragraph, which moves the live "descriptions' cost" figure to 84,778. `docs/meta.md` still quotes the old figure (84,372) — this is the blast-radius case `review.md` names directly: a diff that changes a file convention (a description's byte length) without touching the diagnostic that quotes it. Regenerate/update `docs/meta.md`'s quoted figure and this should go green; it is deterministic across all three failing legs, not a flake.

**Self-review finding #3 (raw `gh api PATCH .../issues/comments/{id}` left unguarded by the `replaces` system) — routed below-bar, not filed.** Checked against `findings.md`'s intake bar: no cost was paid yet (no red build, no wrong answer that reached a consumer, no maintainer round-trip caused by it), and it is explicitly *not* the first reachable instance of the class — the PR body itself says `gh-pr-edit`'s own PATCH route already has the identical gap. Recording the decision here rather than opening a new issue or silently dropping it, per `review.md`'s fourth receipt.

[AI-generated]

